PR #2103, opened 2014-03-16
Introduce new RqlConnectionError

josephglanville:
RqlRuntimeError has too large of a scope and it
is not possible to easily determine which of these
instances are retry-able errors.

RqlConnectionError is to be used for retry-able connection
and server errors.

A separate exception should be introduced for cases where
the cluster is unwritable due to loss of the primary shard.

### DIFF
--- a/drivers/javascript/errors.coffee
+++ b/drivers/javascript/errors.coffee
@@ -4,6 +4,8 @@ class RqlDriverError extends Error
         @msg = msg
         @message = msg
 
+class RqlConnectionError extends RqlDriverError
+
 class RqlServerError extends Error
     constructor: (msg, term, frames) ->
         @name = @constructor.name
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -47,7 +47,7 @@ class Connection extends events.EventEmitter
             if e instanceof err.RqlDriverError
                 callback e
             else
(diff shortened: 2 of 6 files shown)


danielmewes:
@josephglanville Sorry this has been dormant for so long. We really appreciate your help.
This was overshadowed by some other work we had to do, but it's still on our list.

I'm hopeful that we can get it into the release after 1.16.

@mlucy In case you end up getting all other release critical things done before the 1.16 feature freeze on the 16th, I have no objection in shipping this with 1.16. Though I think more likely it's going to go into the next one...

danielmewes:
Argh this again didn't make it into the 2.0 branch. I'd still really like to get this in soon. Moving to 2.1.

danielmewes:
We ended up implementing a new error hierarchy in 2.1.0 that builds on @josephglanville's idea.

See http://rethinkdb.com/docs/error-types/ for details.

The retry-able error that is thrown if a server is not available is the `ReqlAvailabilityError`.

danielmewes:
Also thanks again @josephglanville for your pull request. Sorry we didn't end up using your implementation (this was simply because we took the opportunity to change other error types as well while we were at it).

josephglanville:
@danielmewes all good, the end result is awesome. Thanks for eventually getting around to it. :)

PR #3604, opened 2015-01-19
node: use semver correctly

jonathanong:
If bluebird releases a version 3, then a breaking change was introduced, and there's a good chance your entire driver could break. Luckily, the maintainer of bluebird is actually good so I doubt this would actually happen.

### DIFF
--- a/drivers/javascript/package.json
+++ b/drivers/javascript/package.json
@@ -14,5 +14,5 @@
     , "url" : "http://github.com/rethinkdb/rethinkdb.git"
     }
 , "engines" : { "node": ">= 0.10.0" }
-, "dependencies" : { "bluebird": ">= 2.3.2" }
+, "dependencies" : { "bluebird": "^2.3.2" }
 }


danielmewes:
Thank you for your pull request @jonathanong.
If you haven't already, would you mind signing our contributor agreement http://rethinkdb.com/community/cla/ ? 

@AtnNn do you have any thoughts on this?

AtnNn:
@jonathanong Would it not be best to add an extra `<` constraint to remain compatible with older versions of npm?

jonathanong:
@AtnNn you only support node `0.10`, so you won't hit that problem. the "old" version of npm is only included with `0.8`, which you don't support anymore. as long as users aren't installing a really old version of `0.10`, then it should be okay.

but yeah, `>= 2.3.2 < 3` would work as well.

danielmewes:
@AtnNn what should we do about this one?

AtnNn:
@danielmewes Implement the solution described in https://github.com/rethinkdb/rethinkdb/pull/3604#issuecomment-70579748

danielmewes:
@jonathanong Could you update the pull request to use a version range?

jonathanong:
https://github.com/rethinkdb/rethinkdb/pull/3823 (doing this through the github CLI).

PR #3701, opened 2015-01-30
Add --keys and --index options to export script

mbrock:
This lets you pick a specific subset of keys to export, or to export by lookup on secondary index.

### DIFF
--- a/drivers/python/rethinkdb/_export.py
+++ b/drivers/python/rethinkdb/_export.py
@@ -45,6 +45,8 @@ def print_export_help():
     print("  --fields FIELD,FIELD...          limit the exported fields to those specified")
     print("                                   (required for CSV format)")
     print("  --keys KEY,KEY...                limit the exported documents by primary key")
+    print("                                   (another index may be given by --index)")
+    print("  --index INDEX                    used with --keys to search by a non-primary index")
     print("  -e [ --export ] (DB | DB.TABLE)  limit dump to the given database or table (may")
     print("                                   be specified multiple times)")
     print("  --clients NUM                    number of tables to export simultaneously (defaults")
@@ -75,6 +77,7 @@ def parse_options():
     parser.add_option("-e", "--export", dest="tables", metavar="DB | DB.TABLE", default=[], action="append", type="string")
     parser.add_option("--fields", dest="fields", metavar="<FIELD>,<FIELD>...", default=None, type="string")
     parser.add_option("--keys", dest="keys", metavar="<KEY>,<KEY>...", default=None, type="string")
+    parser.add_option("--index", dest="index", metavar="<INDEX>", default=None, type="string")
     parser.add_option("--clients", dest="clients", metavar="NUM", default=3, type="int")
     parser.add_option("-h", "--help", dest="help", default=False, action="store_true")
     parser.add_option("--debug", dest="debug", default=False, action="store_true")
@@ -128,6 +131,13 @@ def parse_options():
         res["keys"] = None
     else:
         res["keys"] = options.keys.split(",")
+
+    if options.index is None:
+        res["index"] = None
+    elif options.keys is None:
+        raise RuntimeError("Error: the '--index' option used without '--keys'")
+    else:
+        res["index"] = options.index
 
     # Get number of clients
     if options.clients < 1:
@@ -195,7 +205,7 @@ def write_table_metadata(progress, conn, db, table, base_path):
 # This is called through rdb_call_wrapper and may be called multiple times if
 # connection errors occur.  In order to facilitate this, we do an order_by by the
 # primary key so that we only ever get a given row once.
-def read_table_into_queue(progress, conn, db, table, pkey, keys, task_queue, progress_info, exit_event):
+def read_table_into_queue(progress, conn, db, table, pkey, keys, index, task_queue, progress_info, exit_event):
     read_rows = 0
     table = r.db(db).table(table)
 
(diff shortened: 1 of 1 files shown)


mglukhovsky:
@mbrock -- thanks for the pull request! Could you please sign the CLA if you haven't already (http://rethinkdb.com/community/cla/)?

cc'ing @Tryneus to take a look at this.

mbrock:
@mglukhovsky Sure, no problem.

drivers/python/rethinkdb/_export.py
```diff
@@ -44,6 +44,9 @@ def print_export_help():
     print("  --format (csv | json)            format to write (defaults to json)")
     print("  --fields FIELD,FIELD...          limit the exported fields to those specified")
     print("                                   (required for CSV format)")
+    print("  --keys KEY,KEY...                limit the exported documents by primary key")
```
Tryneus: This doesn't make it very clear what is going to happen - how about "only export documents matching these keys"

drivers/python/rethinkdb/_export.py
```diff
@@ -122,6 +127,18 @@ def parse_options():
...
 
+    if options.keys is None:
+        res["keys"] = None
+    else:
+        res["keys"] = options.keys.split(",")
+
+    if options.index is None:
+        res["index"] = None
+    elif options.keys is None:
+        raise RuntimeError("Error: the '--index' option used without '--keys'")
```
Tryneus: Probably clearer as "the '--index' option cannot be used without '--keys'"

drivers/python/rethinkdb/_export.py
```diff
@@ -188,12 +205,21 @@ def write_table_metadata(progress, conn, db, table, base_path):
...
+    if progress[0] is not None:
+        table = table.between(progress[0], None, left_bound="open")
+
+    if keys is None:
+        table = table.order_by(index=pkey)
+    elif index is not None:
+        table = table.get_all(*keys, index=index).order_by(pkey)
     else:
-        cursor = r.db(db).table(table).between(progress[0], None, left_bound="open").order_by(index=pkey).run(conn, time_format="raw", binary_format='raw')
+        table = table.get_all(*keys).order_by(pkey)
```
Tryneus: This should be `order_by(index=pkey)` for efficiency (not that this query will realistically be returning a ton of rows).

drivers/python/rethinkdb/_export.py
```diff
@@ -188,12 +205,21 @@ def write_table_metadata(progress, conn, db, table, base_path):
...
-        cursor = r.db(db).table(table).order_by(index=pkey).run(conn, time_format="raw", binary_format='raw')
+    table = r.db(db).table(table)
+
+    if progress[0] is not None:
+        table = table.between(progress[0], None, left_bound="open")
+
+    if keys is None:
+        table = table.order_by(index=pkey)
+    elif index is not None:
+        table = table.get_all(*keys, index=index).order_by(pkey)
```
Tryneus: I *think* this should be `order_by(index=index)`, but I need to verify that this can continue where it left off without missing any rows if the connection is interrupted.

drivers/python/rethinkdb/_export.py
mbrock: I tried that, but it turns out that RethinkDB actually doesn't support indexed `order_by` on `get_all` queries (or indeed any non-table selection).

Doing the `get_all` after the `order_by` doesn't work either, because `get_all` is defined only on tables.

drivers/python/rethinkdb/_export.py
mbrock: Hmm. Consider the case when we are exporting the `employees` table with `--keys 1 --index company_id`. In this case, I don't think it makes sense to `order_by(index="company_id")`. But ordering on the primary key of `employees` seems like it should work.

Tryneus: Actually, this gets a lot more complicated when trying to bulletproof these exports for connection loss and resumption - working on a set of queries...

drivers/python/rethinkdb/_export.py
chrisvariety: You can chain a `order_by` after a `between` if they are using the same index (which, I imagine, is why the previous query was using `between`)

mbrock: @chrisvariety I'm not sure I understand. The query uses `between` when there is a `progress[0]` in order to resume after reconnection. Then we optionally do a `get_all` to specify a subset of documents, and then an `order_by` on the primary key (using an index only when selecting all documents). Processing the documents in primary key order ensures that the export progress is predictable.

As far as I can see, the only issue is that `order_by` cannot use the index when we are selecting a subset using `get_all`, but this is inescapable.

drivers/python/rethinkdb/_export.py
danielmewes: Did you get around to looking into this @Tryneus ?

Tryneus: Sorry, guys, lost track of this.  It should be possible to get this to work, but it requires 6 different queries based on whether `keys` or `index` are set, and whether we're resuming after a connection interruption.  I would like to merge this functionality into `next`, but it requires some careful handling of the queries so I'd much rather take care of it myself.

Would it be alright if I branched this and merged it in separately?

srh:
Closing.  Reasons:

1. This is a very old PR and I don't know if it still works.
2. It didn't get merged for some reason.
3. We shouldn't turn the export script command into some peculiar get_all query language thing.  Users can make their own script that contains a custom query and dumps it.

PR #3733, opened 2015-02-05
javscript: Emit connection error

tundrax:
`timeout` configuration option has no effect in case of connection refused error (server down). Timeout is cleared in `@rawSocket.once` and no error ever emitted. Since `@rawSocket.on "close"` event is called directly after `@rawSocket.on "error"` we can just emit a connection error.

### DIFF
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -399,7 +399,7 @@ class TcpConnection extends Connection
         @rawSocket.on 'error', (args...) =>
             if @isOpen()
                 @close({noreplyWait:false})
-            @emit 'close'
+            @emit 'error', new err.RqlDriverError "Could not connect to " + @host + ":" + @port
 
         @rawSocket.on 'close', =>
             if @isOpen()
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -396,10 +396,7 @@ class TcpConnection extends Connection
 
             @rawSocket.on 'data', handshake_callback
 
-        @rawSocket.on 'error', (args...) =>
-            if @isOpen()
-                @close({noreplyWait:false})
-            @emit 'error', new err.RqlDriverError "Could not connect to " + @host + ":" + @port
+        @rawSocket.on 'error', (err) => @emit 'error', err
 
         @rawSocket.on 'close', =>
             if @isOpen()


danielmewes:
Hi @tundrax , thanks a lot for your pull request.

For legal reasons, we require contributors to sign a CLA before we can merge a pull request. You can find all the details and sign it online at http://rethinkdb.com/community/cla/.

@deontologician can you please take a look at @tundrax's patch?

deontologician:
Yes, taking a look now. @tundrax thanks for the PR, I'll try to review it reasonably quickly, but it may be a day or two

tundrax:
Hi @danielmewes, @deontologician. Just signed the CLA, thank you. I will try to check if the same problem exists with other drivers.

drivers/javascript/net.coffee
```diff
@@ -399,7 +399,7 @@ class TcpConnection extends Connection
         @rawSocket.on 'error', (args...) =>
             if @isOpen()
                 @close({noreplyWait:false})
-            @emit 'close'
+            @emit 'error', new err.RqlDriverError "Could not connect to " + @host + ":" + @port
```
deontologician: OK, thanks for being patient, I finally understand the driver code in depth enough to form an educated opinion on this PR.

I agree the `"error"` event should be raised again instead of `"close"`. But I think we should make three changes to the patch here:

1. Change the callback from taking `(args...)` to just `(err)`, since Node only passes an error object, not a bunch of args (this was an oversight in the existing code)
2. Instead of passing a new `RqlDriverError` as the argument to `emit`, just pass `err`. In the Connection superclass, it registers a listener for `"error"` that will raise this same RqlDriverError, so it's redundant here.
3. Remove the `isOpen()` check in here, since (as you mention) the `"close"` event gets emitted immediately afterwards, it's redundant.

The final code should just be like:

```coffee
@rawSocket.on('error', (err) => @emit('error', err))
```

tundrax:
Thank you @deontologician. Updated with redundant checks removed.

deontologician:
Thanks!

AtnNn:
@deontologician is this a breaking change or a bug fix? If it is a bug fix, we should cherry-pick it into the v1.16.x branch and do a release of the JavaScript driver.

stuartpb:
Bugfix, see #3807.

deontologician:
@AtnNn as @stuartpb says, it's a bugfix that needs to be ported. A number of users have run into it, so it's definitely worth backporting

AtnNn:
@stuartpb I've released a new version of the JavaScript driver (1.16.1) with this change.

stuartpb:
Thanks, I've updated the dependencies of my projects that use connections.

PR #3770, opened 2015-02-12
take care of upper-case command line arguments

AmeenAli:
take care of upper-case command line arguments
such as " CREATE , IMPORT , ExpOrT ....

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1,6 +1,6 @@
 // Copyright 2010-2013 RethinkDB, all rights reserved.
 #include <sys/resource.h>
-
+#include <algorithm>
 #include <set>
 
 #include "clustering/administration/main/command_line.hpp"
@@ -28,7 +28,7 @@ int main(int argc, char *argv[]) {
 
     } else {
         std::string subcommand = argv[1];
-
+std::transform(subcommand.begin(), subcommand.end(), subcommand.begin(), ::tolower)
         if (subcommand == "create") {
             return main_rethinkdb_create(argc, argv);
         } else if (subcommand == "serve") {
@@ -59,6 +59,7 @@ int main(int argc, char *argv[]) {
                 return 0;
             } else if (argc == 3) {
                 std::string subcommand2 = argv[2];
+                std::transform(subcommand2.begin(), subcommand2.end(), subcommand2.begin(), ::tolower)
                 if (subcommand2 == "create") {
                     help_rethinkdb_create();
                     return 0;
--- a/src/main.cc
+++ b/src/main.cc
@@ -28,7 +28,7 @@ int main(int argc, char *argv[]) {
 
     } else {
         std::string subcommand = argv[1];
-std::transform(subcommand.begin(), subcommand.end(), subcommand.begin(), ::tolower)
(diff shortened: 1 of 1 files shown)


danielmewes:
Hi @AmeenAli, thank you for your pull request.

May I ask why you need this?
I don't think it's common in the Unix world to ignore the case for command line arguments.

danielmewes:
@AmeenAli Please also note: For legal reasons, we require contributors to sign a contributor agreement before we can merge a contribution into our code base. You can do this easily at http://rethinkdb.com/community/cla/

AmeenAli:
i do ignore the case for command line arguments sometimes ..
as i said now you can handle both upper and lower case commands..
signing done

danielmewes:
I'm still not convinced that we should do this. While I don't have a strong argument for why it would be bad, I'm not aware of any other Unix tools that ignore the case of arguments like this.

As another example of a tool that takes an argument to specify which sort of action you want to perform:
```
$ git help
[prints help]
$ git Help
git: 'Help' is not a git command. See 'git --help'.

Did you mean this?
	help
$ git HELP
git: 'HELP' is not a git command. See 'git --help'.
```

mglukhovsky:
Interestingly, `git` seems to support mixed case for `git pull`, but not `git status`:

```
$ git PULL
Already up-to-date.
$ git PuLL
Already up-to-date.
$ git STATUS
fatal: cannot handle STATUS as a builtin
```

AmeenAli:
so whats the problem if you ignore the case of arguments!?
 thats would make it easy to use ..

danielmewes:
@mglukhovsky That's interesting. It doesn't work on newton:
```
$ git PULL
git: 'PULL' is not a git command. See 'git --help'.
```

Let's see what other people are thinking. Pinging @AtnNn, @larkost, @gchpaco

gchpaco:
> Interestingly, git seems to support mixed case for git pull, but not git status:

Are you doing that on OS X?  `git pull` invokes `git-pull` which is a thing on the filesystem, whereas the builtins may not be the same thing.

gchpaco:
> @mglukhovsky That's interesting. It doesn't work on newton:

Yeah, that sounds like HFS+ at work.

danielmewes:
@AmeenAli One small problem with it is that it would be inconsistent with the other command line arguments `rethinkdb` takes, such as `--help`, `--cache-size`, `` etc.
We could make those case-insensitive as well, but that would definitely be at odds with other commands then.

Some Unix tools even assign different meanings to certain single-letter arguments depending on whether they are in lower or upper case.
For example GNU `ls`:
```
-c     with  -lt:  sort  by, and show, ctime (time of last modification of file status information) with -l: show ctime and sort by name otherwise: sort by ctime, newest first

-C     list entries by columns
```

mglukhovsky:
@gchpaco: yup, on OS X, so that makes sense. I was surprised, I generally expect case sensitivity to be meaningful (thanks for the explanation!)

danielmewes:
@AmeenAli Sorry this didn't work out. Just wanted to thank you again for opening the pull request! Would be happy to see more patches from you in the future.

PR #3782, opened 2015-02-13
release notes for 1.16.2

AtnNn:
@danielmewes and @mglukhovsky could you please review?

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,3 +1,16 @@
+# Release 1.16.2 (Stand By Me)
+
+Released on 2015-02-16
+
+Bug fix update.
+
+* Fixed a bug in `r.range` that caused a crash in debug mode (#3767)
+* Fixed a race condition in the implementation of `.order_by.limit.changes` (#3765)
+* Fixed a build errror that caused really slow `r.js` performance (#3757)
+* Removed spurious comma in the Data Explorer (#3730)
+
+--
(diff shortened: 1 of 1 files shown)


danielmewes:
Looks good to me.

NOTES.md
```diff
@@ -1,3 +1,16 @@
+# Release 1.16.2 (Stand By Me)
+
+Released on 2015-02-16
+
+Bug fix update.
+
+* Fixed a bug in `r.range` that caused a crash in debug mode (#3767)
```
Tryneus: This would also cause queries to fail in release mode - so I would just say "that caused query failures"

timmaxw:
"error" is misspelled.

NOTES.md
```diff
@@ -1,3 +1,16 @@
...
+* Fixed a race condition in the implementation of `.order_by.limit.changes` (#3765)
+* Fixed a build errror that caused really slow `r.js` performance (#3757)
```
timmaxw: This might be nitpicking, but I'd replace "really" with "very"; "really" sounds too informal to me.

mglukhovsky:
Agreed with both @tim and @Tryneus' comments. Otherwise, looks good.

AtnNn:
Thanks. Merged in 1427bd0.

PR #3814, opened 2015-02-21
Fix #2891

### LINKED ISSUE #2891: Heartbeat timeout message should be more explicit about the connection
When the communications fail between two servers the current message looks like:
```error: Heartbeat timeout, killing connection to peer 127.0.0.1```

The problem is that in this case (testing) all of the servers are on `127.0.0.1`, so that does not actually designate a server. We should probably include the port information as well.

eliangidoni:


danielmewes:
Great catch Elian. Thank you!
I suspect someone was confusing it with the system library `::getpeername()` which returns 0 on success.

danielmewes:
Will ship in RethinkDB 2.0

PR #3816, opened 2015-02-22
Fix #2117

### LINKED ISSUE #2117: log_serializer_t::shutdown_in_one_shot variable cleanup
Thanks to fixes in #2113, the variable log_serializer_t::shutdown_in_one_shot is useless.  We never shutdown in one shot.  Nor should we -- it was a bad idea to have a complicated dual-control-flow interface in the first place.  Somebody get a phone and call 2010...

eliangidoni:


src/serializer/log/log_serializer.cc
```diff
@@ -874,7 +867,7 @@ bool log_serializer_t::next_shutdown_step() {
 
         // Don't call the callback if we went through the entire
```
danielmewes: This comment doesn't apply anymore after these changes. Can you delete it?

danielmewes:
Thank you for the patch @eliangidoni !

A few small things to change before we can merge:
- `bool log_serializer_t::shutdown(cond_t *cb)` never returns `true`. We should make it void, so it becomes clear that it always pulses `cb`.
- same for `bool log_serializer_t::next_shutdown_step()`. It should be void.

eliangidoni:
@danielmewes Just fixed, review it. :)

danielmewes:
Looks good, thank you @eliangidoni !

PR #3823, opened 2015-02-23
node: use semver correctly

jonathanong:
If bluebird releases a version 3, then a breaking change was introduced, and there's a good chance your entire driver could break. Luckily, the maintainer of bluebird is actually good so I doubt this would actually happen.


danielmewes:
Thanks for the new pull request @jonathanong !

One more thing: Would you mind signing our CLA http://rethinkdb.com/community/cla/?

Once you've done that please let me know and we can merge this in.

jonathanong:
done

AtnNn:
Thanks!

stuartpb:
Wouldn't this be better as `"^2.3.2"`?

Issue #3856, opened 2015-03-01
Launch Data Explorer directly to table


joshhunt:
Just a small suggestion to make the Admin Console a little bit easier to use - It would be really handy if in the Tables view, there was a button next to each table that would launch the Dataexplorer, prefilled in with a query for that table, like `r.db('myApi').table('users')`

joshhunt:
If I were to make this change myself, I would probably put a button on each table row that shows just on hover. When clicked, it would set the localStorage value `current_query` to the prefillled in query (just `r.db().table()`) and then navigate to the data explorer view.

Thoughts? If I were to approach it this way, would the PR be accepted?

deontologician:
Very similar to #3509. It's not scheduled immediately, so pull requests are definitely welcome!

danielmewes:
Let me actually merge this with #3509 since they are really similar.

Issue #3509, opened 2015-01-07
Add a "work with this table in the data explorer" button to table pages


deontologician:
The idea being that it loads the data explorer tab immediately with `r.db(appropriate_db).table(current_table)` so the user can mess with it

mglukhovsky:
:+1:, good idea.

danielmewes:
Sounds good (assuming we can find a good place to put it).

danielmewes:
Adding suggestion from #3856: 
> If I were to make this change myself, I would probably put a button on each table row that shows just on hover. When clicked, it would set the localStorage value current_query to the prefillled in query (just r.db().table()) and then navigate to the data explorer view.

srh:
Fixed by #3857.

Issue #3825, opened 2015-02-24
Table listing redesign


deontologician:
The current table listing is sub-optimal. It's grown a lot of buttons, and it's unnecessarily complicated to do some normal kinds of operations.

Case in point:

Deleting a table:

1. select the checkbox by the table
2. go to the top right and click "delete selected tables"
3. A modal pops up, asking you to confirm.
4. Click ok on the modal
5. the table disappears from the list
6. A success notification pops up saying you deleted the table.
7. Click the X on the notification to get rid of it.

What should happen:

1. Click a "delete table" icon on the table row.
2. Some small confirmation action (no modal)
3. The table disappears from the list.

We should have similarly easy actions for stuff like:

- renaming a table
- creating a new table/database

The big buttons aren't great either, we need some icons for them to make them stand out. Right now there's just a ton of huge grey buttons with nothing really distinguishing them.

mglukhovsky:
Agreed. The biggest change we can make is inline actions (e.g. renaming happens inline, deletions happen inline, etc.). This would require creating some visual patterns, but it'll cut down on the chrome and set of steps you have to go through for basic tasks.

danielmewes:
That sounds great. I don't think modals are particularly terrible, but I wouldn't be sad to see some of them go away either.
Adding small icons for performing specific actions to each individual row would be great, as would adding icons to the large buttons.

deontologician:
#708 is relevant, the ui suffers a bit because there are so many indistinguishable buttons

marshall007:
One thing I would really like to see added to the tables view is aggregated secondary index construction stats/progress.

deontologician:
As part of this, we should allow renaming databases again

PR #3857, opened 2015-03-02
Added explore button to each database table row

joshhunt:
My quick attempt at completing issue #3856 and #3509. 

screen shot 2015-03-02 at 2 56 20 pm

Each row gets a new 'Explore' button, visible only on row hover (to avoid the clutter of lots of repeating buttons).
Clicking the button navigates to the data explorer, profiled in with the query `r.db('databaseName').table('tableName')`

I wasn't sure of the best way to do this, so I went for the simplest approach:

``` coffee
explore_table: =>
  window.localStorage.current_query = JSON.stringify "r.db('#{@model.get('db')}').table('#{@model.get('name')}')"
  location.hash = 'dataexplorer'
```

Because it relies on localStorage to refill in the command, the button is only displayed in browsers that support localStorage.

Thoughts?

### DIFF
--- a/admin/static/coffee/tables/index.coffee
+++ b/admin/static/coffee/tables/index.coffee
@@ -281,8 +281,16 @@ module 'TablesView', ->
     class @TableView extends Backbone.View
         className: 'table_container'
         template: Handlebars.templates['table-template']
+
+        events:
+           'click button.explore-table': 'explore_table'
+
         initialize: =>
             @listenTo @model, 'change', @render
+
+        explore_table: =>
+          window.localStorage.current_query = JSON.stringify "r.db('#{@model.get('db')}').table('#{@model.get('name')}')"
+          location.hash = 'dataexplorer'
 
         render: =>
             @$el.html @template
@@ -295,6 +303,7 @@ module 'TablesView', ->
                 replicas: @model.get 'replicas'
                 replicas_ready: @model.get 'replicas_ready'
                 status: @model.get 'status'
+                displayExploreButton: window.hasOwnProperty('localStorage')
             @
 
         remove: =>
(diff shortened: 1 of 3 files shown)


joshhunt:
My main concern with my attempt is if `@model.get('db')` or `@model.get('name')` contains characters which need to be escaped. Is that possible, or is it basically only alphanumeric characters which can appear there?

deontologician:
Will check this out today. Thanks for the PR

deontologician:
I like it, the functionality is nice and clean. The buttons only showing up on hover is probably not the way to go though. We don't use that mechanism anywhere else in the ui. @mglukhovsky also mentioned we try to remove as much as we can from the table listing to keep it uncluttered.

What I think needs to happen:

1.  move the "Explore" button to the top of the table page (next to ", and leave it off the table listing page for now. We have an issue open for redesigning the table listing page in the medium-near future (#3825), so we'll probably revisit whether we can make it work at that point.
2. Instead of setting location.hash directly, use the backbone router like: ```main_view.router.navigate("#dataexplorer")```

danielmewes:
@joshhunt Are you still interested in updating the PR? (in particular moving the "Explore" button to the top of the individual table page as @deontologician described)

We can try to use some of your code as a basis for our own modifications, but I'm not sure when we'll get to that.

Had you signed our CLA http://rethinkdb.com/community/cla/ btw?

bsharpe:
This would be a great addition... hope it makes it :)

srh:
Fixed merge conflicts, followed @deontologician's advice about navigation, and cherry-picked into next as of 05be59f20bbe17745f23e4c411adfc2e15c3d656.  Thank you!

PR #3867, opened 2015-03-03
fix `__builtins__` field access

robert-zaremba:
Fields from `__builtins__` should be accessed using a field syntax (`a.b`). 
The `__builtins__['some_field']` syntax is CPython related and not supported by PyPy. The standard way to do it is  `__builtins__.some_field` - this works with CPython, PyPy and other Python implementations.

### DIFF
--- a/drivers/python/rethinkdb/__init__.py
+++ b/drivers/python/rethinkdb/__init__.py
@@ -6,7 +6,7 @@ from .errors import *
 from .ast import *
 from . import docs
 
-class r(__builtins__['object']): # defends against re-importing obscuring object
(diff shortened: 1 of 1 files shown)


danielmewes:
Thank you for your pull request @robert-zaremba !
Had you signed our CLA before? If not, you can do this online at http://rethinkdb.com/community/cla/ 

@gchpaco could you take a look at the patch please?

robert-zaremba:
@danielmewes - I've just signed. Thanks.

danielmewes:
@robert-zaremba Sorry this took so long. This was actually fixed by an independent change in the meantime and should be working with the 1.16.3 driver and later.

PR #3905, opened 2015-03-07
config options with keys as String

aderyabin:
Currently it does not work in this case
```ruby
r.connect('host' => 'localhost', 'port' => 28015)
```
because keys of config are String.
I suggest to convert the keys to Symbol.

It useful for case if connection config is located in yaml file, for example
```ruby
r.connect YAML.load_file('config/database.yml')
```

and `config/database.yml` is

```yaml
  db: 'test'
  host: '127.0.0.1'
  port: 28015
```

### DIFF
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -141,9 +141,10 @@ module RethinkDB
         @abort_module = Faux_Abort
       end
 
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks @aderyabin .
Pinging @mlucy

mlucy:
Looks good to me.  Merging.

PR #3906, opened 2015-03-07
Avoid self where not required

aderyabin:
`self` is only required when calling a self write accessor [link](https://github.com/bbatsov/ruby-style-guide#no-self-unless-required)


danielmewes:
Thanks again!

@mlucy

mlucy:
Looks good to me.  Thanks!

PR #3907, opened 2015-03-07
Add timeout option to r.connect in Ruby driver (Closes #1666)

### LINKED ISSUE #1666: Ruby driver - r.connect doesn't have a timeout
The Python and JavaScript driver have a timeout argument for `r.connect` but the Ruby driver does not.
We should eventually fix that.

Tagging as backlog as it doesn't seem really important for now.

aderyabin:


danielmewes:
Thanks!

Pinging @mlucy

mlucy:
Looks good to me.  Thanks again!

Issue #3908, opened 2015-03-07
Web UI: Modify query return limit in Web UI Settings


justinmealey:
I'd like to change the (query) `limit` constant in `dataexplorer.coffee` to be a user-configurable variable stored in local storage, similar to `check_update`.  This would be added to the Web UI settings next to automatically looking for new updates, and would just be a simple user input taking an integer.

deontologician:
I think this is a good idea. It's not a super high priority right now, but if you submit a PR we could probably get it into the 2.1 release

justinmealey:
Awesome, will start working on it.

PR #3909, opened 2015-03-08
#3908 Added modifiable query limit to local storage

justinmealey:


justinmealey:
Going to retract this and add error handling for the input field.

PR #3910, opened 2015-03-08
#3908 Added modifiable query limit to local storage

justinmealey:


danielmewes:
Thank you for the pull request @justinmealey !

@deontologician Can you take a look at the patch please?

@justinmealey Unless you've done so already, would you mind signing our CLA? http://rethinkdb.com/community/cla/ We need this for legal reasons before we can accept a pull request.

deontologician:
@justinmealey  This looks good overall, I think it just needs a couple changes:

1. Wrap up the "localStorage or 40" check up into a single method, right now the logic is duplicated a lot
2. Move the option into the dataexplorer options template `admin/static/handlebars/dataexplorer-options-template.handlebars`. The option panel it's in now is for overall webui options, and this only applies to the data explorer.

Other than that this looks good, it'll make a nice addition

chrisvariety:
Just a note I don't think you'll be able to set this to e.g. 100,000 and expect everything to work... at some point the browser will simply hang or crash trying to display that much data, which may end up looking like a Web UI bug in the future. Not sure if an upper limit makes sense or a warning or what, but figured I would bring it up.

Also, you should *always* specify a radix when using `parseInt`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt for more details.

justinmealey:
Huh, did not know that about specifying the radix.

Signed the CLA.

Will make the appropriate changes and resubmit.

justinmealey:
@deontologician query_limit is stored as a data explorer option and pulls from state, and the check is only made upon changing.

@chrisvariety since time increase from large row returns is completely subjective to the data returned, I just opted to put a note next to the input box that lets the user know that returning more rows will increase the load time.  Also, changed parseInt to use base 10 radix.

deontologician:
This looks good, seems to work fine in the webui. Thanks for the contribution @justinmealey !

PR #3915, opened 2015-03-10
export: expose csv module's "delimiter" param

elifarley:


drivers/python/rethinkdb/_export.py
```diff
@@ -122,6 +124,11 @@ def parse_options():
     else:
         res["fields"] = options.fields.split(",")
 
+    if options.delimiter is not None and options.format != "csv":
+        raise RuntimeError("Error: You can only specify a delimiter when writing in CSV format.")
+
+    res["delimiter"] = ',' if options.delimiter is None else options.delimiter
```
Tryneus: Similar code in `_import.py` verifies that the `delimiter` option is at most one character, and also allows for "\t" to be specified.

Tryneus:
Looks good, although I had a comment regarding the validation of the parameter.  Thanks for the PR, @elifarley!

elifarley:
@Tryneus I have an improved version of this change at another PR at #3925

Not sure if this was the right way to provide an improved version of a PR though (I'm used to the old way of sending patches to a mailing list).

danielmewes:
@elifarley In general you can just commit & push additional changes to your branch and they will be added to the pull request automatically.

Issue #3916, opened 2015-03-10
Expose csv module's "delimiter" param for the export command


elifarley:
Sometimes it's convenient to specify a custom field delimiter when exporting to CSV format.

danielmewes:
That makes total sense.

(@Tryneus)

danielmewes:
Closed by https://github.com/rethinkdb/rethinkdb/pull/3925#issuecomment-83158998 , thank you @elifarley !
Unfortunately it didn't make it into RethinkDB 2.0, but will ship with RethinkDB 2.1.

PR #3925, opened 2015-03-11
export: expose csv module's "delimiter" param

elifarley:
This time the change is more in line with the code from _import.py

Perhaps this could be applied to the v1.16.x branch too...


danielmewes:
Thank you @elifarley !

@Tryneus can you check this when you have time?

Linking https://github.com/rethinkdb/rethinkdb/issues/3916 here for completeness.

Tryneus:
Ah, right, lost track of this.  Looks good @elifarley, merging it.  Thanks for the help!

Issue #3100, opened 2014-09-25
Add a __version__ variable to the Python driver


larkost:
For remote troubleshooting it would be very helpful for the Python driver to have a `__version__` variable, making it easy to figure out. This would probably require having the makefile insert this into at least `__init__.py` based on the output of `scripts/gen-version.sh`. It is probably overkill to put it into each file. The only issue there is that we have driver versions separate from the server version, so that will need to be resolved.

The same thing should also be done for the Ruby, and the JavaScript driver if it supports it.

danielmewes:
Closed by https://github.com/rethinkdb/rethinkdb/pull/3971 thanks to @yograterol

PR #3948, opened 2015-03-18
Added version var to Python Driver

yograterol:
QuickFix https://github.com/rethinkdb/rethinkdb/issues/3100


danielmewes:
Hi @yograterol, thank you for your pull request!

This solution isn't very practical unfortunately since it requires manual updating for every new version.
What we need is something similar to this line from the web UI makefile https://github.com/rethinkdb/rethinkdb/blob/next/mk/webui.mk#L89 that injects the RethinkDB version as part of the make process.

Have you signed our CLA http://rethinkdb.com/community/cla/ yet? We'll need that before we can merge a pull request.

AtnNn:
Having  `r.__version__` would be great, but as @danielmewes says, the version is already specified elsewhere. In this case, it's in `setup.py`.

Perhaps you could do something like put `__version__ = "1.16.0-0"` in `rethinkdb/version.py`, using `execfile` to access it from `setup.py` and `import` to access it from `__init__.py`

AtnNn:
PEP 396 mentions having a `setup.cfg` file with a `version-from-file` field. I wonder if that would work.

yograterol:
@danielmewes @AtnNn You are right, well I can do version.py and setup.cfg, It can be use with automatic rethinkdb build. 

BTW, I will sing the CLA.

yograterol:
@danielmewes @AtnNn I did a change for use the version from setup, but exist a solution with metadata https://github.com/openstack-dev/pbr.

AtnNn:
@yograterol I would prefer not introducing a dependency on setuptools or pbr.

yograterol:
@AtnNn ok, here you go my changes

AtnNn:
@yograterol this change depends on setuptools.

yograterol:
Close this PR, in the rebase get changes than I don't want.

Issue #3061, opened 2014-09-16
Array over size limit should provide more guidance


neumino:
If the table `test` has more than 100k documents, this will return an error:

```
r.db('test').table('test').orderBy('id')
```

The error being:
```
RqlRuntimeError: Array over size limit `100000` in:
r.db("test").table("test").orderBy("id")
```

We should mention the `arrayLimit` option or the fact that they can use an index.

llambda:
Yes e.g. orderBy({index: r.desc('id')}) would be helpful to know from the error message; this caught me up.

gyelwande:
The issue is fixed.

danielmewes:
@gyelwande I might have missed this. Which change fixed this?

PR #3977, opened 2015-03-25
Add mention of arrayLimit option and indexing to error message

t3hmrman:
Minimal fix to issue #3061 , Just added a message mentioning arrayLimit option and using an index.

### DIFF
--- a/src/rdb_protocol/error.hpp
+++ b/src/rdb_protocol/error.hpp
@@ -119,13 +119,13 @@ private:
 #define rcheck_array_size_datum(arr, limit, type) do {                  \
         auto _limit = (limit);                                          \
         rcheck_datum((arr).size() <= _limit.array_size_limit(), (type),  \
-                     strprintf("Array over size limit `%zu`.",          \
+                     strprintf("Array over size limit `%zu`. To change the number of allowed documents, modify the 'arrayLimit' option, or use an index.",          \
                                _limit.array_size_limit()).c_str());     \
     } while (0)
 #define rcheck_array_size(arr, limit, type) do {                        \
         auto _limit = (limit);                                          \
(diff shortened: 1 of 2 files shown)


t3hmrman:
Not a substantial change, but would be great if someone looked over the prose -- figured I could at least knock this easy ticket out while I was visiting the office :smile:

danielmewes:
Thanks @t3hmrman !
Would you mind signing our CLA http://rethinkdb.com/community/cla/ ? We require this before we can merge a pull request.

About the change itself: The option should be referred to as `array_limit` not `arrayLimit`. The camel case name is specific to the JavaScript driver (and maybe a few third part ones), while all of our drivers accept the underscore form `array_limit`. Other than that it looks good to me.

danielmewes:
Oh also it should be "elements" instead of "documents" I think, since the array can for example be an array of numbers or strings.

t3hmrman:
Hey @danielmewes , I just signed the CLA, and made the "elements" change! 

Also, the strings are kind of long -- is there a style guide?

danielmewes:
Ah yes, good you mention it. We have an 89 character line length limit. Mind breaking the strings up?

t3hmrman:
Just did some of the string splitting, but some tests are currently failing so looking into it (whether I caused them or not) -- would suggest not merging this in yet

danielmewes:
@t3hmrman thanks. Which tests did you run? It's possible that some of them are testing for a specific error message and need to be adjusted.
The relevant ones are defined in `test/rql_test/src` and you can run them through the `test-runner` script in `test/rql_test`.

t3hmrman:
Closing since I haven't worked on this in a super long time :<

Issue #1771, opened 2013-12-13
Memory table


wojons:
Just a pure raw memory based table. I know that I can use tmpfs for memory based tables but A built in feature would be a nice idea that also allows capping the size of the table by number of docs or by memory usage.

danielmewes:
To copy over the conclusion from #4003:
> I think it might be better to make the cache never evict any current pages, and have the serializer just be a "black hole" serializer that throws away all writes and crashes if you try to read from it.

analytik:
+1 for this feature.

Our use case is that we want to save I/O on non-SSD nodes. Maximum size and related complications are not a concern, as the table would be basically a view for another table, of which historical data creates a big part (growing, let's say, 100k~500k documents per day), and the "active" data is only 10~100k documents.

Technically, of course, it's not necessary to have such table in our use case, but it seems easier to write to two tables than to risk having queries get progressively slower as the data grows.

raarts:
My use case for memory tables is keeping status for short lived connections, and distribute them like streams. OP said this could be done with tempfs. Is that possible for one table only?

danielmewes:
@raarts You could create the table, note the table's UUID (e.g. from `r.db('rethinkdb').table('table_config')`) shut down the server, move the file for that table to the tempfs, create a symlink from the rethinkdb data folder to the table file on tempfs, and restart the server.

I'm not sure if this will still work properly though once the table file gets deleted after a restart.

adamierymenko:
@danielmewes What happens if I do that and then reboot and the table file is lost? Will RethinkDB recreate it and (if possible) re-sync its data from other replicas in the cluster?

I think a simple easy way to do memory tables would be to allow us to specify that some tables are stored in tmpfs (/dev/shm on most systems) and have exactly that behavior -- they replicate as normal and on reboot or failure are restored from the rest of the cluster. If perchance the whole cluster loses them, they are recreated as empty.

That would be really useful to decrease write overhead on rapidly changing data. We have an issue with that right now. Performance is fine but over time RethinkDB writes so much it will eventually nuke our SSDs.

adamierymenko:
An even cooler twist would be to be able to specify that memory tables are flushed to disk periodically on a configurable period. This would allow something akin to how Redis works-- everything is in RAM and is periodically flushed to disk.

adamierymenko:
[...]

PR #4003, opened 2015-04-01
memory tables hell yeah

wojons:
This is fully a work in progress but something that should not be taken lightly. There is no ways to select between disk backed and memory backed tables so for testing purposes all tables will be memory tables.

### DIFF
--- a/src/arch/types.cc
+++ b/src/arch/types.cc
@@ -17,10 +17,14 @@ address_in_use_exc_t::address_in_use_exc_t(const char* hostname, int port) throw
 
 file_account_t::file_account_t(file_t *par, int pri, int outstanding_requests_limit) :
     parent(par),
-    account(parent->create_account(pri, outstanding_requests_limit)) { }
+    account(par == nullptr
+            ? nullptr
+            : parent->create_account(pri, outstanding_requests_limit)) { }
 
 file_account_t::~file_account_t() {
-    parent->destroy_account(account);
+    if (parent != nullptr) {
+        parent->destroy_account(account);
+    }
 }
 
 void linux_iocallback_t::on_io_failure(int errsv, int64_t offset, int64_t count) {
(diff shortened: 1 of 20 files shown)


wojons:
related to #1771

danielmewes:
That's amazing @wojons, thanks a lot! I didn't know you were learning C++.
You had already signed the CLA hadn't you?

We'll take a look and let you know what's left to do.

wojons:
Yeah i signed the CLA a while ago. I had the pull request a while ago. I have loved rethinkdb so i had to start relearning C++ a few months ago and thought this would be great.

danielmewes:
We haven't discussed an interface in https://github.com/rethinkdb/rethinkdb/issues/1771 yet, but we should probably expose the memory vs disk table option as an opt arg to `table_create`. Similarly it should show up in `r.db('rethinkdb').table('table_config')`.
(Ping @mlucy and @timmaxw.)

I wonder how this interacts with the current way we manage secondary indexes on replicated/sharded tables. From what I can tell if you restart a server with memory tables like these, those replicas will have empty tables without any secondary indexes (since the indexes are stored in the table itself). We might want to wait for Raft where @timmaxw has changed that before including this.

@wojons I'm curious, have you done any benchmarks with this? I wonder how much faster this is compared to a normal table that's completely in cache.

timmaxw:
We should create one `memory_serializer_t` for each hash-shard; this way, we won't have to switch threads to "read" and "write" the memory serializer. The merger serializer and serializer multiplexer shouldn't be used with a memory serializer, because they add unnecessary overhead.

timmaxw:
Also, it looks like this is keeping two copies of the data in RAM: one in the page cache and one in the memory serializer. That seems suboptimal.

wojons:
@timmaxw yeah it could be better it was rush job to get the idea across.

v3ss0n:
So the memory tables are coming soon? NEAT!

danielmewes:
Revisiting this, I think it might be better to make the cache never evict any current pages, and have the serializer just be a "black hole" serializer that throws away all writes and crashes if you try to read from it.
With the in-memory serializer, there's the problem of keeping the same data in memory twice, once in the serializer and once in the cache.

danielmewes:
Closing this since we're going to implement it differently.

Issue #4011, opened 2015-04-03
The server replies "SUCCESS" and immediately fails with an error


AtnNn:
After a handshake with a wrong protocol version, the server sends back:

```
SUCCESS\0Unrecognized protocol specified: '1074539808'
```

I think it shouldn't send back "SUCCESS" if the magic number is wrong.

danielmewes:
@grandquista fixed this through https://github.com/rethinkdb/rethinkdb/pull/4012. Also backported to v2.0.x.

PR #4012, opened 2015-04-03
Write `SUCCESS\0` after verifying protocol magic number.

grandquista:
Fix for #4011.


danielmewes:
Thank you @grandquista! I added a note to your commit.

grandquista:
I moved the protocol verification so that `SUCCESS` is safe to write before the connection loop starts. I used a switch so that compilers can check for new protocol types in the enumeration.

danielmewes:
Thanks. I'm going to merge this into v2.0.x so I can base the fix for https://github.com/rethinkdb/rethinkdb/issues/4015 on it.

PR #4071, opened 2015-04-16
Add asyncio event loop support

takluyver:
This probably needs quite a bit of work, but it's working for a basic example, and I wanted to communicate early. @techdragon talked me into working on this at the PyCon sprints.

Asyncio is part of the standard library in Python 3.4. It's a rather similar API to tornado, so I copied and pasted the net_tornado code that's already in rethinkdb and modified it until it works. The syntax is only valid on Python 3, but the module shouldn't be imported until you do `r.set_loop_type('asyncio')`, so that shouldn't be a problem.

Here's the asyncio version of the tornado example from your [async drivers blog post](http://www.rethinkdb.com/blog/async-drivers/) (BTW, that example is missing a colon on the `while` line):

```python
import asyncio
import rethinkdb as r

r.set_loop_type("asyncio")

@asyncio.coroutine
def print_changes():
    conn = yield from r.connect(host="localhost", port=28015)
    feed = yield from r.table("tv_shows").changes().run(conn)
    while (yield from feed.fetch_next()):
        change = yield from feed.next()
        print(change)

loop = asyncio.get_event_loop()
loop.run_until_complete(print_changes())
```

### DIFF
--- a/drivers/python/rethinkdb/net_asyncio.py
+++ b/drivers/python/rethinkdb/net_asyncio.py
@@ -28,6 +28,31 @@ def _read_until(streamreader, delimiter):
             break
     return bytes(buffer)
 
+
+def reusable_waiter(loop, timeout):
+    """Wait for something, with a timeout from when the waiter was created.
+
+    This can be used in loops::
+
+        waiter = reusable_waiter(event_loop, 10.0)
+        while some_condition:
+            yield from waiter(some_future)
+    """
+    if timeout is not None:
+        deadline = loop.time() + timeout
+    else:
+        deadline = None
+
+    @asyncio.coroutine
+    def wait(future):
+        if deadline is not None:
+            new_timeout = max(deadline - loop.time(), 0)
+        else:
+            new_timeout = None
+        return (yield from asyncio.wait_for(future, new_timeout, loop=loop))
+
+    return wait
+
 @contextlib.contextmanager
 def translate_timeout_errors():
     try:
@@ -56,10 +81,11 @@ class AsyncioCursor(Cursor):
     @asyncio.coroutine
     def fetch_next(self, wait=True):
         timeout = Cursor._wait_to_timeout(wait)
+        waiter = reusable_waiter(self.conn._io_loop, timeout)
         while len(self.items) == 0 and self.error is None:
             self._maybe_fetch_batch()
             with translate_timeout_errors():
-                yield from asyncio.wait_for(self.new_response, timeout)
+                yield from waiter(self.new_response)
         # If there is a (non-empty) error to be received, we return True, so the
         # user will receive it on the next `next` call.
         return len(self.items) != 0 or not isinstance(self.error, RqlCursorEmpty)
@@ -71,12 +97,13 @@ class AsyncioCursor(Cursor):
 
     @asyncio.coroutine
     def _get_next(self, timeout):
+        waiter = reusable_waiter(self.conn._io_loop, timeout)
         while len(self.items) == 0:
             self._maybe_fetch_batch()
             if self.error is not None:
                 raise self.error
             with translate_timeout_errors():
-                yield from asyncio.wait_for(self.new_response, timeout)
+                yield from waiter(self.new_response)
         return convert_pseudo(self.items.pop(0), self.query)
 
     def _maybe_fetch_batch(self):
--- a/drivers/python/rethinkdb/net_asyncio.py
+++ b/drivers/python/rethinkdb/net_asyncio.py
@@ -1,6 +1,7 @@
 # Copyright 2015 RethinkDB, all rights reserved.
 
 import asyncio
+import contextlib
 import socket
 import struct
 
@@ -27,6 +28,13 @@ def _read_until(streamreader, delimiter):
             break
     return bytes(buffer)
 
+@contextlib.contextmanager
+def translate_timeout_errors():
+    try:
+        yield
+    except asyncio.TimeoutError:
+        raise RqlTimeoutError
+
 
 # The asyncio implementation of the Cursor object:
 # The `new_response` Future notifies any waiting coroutines that the can attempt
@@ -50,7 +58,8 @@ class AsyncioCursor(Cursor):
         timeout = Cursor._wait_to_timeout(wait)
         while len(self.items) == 0 and self.error is None:
             self._maybe_fetch_batch()
-            yield from asyncio.wait_for(self.new_response, timeout)
+            with translate_timeout_errors():
+                yield from asyncio.wait_for(self.new_response, timeout)
         # If there is a (non-empty) error to be received, we return True, so the
         # user will receive it on the next `next` call.
         return len(self.items) != 0 or not isinstance(self.error, RqlCursorEmpty)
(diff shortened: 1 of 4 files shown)


danielmewes:
That's really cool, thanks a lot for the pull request @takluyver !

@Tryneus I think this is one for you to review. Can you take a look when you find the time?

@takluyver Have you signed our CLA yet? http://rethinkdb.com/community/cla/

drivers/python/rethinkdb/net_asyncio.py
```diff
@@ -0,0 +1,236 @@
...
+            self._maybe_fetch_batch()
+            with translate_timeout_errors():
+                yield from asyncio.wait_for(self.new_response, timeout)
+        # If there is a (non-empty) error to be received, we return True, so the
+        # user will receive it on the next `next` call.
+        return len(self.items) != 0 or not isinstance(self.error, RqlCursorEmpty)
+
+    def _empty_error(self):
+        # We do not have RqlCursorEmpty inherit from StopIteration as that interferes
+        # with Tornado's gen.coroutine and is the equivalent of gen.Return(None).
```
Tryneus: This should be `asyncio`, not `Tornado` - and the `gen.Return(None)` thing isn't quite true when using `yield from`.

drivers/python/rethinkdb/net_asyncio.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # Convenience function so users know when they've hit the end of the cursor
+    # without having to catch an exception
+    @asyncio.coroutine
+    def fetch_next(self, wait=True):
+        timeout = Cursor._wait_to_timeout(wait)
+        while len(self.items) == 0 and self.error is None:
+            self._maybe_fetch_batch()
+            with translate_timeout_errors():
+                yield from asyncio.wait_for(self.new_response, timeout)
```
Tryneus: The Tornado implementation uses a deadline rather than a timeout because if this loop is performed multiple times, we don't want to exceed the original timeout - perhaps implementing an analog to `with_absolute_timeout` from `net_tornado.py` would help.

drivers/python/rethinkdb/net_asyncio.py
```diff
@@ -0,0 +1,236 @@
...
+        return RqlCursorEmpty(self.query.term)
+
+    @asyncio.coroutine
+    def _get_next(self, timeout):
+        while len(self.items) == 0:
+            self._maybe_fetch_batch()
+            if self.error is not None:
+                raise self.error
+            with translate_timeout_errors():
+                yield from asyncio.wait_for(self.new_response, timeout)
```
Tryneus: Same comment here about deadlines vs timeouts.

drivers/python/rethinkdb/net_asyncio.py
```diff
@@ -0,0 +1,236 @@
...
+        if self._io_loop is None:
+            self._io_loop = asyncio.get_event_loop()
+
+        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+        self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+
+    @asyncio.coroutine
+    def connect(self, timeout):
+        try:
+            self._socket.connect((self._parent.host, self._parent.port))
```
Tryneus: This call looks like it will block - checking out the `asyncio` API, I think you could combine this with `asyncio.open_connection` below.

drivers/python/rethinkdb/net_asyncio.py
```diff
@@ -0,0 +1,236 @@
...
+    def _get_next(self, timeout):
+        while len(self.items) == 0:
+            self._maybe_fetch_batch()
+            if self.error is not None:
+                raise self.error
+            with translate_timeout_errors():
+                yield from asyncio.wait_for(self.new_response, timeout)
+        return convert_pseudo(self.items.pop(0), self.query)
+
+    def _maybe_fetch_batch(self):
```
Tryneus: This doesn't look correct - it isn't actually attempting to fetch the next batch.

Tryneus:
@takluyver, thanks for putting this up, it looks great, although I made a few comments/suggestions.

takluyver:
Thanks, I was fully expecting that it wouldn't be quite right. I'll respond to comments inline.

@danielmewes I just went and signed the CLA now. The address field mentioned that you might send me something - if that's a t-shirt, then save yourselves the postage, because I already picked up two at PyCon. ;-)

drivers/python/rethinkdb/net_asyncio.py
takluyver: I think the same problem would still arise - `return foo` in a coroutine still works by raising `StopIteration(foo)`, I believe. I'll update the comment.

drivers/python/rethinkdb/net_asyncio.py
takluyver: Aha, good point. I stripped out `with_absolute_timeout` without thinking about use inside loops. It should be easy enough to do that.

danielmewes:
Good to know about the shirts @takluyver :)

drivers/python/rethinkdb/net_asyncio.py
takluyver: asyncio.open_connection() takes either a socket or a hostname+port, but not both. I couldn't see an obvious way to duplicate the `socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)` when creating the socket with asyncio. Do you know of a way, or can it be avoided? Obviously the code could reach in to get the newly created socket object, but you have to dig in a couple of layers, at least one of them `_private`, which felt a bit wrong.

drivers/python/rethinkdb/net_asyncio.py
takluyver: Oh, strange, I appear to have accidentally removed the last line from it in https://github.com/takluyver/rethinkdb/commit/a3a921fe46a3842d49fed27ef0f6991f8f734d5c. Putting that back...

drivers/python/rethinkdb/net_asyncio.py
Tryneus: Not that I've actually tried, but it looks like you could do:

```py
try:
  self._streamreader, self._streamwriter = yield from \
    asyncio.open_connection(self._parent.host, self._parent.port,
                            family=socket.AF_INET, proto=socket.SOCK_STREAM,
                            loop=self._io_loop)
  self._streamwriter.get_extra_info('socket').setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
except Exception as err:
  ...
```

test/rql_test/connections/asyncio_connection.py
```diff
@@ -0,0 +1,923 @@
...
+        while sum(cursor_counts) < 4000:
+            for cursor_index in xrange(self.num_cursors):
+                for read_count in xrange(random.randint(0, 10)):
+                    yield from get_next(cursor_index)
+
+        [cursor.close() for cursor in cursors]
+
+        return (sum(cursor_counts), sum(cursor_timeouts))
+
+    # @asyncio.coroutine
```
Tryneus: Is there a reason these tests are commented out?  Were they taking too long?  If I remember correctly they're the longest-running tests in the suite.

takluyver: Oh, I was commenting out chunks of the tests just so that I could run them quickly. I didn't mean to commit them commented out.

drivers/python/rethinkdb/net_asyncio.py
takluyver: Thanks, that appears to work.

takluyver:
Tests are now uncommented and passing, and I think I've addressed all of the review comments so far.

Tryneus:
Awesome, looks great, @takluyver!  Merging this into `next`, it should make it into the 2.1 release.  I'll try to get more tests in for the alternate `net.py` implementations before then.

Thanks for all the help!

takluyver:
You're welcome :-)

Issue #3158, opened 2014-10-08
Consider allowing drivers to use encrypted connections


danielmewes:
Similar to https://github.com/rethinkdb/rethinkdb/issues/3151 for intra-cluster communication, I believe there are a few things we can do to make working with encrypted client connections much easier.

Here's my proposal:
  Assume the user has set up a tool such as stunnel (https://www.stunnel.org/index.html) on the server to allow incoming SSL connections to RethinkDB's client port.
  
  We could extend the official drivers to support connecting to the server through SSL. From the user's perspective, this would take two things:
  * a flag to `r.connect()` to enable SSL
  * a way to supply a certificate to validate that the client is connecting to the right server (the path to the certificate file could be provided through the same flag that enables SSL actually)

In combination with the auth keys we already have and a firewall that restricts incoming connections to specific IP addresses, this would provide some reasonable level of security for running RethinkDB on unprotected networks.

It's not quite sufficient for just exposing the encrypted client port to the Internet though, since we might expose security vulnerabilities even before the auth key is validated (plus there's a high potential for DoS and brute force attacks).
To make that work, we should also investigate options to use a client-side certificate. I'm not very familiar with how that works with SSL, but it would be great if one could just configure stunnel on the server to check for a certain client certificate, and have another option in our drivers to supply one.

danielmewes:
@joaojeronimo sorry for the late response:
> Also, what's keeping you guys from making rethinkdb serve it's API over TLS connections ?

We'd actually really like to do that. There are different options for this, and we discussed some of them in https://github.com/rethinkdb/rethinkdb/issues/3151#issuecomment-58932433 . While that thread is about intra-cluster connections, we'll probably provide the same thing for client connections as well.

joaojeronimo:
That would be great. Do you guys have this scheduled ?

danielmewes:
@joaojeronimo Not yet. We'll rely on external proxies (like `stunnel`) for now, and try to make its setup as easy as possible by providing adequate documentation and examples.

danielmewes:
Good news: Server-side TLS integration is coming with RethinkDB 2.3 thanks to this pull request https://github.com/rethinkdb/rethinkdb/pull/5381 by @jlhawn

jlhawn:
Cross-posting this for more exposure: I made a [guide to setting up a RethinkDB cluster with native TLS](https://gist.github.com/jlhawn/43f84f86579012470dc9) which uses an experimental build from that aforementioned pull request.

PR #4075, opened 2015-04-17
[ruby] support for ssl encryption and CA verification

jipperinbham:
[Addresses Ruby Driver](https://github.com/rethinkdb/rethinkdb/issues/3158)

The following abstracts the socket connection out a bit to allow for creating an SSL connection to a cluster. In order for this to be used, you'd need a proxy (Haproxy, nginx, etc) sitting in front of the RethinkDB cluster performing the SSL verification.

### DIFF
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -4,6 +4,7 @@ require 'socket'
 require 'thread'
 require 'timeout'
 require 'pp' # This is needed for pretty_inspect
+require 'openssl'
 
 module RethinkDB
   module Faux_Abort
@@ -434,6 +435,7 @@ module RethinkDB
   end
 
   class Connection
+    include OpenSSL
     def auto_reconnect(x=true)
       @auto_reconnect = x
       self
@@ -455,6 +457,7 @@ module RethinkDB
       @auth_key = opts[:auth_key] || ""
       @timeout = opts[:timeout].to_i
       @timeout = 20 if @timeout <= 0
+      @ssl_opts = opts.reject { |k, v| !k.to_s.start_with?('ssl') } if opts.is_a?(Hash)
 
       @@last = self
       @default_opts = @default_db ? {:db => RQL.new.db(@default_db)} : {}
@@ -636,8 +639,7 @@ module RethinkDB
 
     def connect()
       raise RuntimeError, "Connection must be closed before calling connect." if @socket
-      @socket = TCPSocket.open(@host, @port)
-      @socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
+      init_socket
       @mon = Monitor.new
       @waiters = {}
       @opts = {}
@@ -645,6 +647,44 @@ module RethinkDB
       @conn_id += 1
       start_listener
       self
+    end
+
+    def init_socket
+      unless @ssl_opts.empty?
+        @tcp_socket = base_socket
+        @tcp_socket.connect(::Socket.pack_sockaddr_in(@port, @host))
+        context = create_context(@ssl_opts)
+        @socket = OpenSSL::SSL::SSLSocket.new(@tcp_socket, context)
+        @socket.sync_close = true
+        @socket.connect
+        verify_cert!(@socket, context)
+      else
+        @socket = base_socket
+        @socket.connect(Socket.pack_sockaddr_in(@port, @host))
+      end
+    end
+
+    def base_socket
+      socket = Socket.new(Socket::PF_INET, Socket::SOCK_STREAM, 0)
+      socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
+      socket
+    end
+
+    def create_context(options)
+      context = OpenSSL::SSL::SSLContext.new
+      if options[:ssl_verify] || options[:ssl_ca_cert]
+        context.ca_file = options[:ssl_ca_cert]
+        context.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      end
+      context
+    end
+
+    def verify_cert!(socket, context)
+      if context.verify_mode == OpenSSL::SSL::VERIFY_PEER
+        unless OpenSSL::SSL.verify_certificate_identity(socket.peer_cert, host)
+          raise 'SSL handshake failed due to a hostname mismatch.'
+        end
+      end
     end
 
     def is_open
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks a bunch for the patch @jipperinbham.

@mlucy Can you take a look at the code please? (@jipperinbham has already signed the CLA)

I think we might want to polish the options that `r.connect` takes a bit, but I haven't thought through it in detail yet.

Another question I had: Is the `openssl` module something that's usually available in a Ruby setup? If not, we should try to find a way to require it only if ssl is actually being used.

jipperinbham:
`openssl` is part of the stdlib and the include has to happen at the class level. I could spend some time breaking out the sockets into individual classes which would make it possible to only include the `openssl` library when needed.

danielmewes:
I think if it's in the standard library we're good.

jipperinbham:
re: SSL options in the connection, I tried to make it such that adding client cert authentication would be pretty straightforward in the future and the number of SSL arguments could get large so just sticking them all in a hash and documenting properly could do the trick. However, I'm also good with being more explicit with the arguments if that's preferred.

danielmewes:
I personally think passing through the ssl options to OpenSSL is fine.

I think we probably want to have a separate option though to enable openSSL in the first place.
Maybe something like `ssl: true`?

I'd like to wait for @mlucy's thoughts on this as well.

drivers/ruby/lib/net.rb
```diff
@@ -455,6 +457,7 @@ def initialize(opts={})
       @auth_key = opts[:auth_key] || ""
       @timeout = opts[:timeout].to_i
       @timeout = 20 if @timeout <= 0
+      @ssl_opts = opts.reject { |k, v| !k.to_s.start_with?('ssl') } if opts.is_a?(Hash)
```
mlucy: It looks like this will set `@ssl_opts` to `nil` if `!opts.is_a?(Hash)`, but in `init_socket` we call `@ssl_opts.empty?` without checking whether or not it's `nil`.

drivers/ruby/lib/net.rb
```diff
@@ -647,6 +649,44 @@ def connect()
...
+        @socket.connect
+        verify_cert!(@socket, context)
+      else
+        @socket = base_socket
+        @socket.connect(Socket.pack_sockaddr_in(@port, @host))
+      end
+    end
+
+    def base_socket
+      socket = Socket.new(Socket::PF_INET, Socket::SOCK_STREAM, 0)
```
mlucy: This used to be `TCPSocket.open`, right?  What was the motivation for the change?  (Also, are there any cases where this behaves differently?)

mlucy: Are there any cases where this followed by `.connect` behaves differently than `TCPSocket.open`?

mlucy:
This looks good to me aside from the two small comments.  Thanks for the contribution, @jipperinbham !

I think I agree with @danielmewes that turning on SSL automatically when you pass an argument starting with `ssl` is a little bit too much magic.  I think we should instead make `connect` take a single optional argument `ssl` which can be set to whatever options the user wants to pass through to `create_context`, and if that argument isn't set we don't turn on SSL.  (So it would look like `.connect(port: 100, ssl: {opt1: true, opt2: true))` instead of `.connect(port: 100, sslopt1: true, sslopt2: true)`.)  @jipperinbham, @danielmewes, how does that sound to you?

@danielmewes -- We'll also need to write some documentation for this, and a test so the code doesn't bitrot.  If we do what I suggested above, I think we can get away with just adding documentation for the `ssl` optarg to the `connect` command.

drivers/ruby/lib/net.rb
jipperinbham: I switched back to using `TCPSocket.open` as it is just a convenience method for `.new` but also handles calling `connect`.

jipperinbham:
Ok, so I made the options change to nest the SSL specific options in the `:ssl` Hash. Currently, the only supported and required option is `ca_certs`.

re: tests, this seems a bit tricky as you'll need a proxy (Haproxy, nginx or other) running to do the SSL certificate validation. Open to suggestions on how best to do that.

drivers/ruby/lib/net.rb
jipperinbham: This was refactored as part of nesting the SSL options in a Hash

danielmewes:
The single `ssl` option sounds good.

@jipperinbham don't worry about the testing, we will take care of that.
@larkost any ideas for how we could test SSL connectivity without having native SSL support in the server and ideally without adding too many dependencies? Another tool that we could use in the tests besides haproxy and nginx is https://www.stunnel.org/index.html .

larkost:
We already have SSL support in `test/rql_test/connections/http_support/server.py` and it should not be too difficult to add another end-point to that as a simple dumb proxy to a server instance that it could spin-up on-demand. There should not be any additional dependencies.

danielmewes:
@larkost perfect, that sounds good.

larkost:
I have separated the SSL regression testing into a separate issue: #4084

mlucy:
Cool, the code looks good to me.  I opened a new issue in `docs` for the documentation.

@danielmewes -- I assume this should go in both `next` and `2.0.x`?

danielmewes:
Only next I think.

mlucy:
Alright, merged it in.  Thanks again @jipperinbham !

joaojeronimo:
Why is the `ca_certs` option required ? If the server has a valid verified certificate chain (i.e. a certificate bought from a trusted CA) then most SSL implementations verify it on their own (honestly I don't know how but I imagine most of them use openssl underneath ?). Isn't it the case with ruby also ?

jipperinbham:
@joaojeronimo the ca_certs is required mostly because the typical use case for certificates used to encrypt database communication use self-signed certs.

joaojeronimo:
@jipperinbham exactly, **most**. It is still however possible to use a certificate with a valid chain, and in that case I don't want to use the ca_certs argument when connecting with the driver, therefore it must be an optional argument.

By the way I do use a certificate with a valid chain on my servers because they are cheap and I don't want to keep distributing certificates to the clients :) And they will be [free](https://letsencrypt.org/) soon-ish :)

danielmewes:
@joaojeronimo that makes sense. I opened https://github.com/rethinkdb/rethinkdb/issues/4089 so we don't forget checking and/or changing that.

PR #4076, opened 2015-04-17
[python] support for ssl encryption and CA verification

jipperinbham:
[Addresses Python Driver](https://github.com/rethinkdb/rethinkdb/issues/3158)

The following abstracts the socket connection out a bit to allow for creating an SSL connection to a cluster. In order for this to be used, you'd need a proxy (Haproxy, nginx, etc) sitting in front of the RethinkDB cluster performing the SSL verification.

For the implementation, I'm using the additional kwargs to add `ssl` and `ca_certs`.

### DIFF
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -6,6 +6,7 @@ import numbers
 import socket
 import struct
 import time
+import ssl
 try:
     from importlib import import_module
 except ImportError:
@@ -24,6 +25,11 @@ pQuery = p.Query.QueryType
 from .errors import *
 from .ast import RqlQuery, RqlTopLevelQuery, DB, Repl
 from .ast import recursively_convert_pseudotypes
+
+try:
+    from ssl import match_hostname, CertificateError
+except ImportError:
+    from backports.ssl_match_hostname import match_hostname, CertificateError
 
 try:
     xrange
@@ -235,6 +241,8 @@ class SocketWrapper(object):
         self.port = parent._parent.port
         self._read_buffer = None
         self._socket = None
+        self._use_ssl = parent._use_ssl
+        self._ca_certs = parent._ca_certs
 
         deadline = time.time() + timeout
 
@@ -242,6 +250,20 @@ class SocketWrapper(object):
             self._socket = \
                 socket.create_connection((self.host, self.port), timeout)
             self._socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+
+            if self._use_ssl:
+                ssl_context = self._get_ssl_context(self._ca_certs)
+                try:
+                    self._socket = ssl_context.wrap_socket(self._socket,
+                                                           server_hostname=self.host)
+                except IOError as exc:
+                    self._socket.close()
+                    raise RqlDriverError("SSL handshake failed: %s" % (str(exc),))
+                try:
+                    match_hostname(self._socket.getpeercert(), hostname=self.host)
+                except CertificateError:
+                    self._socket.close()
+                    raise
 
             self.sendall(parent._parent.handshake)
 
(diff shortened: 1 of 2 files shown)


drivers/python/rethinkdb/net.py
```diff
@@ -26,6 +27,11 @@ def import_module(name, package=None):
 from .ast import recursively_convert_pseudotypes
 
 try:
+    from ssl import match_hostname, CertificateError
+except ImportError:
+    from backports.ssl_match_hostname import match_hostname, CertificateError
```
danielmewes: As mentioned, I think we should make this import optional in case the dependency isn't installed.

drivers/python/rethinkdb/net.py
```diff
@@ -243,6 +251,20 @@ def __init__(self, parent, timeout):
...
+            if self._use_ssl:
+                ssl_context = self._get_ssl_context(self._ca_certs)
+                try:
+                    self._socket = ssl_context.wrap_socket(self._socket,
+                                                           server_hostname=self.host)
+                except IOError as exc:
+                    self._socket.close()
+                    raise RqlDriverError("SSL handshake failed: %s" % (str(exc),))
+                try:
+                    match_hostname(self._socket.getpeercert(), hostname=self.host)
```
danielmewes: @Tryneus When using Tornado, is connection setup currently asynchronous? If it is, I wonder how that will interact with the SSL handshake.

danielmewes:
Thank you for this one too @jipperinbham !

@Tryneus: Can you take a look at this please?

(related PR for the Ruby driver: https://github.com/rethinkdb/rethinkdb/pull/4075 )

drivers/python/rethinkdb/net.py
Tryneus: Tornado doesn't use this code path because it uses `tornado.iostream` instead.  This will need to be added to `net_tornado.py`:`ConnectionInstance` for SSL to work there.  Should be fairly easy using [`tornado.iostream.SSLIOStream`](http://tornado.readthedocs.org/en/latest/iostream.html#tornado.iostream.SSLIOStream).

For that class, it says that when connecting (which we do asynchronously), the future will be fulfilled when both the connection and handshake have completed.

drivers/python/rethinkdb/net.py
```diff
@@ -346,14 +368,27 @@ def sendall(self, data):
                 self.close()
                 raise
 
+    def _get_ssl_context(self, ca_certs):
```
Tryneus: For the purpose of sharing code with `net_tornado.py`, it may be useful to pull this out of `SocketWrapper`.  It doesn't appear to use any member data.

Tryneus:
@jipperinbham, this looks pretty good, but I had some comments, mostly regarding Tornado support.  If you aren't interested in doing the Tornado implementation at the moment, it's not necessary.

jipperinbham:
@Tryneus, I'll get tornado support in this PR and refactor out the SocketWrapper from net.py so it can be shared.

jipperinbham:
@Tryneus, after looking into the `iostream` module, it doesn't require a context but when making the `SSLIOStream` instance. I don't know how to properly test the tornado implementation so if you could point me in a direction, I'll make sure it works properly.

danielmewes:
I think we should also try to make the optional arguments consistent with those of the Ruby implementation (see https://github.com/rethinkdb/rethinkdb/pull/4075#issuecomment-94173834).

jipperinbham:
@danielmewes I've updated the connect method to have its own ssl argument and be of type `dict` so it's identical to the ruby implementation now.

jipperinbham:
@danielmewes is there anything outstanding here?

Tryneus:
@jipperinbham, this looks good, although the merge isn't trivial now, I'll perform the merge manually.

Tryneus:
Merged to `next`, thanks for the help, @jipperinbham!  We'll see about adding tests for this and extending it to any new `net.py` implementations (e.g. an asyncio PR was accepted yesterday, and I have a gevent implementation on the backburner).

Issue #4074, opened 2015-04-17
Integrate SSH into client driver


stuartpb:
As originally proposed (obliquely) in https://github.com/rethinkdb/rethinkdb/issues/266#issuecomment-18477141

As long as SSH tunneling is the official way to secure RethinkDB traffic ([and the only supported interface to RethinkDB by Compose](https://blog.compose.io/tunneling-from-heroku-to-compose-rethinkdb/)), it should be a feature of the client driver to handle connecting to the remote through an SSH tunnel (if the options to r.connect include a `privateKey` and `username`).

In the JS driver, this could be handled with the [ssh2](https://github.com/mscdex/ssh2) module - I don't know about Python or Ruby, but it sounds like they have analogous native-SSH modules such functionality could be built upon.

Also, this is unrelated, but while we're on the subject of SSH-as-mechanism in RethinkDB, I should probably add `sshd` (and maybe `iptables` although the Docker bridge already handles 50% of what you'd use `iptables` for) to the official RethinkDB Docker images, y/n?

danielmewes:
Hi @stuartpb, thanks for the suggestion.

I think we will use SSL for this instead of SSH. We still have to figure out how to do proper client-authentication with SSL, but I think it's the better choice for this. At least in my experience SSH has a pretty slow connection setup, which would be a problem for client connections.

Actually @jipperinbham from Compose has just submitted a pull request to allow the Ruby driver to connect over SSL (not SSH) https://github.com/rethinkdb/rethinkdb/pull/4075 . I hear he's working on support for other drivers as well.

With SSL coming to our drivers, another option you could consider for the Docker image is https://www.stunnel.org/index.html .

joaojeronimo:
@danielmewes Why deal with authentication over SSL ? Why not just do it with the `authKey` ? IMHO SSL should be there just for encryption. Do you meant certificate chain validation on the client ?

Sidenote: If in the future you guys come up with user accounts and an ACL, since the protocol accepts an arbitrary string for the authKey, you could send something like the Authorization http header (Authorization method + ' ' + base64encode(username + ':' + password))

danielmewes:
@joaojeronimo Using the auth key will work in many cases, but we haven't written the connection logic with public access in mind. Hence there might be some vulnerabilities for DoS attacks or other issues there. With SSL client authentication, we could rely on the SSL proxy to take care of these issues for us until we have gotten around to thoroughly test and audit the relevant code paths.
[...]

PR #4080, opened 2015-04-19
[JS] support for ssl encryption and CA verification

jipperinbham:
[Addresses JavaScript Driver](https://github.com/rethinkdb/rethinkdb/issues/3158)

The following will use `tls.connect` when doing an SSL connection to a cluster. In order for this to be used, you'd need a proxy (Haproxy, nginx, etc) sitting in front of the RethinkDB cluster performing the SSL verification.

For the implementation, I'm using additional `r.connect` options: `ssl` and `sslCaCerts` so a connection would look like the following:

```javascript
r.connect({ host: 'localhost',
            port: 28015,
            db: 'marvel',
            authKey: 'hunter2' },
            ssl: true,
            sslCaCerts: '/path/to/ca.crt'
          function(err, conn) { ... })
```

### DIFF
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -15,6 +15,16 @@
 # objects
 net = require('net')
 
+# The [tls module](http://nodejs.org/api/tls.html) is the TLS/SSL
+# networking library Node.js provides. We need it to establish an
+# encrypted connection for `TcpConnection` objects
+tls = require('tls')
+
+# The [fs module](http://nodejs.org/api/fs.html) is the high-level
+# File I/O library Node.js provides. We need it to load any
+# certificate objects for SSL connections with `TcpConnection` objects
+fs = require('fs')
+
 # The [events module](http://nodejs.org/api/events.html) is a core
 # Node.js module that provides the ability for objects to emit events,
 # and for callbacks to be attached to those events.
@@ -142,6 +152,14 @@ class Connection extends events.EventEmitter
 
         @authKey = host.authKey || @DEFAULT_AUTH_KEY
         @timeout = host.timeout || @DEFAULT_TIMEOUT
+
+        # Configuration options for enabling an SSL connection
+        # Currently, the only support option is '"sslCaCerts"' which
+        # is needed to perform host verification and prevent Man in
+        # the Middle attacks.
+        @ssl = host.ssl || false
+        if @ssl
+            @caCerts = host.sslCaCerts
 
         # The protocol allows for responses to queries on the same
         # connection to be returned interleaved. When a query is run
@@ -900,7 +918,16 @@ class TcpConnection extends Connection
         # Next we create the underlying tcp connection to the server
         # using the net module and store it in the `@rawSocket`
         # attribute.
-        @rawSocket = net.connect @port, @host
+        options = {
+            host: @host,
+            port: @port,
+        }
+        if @ssl
+            options.ca = [ fs.readFileSync(@caCerts) ]
+            options.servername = @host
+            @rawSocket = tls.connect options
+        else
+            @rawSocket = net.connect @port, @host
 
         # We disable [Nagle's
         # algorithm](http://en.wikipedia.org/wiki/Nagle%27s_algorithm)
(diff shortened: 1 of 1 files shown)


danielmewes:
Nice, number three :)
@deontologician could you take a look at this patch?

(Also for consistency I think we should do the same with the `ssl` optional argument as for the Ruby driver, https://github.com/rethinkdb/rethinkdb/pull/4075#issuecomment-94173834 .)

jipperinbham:
@danielmewes updated `ssl` arguments here as well.

deontologician:
This looks good code-wise, it's not a lot of code and it sends the right stuff to the tls api. How do we check if it can make the connection? Do you need an SSL proxy?

danielmewes:
@neumino mentioned on another thread (https://github.com/rethinkdb/rethinkdb/issues/4074#issuecomment-94611591):
> the proposed implementation uses readFileSync, which is pretty bad since it will block your whole node.js server the time to load the certificate.

I think we should replace it by an asynchronous method if possible.

danielmewes:
He also suggested that we might want to cache the certificate file somewhere, so we don't have to go to the operating system for re-reading it for every new connection.

joaojeronimo:
IMHO that would be over-optimization, how many times are we opening connections to rethinkdb for that to become a problem because of readFileSync ?

Also, why not follow the same signature of [tls.connect](https://nodejs.org/api/tls.html#tls_tls_connect_port_host_options_callback) and accept an options object that will be passed all the way to tls.connect (because it's what will be used underneath right ?) ? The options object is an object that may have `ca` and `cert` parameters that may be strings or buffers, so we pass on to the user the option of how to obtain those (they decide to use fs.readFile or fs.readFileSync). It would be a lot cleaner IMHO, [it's about 10 SLOC](https://github.com/joaojeronimo/rethinkdbdash/blob/master/lib/connection.js#L41).

jipperinbham:
I agree that trying to optimize `readFileSync` is not necessary and connections should be reused not constantly opened/closed.

@joaojeronimo the connection options have been updated to look like:
```javascript
r.connect({ host: 'localhost',
            port: 28015,
            db: 'marvel',
            authKey: 'hunter2' },
            ssl:{
              caCerts: '/path/to/ca.crt'
            }
          function(err, conn) { ... })
```

So, if I'm understanding right, you'd want to just have the `ssl` hash accept the arguments `tls.connect`? And the example above would become:

```javascript
r.connect({ host: 'localhost',
            port: 28015,
            db: 'marvel',
            authKey: 'hunter2' },
            ssl:{
              ca: [ fs.readFileSync('/path/to/ca.crt') ]
            }
          function(err, conn) { ... })
```

joaojeronimo:
Yes I think having the `ssl` hash be the same as the `options` hash used in `tls.connect` would be cleaner on the implementation (nothing else there to do, plus you support every option that tls.connect supports, which should be every possible option anyone would need), more idiomatic and node people would be a lot more familiar with how to use it.

danielmewes:
I think accepting a file path for the certificate file like it does now is slightly nicer (and it makes it more consistent with the other drivers).

Would it be a big problem to change the file loading to `readFile`? From looking at the API it looks pretty straight forward (you could wrap it into a promise I think to avoid the separate callback, @deontologician could help you with that).

There are definitely a few users who open new connections very frequently, so I would prefer if the driver behaved well under that scenario.
@jipperinbham if you don't want to do this change don't worry. We can do it ourselves after the merge.

jipperinbham:
@danielmewes I've made a [change](https://github.com/compose/rethinkdb/commit/1a80811a2911e492be03f8ca342fbdca3b4d2ac3) that I think is best given how different node is and the differences in how users may want to connect. It's what @joaojeronimo and I discussed above and it's also how my [PR](https://github.com/dancannon/gorethink/pull/179) to the gorethink driver is setup. 

Since the python and ruby drivers setup the SSL/TLS connection differently, I think it's ok to have the connect options differ slightly.

danielmewes:
> Since the python and ruby drivers setup the SSL/TLS connection differently, I think it's ok to have the connect options differ slightly. 

I can live with that.

I'll leave the final review to @deontologician .

deontologician:
Looks good, merged.

joaojeronimo:
:+1: 

Is it going to be on npm right away or just released along with 2.1 ?

danielmewes:
@joaojeronimo This will be released with the 2.1 release.

joaojeronimo:
can't wait :smiley:

pauliusuza:
Fix in PR #4786

PR #4170, opened 2015-05-01
Update proto doc for contains

hamiltop:
While implementing contains for https://github.com/hamiltop/exrethinkdb I found that the behavior of the function to differ from the spec.

1. Multiple predicates are allowed and behave as a logical `AND`
2. A mix of predicates and datum are allowed and behave as a logical `AND`

Number 1 is reflected in the actual docs properly but not reflected in the spec.
Number 2 is in neither the docs nor the comments in the spec.


danielmewes:
Thanks @hamiltop . Looks good. I also opened https://github.com/rethinkdb/docs/issues/747 for updating the API docs about 2.

Issue #3510, opened 2015-01-07
Add websockets interface to the webui/server


deontologician:
This will make it much easier/cleaner for the webui to listen to changefeeds on the admin tables. It has the downside of needing a websockets implementation on the server. I think a custom implementation of a standard protocol is better than the custom implementation of a custom protocol, but it would be potentially a lot of work unless there's a library we could bend into shape. In addition, the abstraction of socket -> websocket is probably closer than what we have now to adapt the js driver to talk to the http connection (i.e. we could dump a lot of code in the js driver) There was an issue about removing the http connection stuff from the js driver that I can't find at the moment.

[Implementing the websockets protocol from MDN](https://developer.mozilla.org/en-US/docs/WebSockets/Writing_WebSocket_servers)

joaojeronimo:
Listening to changefeeds seems unidirectional communication and for that you could and IMHO maybe should use [Server Sent Events](http://www.html5rocks.com/en/tutorials/eventsource/basics/) which has a much simpler implementation and wouldn't make such a mess on the server.

kofalt:
See also #3460

danielmewes:
@joaojeronimo Thanks for the suggestion and link.
I think part of @deontologician 's goal is to make the protocol used by the web UI to connect to the server more similar to the protocol we use for regular clients. With websockets we might need less specialized code in the JavaScript driver for dealing with tunnelling our protocol over HTTP connections. My feeling is that server sent events would actually be further away from our regular client protocol, which is generally based on request/response pairs.
Nonetheless it might still make sense for the special case of changefeeds if there are other benefits.

joaojeronimo:
Sure if you want to replicate your TCP protocol then websockets are clearly the way to go. There are a bunch of node.js modules you can look at to see how they handle streams on websockets on backend and frontend, like https://github.com/maxogden/websocket-stream

deontologician:
Another part of the goal here is to push the pollojg down to the OS/ network layer where it belongs, instead of JavaScript's settimeout

danielmewes:
A big advantage of this is that we wouldn't rely on multiple subsequent HTTP connections to get routed to the same RethinkDB server.

If you have a proxy in front of a RethinkDB cluster and want to do round-robin load balancing, our current protocol is unable to deal with that since it relies on multiple connections to reach the same server.

Issue #218, opened 2013-01-12
User Accounts


emilbayes:
I'm extemely excited about rethinkdb and have been playing around with it for a while. Are user accounts on the road map any time soon?

I'm thinking accounts restricting access to specific databases and read/write restrictions.

dmpayton:
I wanted to add my own +1 to this; user accounts are important for our use case. My wish list is pretty much exactly what @tobymarsden posted.

blazedd:
+1

jampekka:
+1 Although I'd like to see a rethought approach to this too. A good approach would be to think how PostgreSQL does it as a example not to do user handling.

I've been fantasizing about something like having just secret keys (what's the point of having both user name and password anyway?) and definitely no messing about with ident authentication or filtering hosts etc (these are all easier and better done on OS level if needed).

Also no ACL's or per table permissions etc everybody has hated for decades. I think a lot more elegant approach would be that for each key, a separate "view" of the database could be created (using ReQL of course), where the views would have to be explicitly declared. I think this could be surprisingly simple if the current architecture is flexible enough.

Write access with such thing could get a bit tricky if the views are very flexible, but an easy approach would be to allow writing only to very simply filtered views.

On top of my head this could look something like:

```python
r.table("restricted_namespaces").insert({
    "access_keys": ["somestringthatmayormaynotbesecret", "anotherkeyusingthesameview"],
    "view_name": "public_pages",
    "view": r.table("pages").filter({"public": True}).without("secret", "fields"),
    "writable": False
})
```

The views in themselves would be a very powerful thing, especially if they could be stored and used in queries.

jampekka:
Oh, and at least memory and CPU quotas could be better done just running different processes for the different keys, so OS level controls (eg RLIMIT_* stuff on linux) could be used quite elegantly to prevent DOSs.

danielmewes:
Thank you for your input @jampekka . Using general views for access control is an interesting idea.

I don't think we can easily spawn separate processes for each view with our current architecture, so we'll have to do CPU and memory quotas ourselves unfortunately.

jampekka:
In the current architecture this could probably be worked around by setting other replica for access, which could use OS level limits, if there would be option for per-replica key views.

Although in the longer term I'd wager that refactoring the architecture to allow multiple processes would pay off in the long run. It could allow also using RethinkDB as an process-embedded "serverless" database. But this probably should have an another ticket for better discussion.

[...]

PR #4211, opened 2015-05-10
Websocket transport for the Javascript driver

dparnell:
Hi, I've been playing with alternate ways of connecting to a rethinkdb database and have built something that allows a web page to connect to a rethinkdb database via a websocket connection.
I've been using it in a simple internal app for some time now and it seems to be working very well for my needs :smile: 
Thanks for rethinkdb, it's awesome

Daniel


danielmewes:
Hey @dparnell, nice patch. Thank you for the pull request.

I think we are not going to merge this into the main driver for now, because we won't be using it in our web UI and wouldn't have a great way for testing it and making sure that the code keeps working. I hope you understand.

That being said we have thought about switching the web UI to websockets (https://github.com/rethinkdb/rethinkdb/issues/3510), and if we end up doing that, your patches might become useful for the main driver.
Another change in our main server that would make this really useful would be adding user authentication and additional security https://github.com/rethinkdb/rethinkdb/issues/218 . At that point it will become feasible to connect directly from the browser to a database for a larger group of users, and websockets would likely be the way to go about that.

I think for the time being we won't make this part of the main driver, but I'll keep the pull request open nonetheless. Hope that's ok with you.
Do you have any plans of publishing your changes separately? I'm sure there will be a few people who are interested in this.

dparnell:
Thanks, I wondered if that might be the case.  I'm also working on adding the ability to authenticate via a web page.  The idea I'm working on at the moment is to have a proxy that sits in front of the rethinkdb server that uses a provided auth token to determine the user logging in.  This is then used to get the privs that should be granted to the user and so any query that comes in is augmented/checked to ensure the user has the necessary privs on the desired tables.

danielmewes:
@dparnell That sounds like a really interesting project. Are you planning to open source it?

dparnell:
Possibly, it's written in Erlang at the moment which does limit the number of potential developers, but once there is a reference implementation porting it to another language wouldn't be too tricky.  The other thing I'm working on at the moment is some new functions in ReQL to materialize a subselect to stop rethinkdb from re-evaluating it each time through a filter function.  I've got that part working and while it helped the query I'm trying to do it's still too slow so I'm looking at trying to make a new function that sorts a given array so that the `contains` function can use a binary chop search instead of a linear search.  It's been fun digging through the guts of the rethinkdb codebase :smile:

srh:
Closing this.  Why now?  This is in preparation for `v2.4.x` becoming the main RethinkDB branch instead of `next`.

Also note:
- This PR was made before the Apache 2.0 relicensing, so right now its contents could be presumed AGPLv3-licensed, unless we know its author signed the RethinkDB CLA before the company folded.

- The JavaScript driver is now located in https://github.com/rethinkdb/rethinkdb-javascript

- It seems like it could be useful to somebody, which is why the PR had been left open.

dparnell:
I don't recall signing the RethingDB CLA, however I'm happy for this code to be used in any way.

PR #4269, opened 2015-05-22
JS Driver: allow `r.js` to accept a function

### LINKED ISSUE #4260: JS Driver: allow `r.js` to accept a function
Currently `r.js` only accepts a JS expression in the form of a string. This makes even simple expressions difficult to read:

```js
r.js('(function (row) { return row.magazines.length > 5; })')
```

Fortunately, in JS you can call [`Function.prototype.toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString) to get the entire function declaration back as a string. In the JS driver when the user passes a function to `r.js` we could simply convert it to a string before passing it through.

```js
r.js(function (row) {
  return row.magazines.length > 5;
})
```

marshall007:
Fixes #4260.

### DIFF
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -1197,7 +1197,10 @@ rethinkdb.expr = varar 1, 2, (val, nestingDepth=20) ->
     else
         new DatumTerm val
 
-rethinkdb.js = aropt (jssrc, opts) -> new JavaScript opts, jssrc
+rethinkdb.js = aropt (jssrc, opts) ->
+    if typeof jssrc is 'function'
+        jssrc = "(#{jssrc.toString()})"
+    new JavaScript opts, jssrc
 
 rethinkdb.http = aropt (url, opts) -> new Http opts, url
 
--- a/test/rql_test/src/control.yaml
+++ b/test/rql_test/src/control.yaml
@@ -140,6 +140,10 @@ tests:
     - cd: r.expr('foo').do(r.js('(function(x) { return x + "bar"; })'))
       ot: "'foobar'"
 
+    # should accept function argument in JS driver
+    - js: r.expr('foo').do(r.js(function(x) { return x + "bar"; }))
+      ot: "'foobar'"
+
     # js timeout optarg shouldn't be triggered
     - py: r.js('1 + 2', timeout=1.2)
       js: r.js('1 + 2', {timeout:1.2})


deontologician:
This is a neat trick. It won't translate to the other drivers unfortunately, but I think that asymmetry should be ok given we don't have an `r.python` or `r.ruby`. Approved

AtnNn:
I don't think we should include this functionality by default. In most cases it will not work, for example when using built-in functions or when using free variables. It would be more confusing than helpful.

```
> r.js(Object.toString)
{ args: [ { [Function] data: '(function toString() { [native code] })' } ], optargs: {} }
```

```
> var f = function(){ ... }
> r.js(function(){return f()})
{ args: [ { [Function] data: '(function (){return f()})' } ], optargs: {} }
```

deontologician:
I wouldn't say "most cases" it wouldn't work. It definitely has edge cases, but so does our native functions -> reql functions trick. It's a leaky abstraction certainly

marshall007:
@AtnNn this was purely for convenience. It should work in all cases that `r.js` already supported.

If we added an example using this syntax to the docs we could clarify that you can't reference anything outside the function's scope, but I think most people would assume that.

AtnNn:
It definitely is "most cases". The vast majority of functions have free variables. And most functions in a query use reql. But functions passed to r.js can neither have free variables nor can they use ReQL.

I believe the small convenience this adds does not outweigh the complexity that it adds to the API and its documentation. It can instead easily be implemented elsewhere as a one-liner.

Functions passed to ReQL already have semantics and restrictions that are hard to understand. This change introduces another type of function that can be used in ReQL, but with very different restrictions.

RethinkDB error messages are not good enough to help debug errors that could be caused by the misuse of this feature.

deontologician:
Ok, it looks like this was a little more controversial than I'd realized. It probably should have had more discussion in terms of the interface. I'll revert it for now and re-open #4260 so it can be properly debated (I'm leaning slightly towards @AtnNn 's point of view, but I'd like to give other people a chance to weigh in, in case anyone else thinks differently)

timmaxw:
I agree with @AtnNn. For example, with our native functions -> reql trick one can write:
```js
var x = 1;
r.table(...)("score").map(function(y) { return x + y; })...
```
But this won't work for a function passed to `r.js()`. I think this is likely to be very confusing.

Also, there are already enough people confused about how our query evaluation works, because it looks like it's being executed client-side but it's actually being executed server-side. Let's not make it worse by adding something else that looks like it's being executed client side but is actually being executed server-side, but with different semantics and corner cases.

PR #4305, opened 2015-05-29
JS Driver: Allow `.match()` to accept a native RegExp object

marshall007:
This PR allows you to pass `RegExp` objects to `.match()` instead of strings. Example:

```js
r.expr("abcdefg").match("(?i)a(b.e)|B(c.e)")
r.expr("abcdefg").match(/a(b.e)|B(c.e)/i)
```

I realize the `RE2` and JS implementations differ slightly so while this may not work in all cases I think it's a nice convenience for the majority of expressions users are likely to write.


mlucy:
Hi @marshall007,

I opened https://github.com/rethinkdb/rethinkdb/issues/4308 to discuss this, since it isn't 100% obvious to me that it's a good idea.

(As an aside, thanks for contributing so much to the project!  I feel bad bogging down pull requests in post facto discussions, but we generally try to discuss changes to the query language before merging them in to make sure nobody objects.)

marshall007:
@mlucy of course, thanks for opening the issue. The change was trivial so I went ahead and implemented it before discussion, but this was definitely intended to be a proposal.

danielmewes:
As discussed in https://github.com/rethinkdb/rethinkdb/issues/4308 , we're not going to merge this for now.

PR #4349, opened 2015-06-05
JS Driver: db optarg to `.run()` is ignored if default db is already set on the connection

### LINKED ISSUE #4347: JavaScript driver overwrites 'db' optarg passed to 'run' if connection default is set
Both the Ruby and Python drivers use a 'db' optarg passed into 'run' over the default DB set on the connection using 'use'. However the JavaScript driver overwrites the 'db' optarg if there is a connection default without checking if it had already been set.

I think we should change the JS driver to make this consistent and behave more like expected.

marshall007:
Fixes #4347

### DIFF
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -781,10 +781,10 @@ class Connection extends events.EventEmitter
             query.global_optargs[util.fromCamelCase(key)] = r.expr(value).build()
 
         # If the user has specified the `db` on the connection (either
-        # in the constructor, or with the `.use` method, we add that
-        # db to the optargs for the query.
-        if @db?
-            query.global_optargs['db'] = r.db(@db).build()
+        # in the constructor, as an optarg to .run(), or with the `.use` method)
+        # we add that db to the optargs for the query.
+        if opts.db? or @db?
+            query.global_optargs.db = r.db(opts.db or @db).build()
 
         # Next, ensure that the `useOutdated`, `noreply`, and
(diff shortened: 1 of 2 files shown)


drivers/javascript/net.coffee
```diff
@@ -781,10 +781,10 @@ class Connection extends events.EventEmitter
...
 
         # If the user has specified the `db` on the connection (either
-        # in the constructor, or with the `.use` method, we add that
-        # db to the optargs for the query.
-        if @db?
-            query.global_optargs['db'] = r.db(@db).build()
+        # in the constructor, as an optarg to .run(), or with the `.use` method)
+        # we add that db to the optargs for the query.
+        if opts.db? or @db?
+            query.global_optargs.db = r.db(opts.db or @db).build()
```
danielmewes: You can slightly simplify this, since the `db` optarg is already added to query.global_optargs above in 781:
```js
if @db? and not opts.db?
  query.global_optargs['db'] = r.db(@db).build()
```

I believe `db` also works if it's passed to the server as a string instead of an r.db object and this is what we are eventually going to move to entirely (compare https://github.com/rethinkdb/docs/issues/332, though it's unclear when that will happen).

danielmewes: Oh never mind. I was wrong, looks like the server actually requires the argument to be wrapped into a DB term.

danielmewes:
Thanks for the pull request @marshall007 . Looks good.

PR #4396, opened 2015-06-14
Ignore whitespace in database delete confirmation

ianberinger:
Hi, I've recently found myself deleting some databases manually via the web interface and was annoyed that whitespace around the database name wasn't ignored. This PR fixes that.
Feel free to close if you think this is unnecessary. Btw, thanks for the amazing work so far!

### DIFF
--- a/admin/static/coffee/modals.coffee
+++ b/admin/static/coffee/modals.coffee
@@ -92,7 +92,7 @@ class DeleteDatabaseModal extends ui_modals.AbstractModal
 
 
     on_submit: =>
-        if @$('.verification_name').val() isnt @database_to_delete.get('name')
+        if $.trim(@$('.verification_name').val()) isnt @database_to_delete.get('name')
             if @$('.mismatch_container').css('display') is 'none'
                 @$('.mismatch_container').slideDown('fast')
             else


danielmewes:
Thank you for the pull request @ianberinger .
If you haven't done so yet, could you please sign our CLA? http://rethinkdb.com/community/cla/ We require this for legal reasons before we can merge pull requests.

I think it's fine to ignore white spaces at the front and end of the string here.

@deontologician could you check the code please?

admin/static/coffee/modals.coffee
```diff
@@ -92,7 +92,7 @@ class DeleteDatabaseModal extends ui_modals.AbstractModal
 
 
     on_submit: =>
-        if @$('.verification_name').val() isnt @database_to_delete.get('name')
+        if @$('.verification_name').val().replace(/^\s+|\s+$/g,'') isnt @database_to_delete.get('name')
```
marshall007: I think this could be simplified to `@$('.verification_name').val().trim()`.

ianberinger: I would generally agree, but I wasn't sure which browser versions should be supported. IE 8 for instance, doesn't support String.prototype.trim. See: https://kangax.github.io/compat-table/es5/#String.prototype.trim
If IE8 support isn't important, I could change it to `.trim()`.

ianberinger:
I've signed the CLA.

admin/static/coffee/modals.coffee
deontologician: Currently we use a lot that isn't IE8 compatible, but we try to polyfill where appropriate. For now I'd say leave it since it's just a code-clarity issue and it's not adding tons of boilerplate. I'll open an issue for making an official policy on which browsers we want the webui to support (including adding polyfills if necessary)

deontologician: Actually, I just realized, we do have `$.trim`, so that should be ok to use

ianberinger:
I've updated my PR to use jquery's trim function.

deontologician:
Looks good! Thanks @ianberinger and thanks @marshall007 for looking at it as well

PR #4442, opened 2015-06-23
Make `<bool>.branch(...)` work

### LINKED ISSUE #4438: Make `<boolean>.branch(<b1>, <b2>)` work
In the JavaScript driver at least, we currently support
`r.branch(<boolean>, <b1>, <b2>)` but not `<boolean>.branch(<b1>, <b2>)`

I think we should support both syntaxes.

(checking Python and Ruby as well, even though I suspect that it might already be working in Ruby)

marshall007:
Fixes #4438.

### DIFF
--- a/drivers/python/rethinkdb/ast.py
+++ b/drivers/python/rethinkdb/ast.py
@@ -1486,7 +1486,7 @@ class Sync(RqlMethodQuery):
     st = 'sync'
 
 
-class Branch(RqlMethodQuery):
+class Branch(RqlTopLevelQuery):
     tt = pTerm.BRANCH
     st = "branch"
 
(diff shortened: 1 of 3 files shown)


drivers/python/rethinkdb/ast.py
```diff
@@ -1483,7 +1486,7 @@ class Sync(RqlMethodQuery):
     st = 'sync'
 
 
-class Branch(RqlTopLevelQuery):
+class Branch(RqlMethodQuery):
```
danielmewes: I think this only determines how the term is pretty printed. I'd prefer keeping the old behavior here.

danielmewes:
Thanks for the pull request @marshall007 !
I added one comment, but it looks good otherwise.

drivers/python/rethinkdb/ast.py
marshall007: Ok, likewise in the JS driver should I revert [this line](https://github.com/rethinkdb/rethinkdb/pull/4442/files#diff-029874a8dbb2a1f773b7980319d494b5R978)?

danielmewes: Ah, yes please.

danielmewes:
Sorry it took me a while to get back to this. Looks good, thank you @marshall007 !

Issue #4451, opened 2015-06-25
Error in Import on Big json file


sebadiaz:
When I have a big json file an error occur during the parsing phase because the json file is more than the buffer size
misys@ubuntu:~/project/cnam/airbase$ rethinkdb import -f data/AT_meta.xml.json --table test.users --force --debug 
[                                        ]   0% 
0 rows imported in 1 table
Expecting object: line 957981 column 11 (char 67108863)
ValueError traceback: [('/usr/local/lib/python2.7/dist-packages/rethinkdb/_import.py', 548, 'table_reader', 'exit_event)'), ('/usr/local/lib/python2.7/dist-packages/rethinkdb/_import.py', 420, 'json_reader', 'json_data[offset] == "[")'), ('/usr/local/lib/python2.7/dist-packages/rethinkdb/_import.py', 371, 'read_json_array', '(obj, offset) = decoder.raw_decode(json_data, idx=offset)'), ('/usr/lib/python2.7/json/decoder.py', 382, 'raw_decode', 'obj, end = self.scan_once(s, idx)')]
In file: /home/misys/project/cnam/airbase/data/AT_meta.xml.json
Errors occurred during import

danielmewes:
Thanks for the report and pull request (https://github.com/rethinkdb/rethinkdb/pull/4452) @sebadiaz !

I'm wondering if there's a way to avoid having to set the buffer size manually for large files. I think that isn't very user friendly. Could we adapt it automatically to the file size, or use some different data structure that grows automatically? I don't know much about what that buffer is currently used for in the import script unfortunately.
Pinging @Tryneus .

Tryneus:
I suspect this isn't the reason for the failure.  The problem isn't the size of the file, but the size of the rows.  This failure occurs when we have buffered 128 MB of the file and can't parse out a valid json object from it.  I'm fairly sure we can't import rows that large (due to protobuf limits).  I think it's more likely that the JSON file has a syntax error in it.

danielmewes:
Oh... I think our size limit is still 64 MB per query right?

@sebadiaz Were you actually able to import the affected json file with your patch?

sebadiaz:
I suspect an other error, my case is easy I have just one file containing one json structure and not starting with a list. I m not sure that having a limit in term of row, more exactly a json object in multiple file line, is the real error. Perhaps,it is the mix between [{ and not [{ . effectively on the same json I have some json array , sometime in big quantity.
[...]

PR #4452, opened 2015-06-25
make json buffer and chunk size as argument parameters #4451

sebadiaz:
On a big file the buffer is not sufficient. I add argument parameter to change it.
I add too for the chunk size to accelerate import on big file

### DIFF
--- a/drivers/python/rethinkdb/_import.py
+++ b/drivers/python/rethinkdb/_import.py
@@ -12,8 +12,10 @@ import rethinkdb as r
 # Used because of API differences in the csv module, taken from
 # http://python3porting.com/problems.html
 PY3 = sys.version > '3'
-json_read_chunk_size = 32 * 1024
 
+#json parameters
+json_read_chunk_size = 32 * 1024
+json_max_buffer_size = 128 * 1024 * 1024
 try:
     import cPickle as pickle
 except ImportError:
@@ -73,6 +75,10 @@ def print_import_help():
     print("  --custom-header FIELD,FIELD...   header to use (overriding file header), must be")
     print("                                   specified if --no-header")
     print("")
+    print("Import JSON format:")
+    print("  --max-document-size              the maximum size in bytes that a single JSON document")
+    print("                                   can have (defaults to 134217728).")
+    print("")
     print("EXAMPLES:")
     print("")
     print("rethinkdb import -d rdb_export -c mnemosyne:39500 --clients 128")
@@ -105,6 +111,7 @@ def parse_options():
     parser.add_option("--hard-durability", dest="hard", action="store_true", default=False)
     parser.add_option("--force", dest="force", action="store_true", default=False)
     parser.add_option("--debug", dest="debug", action="store_true", default=False)
+    parser.add_option("--max-document-size", dest="max_document_size",  default=0,type="int")
 
     # Directory import options
     parser.add_option("-d", "--directory", dest="directory", metavar="DIRECTORY", default=None, type="string")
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks for the pull request, really appreciate it.

As mentioned in https://github.com/rethinkdb/rethinkdb/issues/4451#issuecomment-115360131 , I'd like to avoid exposing this option if possible. Let's see if @Tryneus has an idea.

On a different note: @sebadiaz have you signed our CLA yet? We require this for legal reasons before we can merge pull requests into the main code base. You can easily sign it online at http://rethinkdb.com/community/cla/ .

Tryneus:
Please see my comment at https://github.com/rethinkdb/rethinkdb/issues/4451#issuecomment-115428306.

sebadiaz:
The reason appear more clear on a big document when an arry is not finished we extend the file reading to continue on the array parsing
code extracted 
before_len = len(json_data) -> we take the size of the actual structure
to_read = max(json_read_chunk_size, before_len) -> we get this previous measure as max
json_data += file_in.read(min(to_read, json_max_buffer_size - before_len)) -> finally we use json_max_buffer_size - before_len because < to_read --> could be negative on non raising step(see next lines)

after reading we have
elif before_len == len(json_data) or len(json_data) >= json_max_buffer_size: --> this is the error , we can't raise on unfinished array 
                raise


Two solutions
  - set a big json_max_buffer_size with two solutions : 
	- fixed at starting
	- set to the file size
  - continue to read unless size limit exceed my new patch proposition in removing json_max_buffer_size
	except (ValueError, IndexError):
            before_len = len(json_data)
            to_read = max(json_read_chunk_size, before_len)
            json_data += file_in.read(to_read)
            if json_array and json_data[offset] == ",":
                offset = json.decoder.WHITESPACE.match(json_data, offset + 1).end()
            elif (not json_array) and before_len == len(json_data):
                break  # End of JSON
            elif before_len == len(json_data):
                raise
            progress_info[0].value = file_offset

sebadiaz:
CLA is signed

danielmewes:
Thanks for the update @sebadiaz . I like this better.
@Tryneus for what reasons do we have the `json_max_buffer_size` in the first place? Is it to avoid using too much memory if a JSON file is faulty or something else?

Tryneus:
@danielmewes - yeah, the idea is that if a user provides the incorrect file or a JSON file with syntax errors, we don't want to go into swap or hit the OOM killer.  There's a chance this will happen even with the 128 MB limit (because we can import multiple tables in parallel), but that number was chosen because it is likely larger than any individual row that can be successfully imported (barring a JSON file with a high ratio of whitespace).

We would like to avoid the OOM killer because if the import is running on the same machine as the database server, there is a good chance the OOM killer will take out the server before the import process.

danielmewes:
Hmm those are certainly valid concerns.
I think the problem might be JSON files that have indentation for each line. In that case the size of a document in the file can easily be a multiple of the actual document size. @sebadiaz is that actually the case in your input or are the whitespaces coming from something else?

Given this information, maybe a command line option like `--max-document-size` to set `json_max_buffer_size` as @sebadiaz had originally implemented isn't the worst idea. I'd prefer that name over `--change-max-buffer-size` since it's less cryptic.
@Tryneus do you think that would be fine?

I'd leave out the `read-chunk-size` flag though.

Tryneus:
@danielmewes, that's reasonable, I would be ok with that.

danielmewes:
Ok sounds good.

Sorry @sebadiaz for the back and forth. I'm going to add some comments on your first commit about the changes I think we need to do, and @Tryneus should probably do a quick pass as well. Thanks again for your help.

sebadiaz:
Done. I hope to be the last round

Tryneus:
Looks good, thanks for your patience, @sebadiaz!

danielmewes:
Thank you @sebadiaz. @Tryneus do you want to push the button or should I?

Tryneus:
@danielmewes - go ahead, technically I'm PTO today.

danielmewes:
Merged :-)

Issue #4472, opened 2015-06-30
Support multiple error types in the Python and JS clients


mlucy:
https://github.com/rethinkdb/rethinkdb/issues/4296 implemented server support, but we have to actually produce multiple exception types on the clients.  (I already did this for Ruby.)

danielmewes:
Moving to 2.1. I think we should support this across all official drivers.

marshall007:
I have a pretty good start on this for the JS driver in [`marshall007/rethinkdb#marshall_4472_js`](https://github.com/marshall007/rethinkdb/tree/marshall_4472_js). Haven't tested yet since they need to be run against 2.1.x server.

marshall007:
I wasn't able to build the server binary off of `next` to test against:

```
    ...
    [317/397] CC build/release_clang/obj/rdb_protocol/query_server.o
src/rdb_protocol/query_server.cc:56:65: error: too few arguments to function call, expected 5, have 4
                       ql::backtrace_registry_t::EMPTY_BACKTRACE);
                                                                ^
./src/rdb_protocol/backtrace.hpp:72:1: note: 'fill_error' declared here
void fill_error(Response *res_out,
^
1 error generated.
make[1]: *** [build/release_clang/obj/rdb_protocol/query_server.o] Error 1
make: *** [make] Error 2
```

Would it be easier if I just did this directly on top of @mlucy's changes (i.e. 1513a25)?

mlucy:
@marshall007 -- it would probably be slightly better to do it off of `next`.  It looks like I introduced the bug myself in the error types merge; just fixed it and pushed, so `next` should compile fine in release mode now.

marshall007:
Ok, thanks. I'll rebase and give it another go.

marshall007:
@mlucy I didn't come across any `RqlOpIndeterminateError` exceptions and want to make sure that wasn't me. Take a look at my comment over there when you get a chance.

neumino:
Are the new errors just sub-types of a `ReqlRuntimeError`?
Can we have a quick description of each type? #4296 doesn't say much about what was implemented

marshall007:
@neumino pretty much but there's a couple cases of multiple inheritance. Take a look at the [relevant lines](https://github.com/marshall007/rethinkdb/blob/marshall_4472_js/drivers/javascript/errors.coffee#L24-L38) from my PR.

danielmewes:
[...]

PR #4483, opened 2015-07-01
JS Driver: Add support for new exception types

marshall007:
See #4472. Includes updates to all polyglot tests, so testing the updates to Python and Ruby drivers should be easier now.

@mlucy there were no existing tests that returned an `RqlOpIndeterminateError`. I'm not sure if that's an oversight on my part or if there simply aren't any existing cases that cover that.

I was planning on running the Ruby tests to make sure all the changes to the polyglot tests line up, but I'm getting:

```
$ ./test/rql_test/test-runner -i rb
Usage: test-runner [options] [patterns]

test-runner: error: Unable to find a valid interpreter for language: rb
```

Also there are a few failing connections tests in the JS driver right now. Two of them are timeouts, and the others are "could not connect to server" errors for `r.http`. So there may be a couple tests in there that still need to be updated, but I was unsure of why they were failing to begin with.


mlucy:
Awesome @marshall007 !

> @mlucy there were no existing tests that returned an RqlOpIndeterminateError. I'm not sure if that's an oversight on my part or if there simply aren't any existing cases that cover that.

That's correct, `OpIndeterminate` is only produced by events that cause internal disconnects in the cluster which the polyglot tests aren't set up to produce right now.

---

Assigning to @deontologician since he seems best able to review it (if that's alright?).

marshall007:
@mlucy great, thanks!

marshall007:
@mlucy oh one another thing was there were a couple weird discrepancies like `r.minval` produces `RqlInternalError` whereas `r.expr(r.minval).coerce_to('string')` produces `RqlLogicError` both with the same error message: `Cannot convert r.minval to JSON.`

mlucy:
I think the first probably shouldn't be an internal error; I'll fix that.

Did you see any other weird ones like that?

marshall007:
`table.index_create('id')` produced `RqlLogicError` instead of `RqlOpFailedError`, but after thinking about it that one probably makes sense. The only other one I remember was invalid optarg `array_limit: -1` producing `RqlInternalError` vs deprecated optarg `return_vals` producing `RqlLogicError`.

deontologician:
This looks good to me, so I think we can merge this once the changes corresponding to #4484 are made

drivers/javascript/errors.coffee
```diff
@@ -24,9 +24,19 @@ class RqlServerError extends Error
 class RqlRuntimeError extends RqlServerError
 
 class RqlCompileError extends RqlServerError
-
 class RqlClientError extends RqlServerError
 
+class RqlInternalError extends RqlRuntimeError
+class RqlResourceError extends RqlRuntimeError
+class RqlLogicError extends RqlRuntimeError
+class RqlNonExistenceError extends RqlRuntimeError
```
mlucy: A `NonExistenceError` is a type of `LogicError` (currently it's the only type we report on its own), so it should probably extend `RqlLogicError` instead.

mlucy: (This probably could be clearer in the protobuf, I tried to encode the hierarchy in the numbers used for the error types but it probably came out pretty opaque.)

marshall007: @mlucy thanks, just pushed a fix.

danielmewes:
Thinking about this again, I think we should merge this regardless of #4484 so we can include it in the 2.1 beta. If the tests need to be adapted, we can do that after #4484 gets in. (@deontologician)

Do the test changes currently in this pull request assume the current behavior or the one with #4484 fixed?

marshall007:
@danielmewes test changes assume current behavior.

danielmewes:
Great, then - assuming everything else is right - I think we should merge this.

marshall007:
The only thing that hasn't been addressed here is the failing connections tests (quoting myself):

> Also there are a few failing connections tests in the JS driver right now. Two of them are timeouts, and the others are "could not connect to server" errors for r.http. So there may be a couple tests in there that still need to be updated, but I was unsure of why they were failing to begin with.

Part of that is probably #4496, but also some tests in `connections/http.js` weren't updated because I couldn't get them to run. I think it has something to do with httpbin being proxied and that wasn't working for me.

deontologician:
OK, I have a patch that will fix the remaining things needed in connection/http.js, so I'm going to merge this and apply that directly. Thanks again for the PR @marshall007

Issue #4475, opened 2015-06-30
RethinkDB driver breaks Node.js domain


svarogg:
It looks like calling .run breaks the domain in which Node.js is, invalidating all data that was stored in it.

A quick-fix looks like wrapping the callback in ast.js/run with the following function:
https://gist.github.com/svarogg/1552cd4e9dd675494021

I don't know coffeescript, and don't have the time to fully study the consequences, so am not submitting a pull request at this time.

Thanks for your assistance.

mlucy:
Ping @deontologician.

deontologician:
> calling .run breaks the domain in which Node.js is, invalidating all data that was stored in it.

@svarogg could you give an example showing what you're running into?

svarogg:
@deontologician I have initially encountered this problem when using the loopback rethinkdb connector. I have opened a ticket in the connector's repository: https://github.com/fuwaneko/loopback-connector-rethinkdb/issues/7 but after some investigation I have seen that the fix could be in the driver (all details in the link).

Nevertheless, when writing a shorter example to demonstrate this issue, I have seen that without the connector things work:
https://gist.github.com/perfectedTech/12e8156e9e882057ca0a

I must admit that I no longer fully understand the problem, and nodejs domains are new to me, but the the conjunction of loopback and rethinkdb definitely creates a problem.

Please advise. If you believe this is not a driver problem, this can be fixed inside the connector.
Thanks.

deontologician:
I'm not familiar with node domains myself, but I'm a bit wary of modifying the code if there isn't an identifiable error. I can look into it more deeply later and make a more informed decision on it. For now I'm going to put it in the backlog, but if you are able to reproduce the error with the driver let me know and we can bump up the priority.

Thanks for reporting @svarogg

svarogg:
@deontologician Hey there.
After some more research I was able to demonstrate this problem with a simple script, with only rethinkdb as a dependency.
https://gist.github.com/svarogg/7d3b0ac47f1833e93a27

neumino:
This is actually an issue "in" bluebird
https://github.com/petkaantonov/bluebird/pull/664

svarogg:
Looks like this was fixed in bluebird.

deontologician:
OK, so it looks like we should update our bluebird dependency to > 2.9.31 to get these benefits.

larkost:
[...]

PR #4493, opened 2015-07-02
Bind .run's callback to current domain(issue #4475).

svarogg:
In relation to https://github.com/rethinkdb/rethinkdb/issues/4475

### DIFF
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -114,7 +114,7 @@ class TermBase
                 # letting the error bubble up.
                 wrappedCb(e)
 
-        ).nodeify callback
+        ).nodeify(util.bindToCurrentDomain callback)
 
     toString: -> err.printQuery(@)
 
--- a/drivers/javascript/util.coffee
+++ b/drivers/javascript/util.coffee
@@ -54,6 +54,13 @@ module.exports.toCamelCase = (token) ->
         match[1].toUpperCase()
     )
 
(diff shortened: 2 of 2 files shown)


danielmewes:
Thanks for the pull request @svarogg !
If you haven't done so yet, would you mind signing our CLA? You can do so at http://rethinkdb.com/community/cla/

@deontologician do you think this is the right way to fix this? Or should we wait for a fix in Bluebird? https://github.com/rethinkdb/rethinkdb/issues/4475

svarogg:
@danielmewes Yeah, I have signed the CLA already.

deontologician:
Ok, so I've read up on domains now. It looks like this should be fine, but given that the bluebird issue is being actively worked on, let's give see if they fix it soon. If not we can merge this. I think ideally upstream fixing it would be better

deontologician:
Also, if we do go with this, we should add some tests. I extended @svarogg 's example a bit here, showing it fails with callbacks and promises:

https://gist.github.com/deontologician/89f0306a804d9062af65

svarogg:
This fix is more or less the thing the guys at mongodb did.

deontologician:
Since bluebird has fixed this (see https://github.com/rethinkdb/rethinkdb/issues/4475#issuecomment-118994291) I'm closing this PR in favor of bumping the bluebird version. Thanks @svarogg for bringing this issue to our attention

PR #4534, opened 2015-07-14
Only set __proto__ in TermBase if it exists

### LINKED ISSUE #162: [JS Driver] Support IE 10
So we don't know why Internet Explorer 10 doesn't support the JavaScript Driver.
IE 10 supports DataView and Uint8Array so it should work...

Tagging in backlog.

deontologician:
IE 10 doesn't have the `__proto__` property. So for JS interpreters without `__proto__` we skip the `__proto__` magic (thereby disabling the `foo('bracket')` syntax in ReQL).

Fixes #162
Fix suggested by @mikemintz

### DIFF
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -52,9 +52,12 @@ hasImplicit = (args) ->
 class TermBase
     showRunWarning: true
     constructor: ->
-        self = (ar (field) -> self.bracket(field))
-        self.__proto__ = @.__proto__
-        return self
+        if {}.__proto__ != undefined
+            self = (ar (field) -> self.bracket(field))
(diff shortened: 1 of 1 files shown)


danielmewes:
@mikemintz since this is your code, would you mind signing our CLA so we can use it in our official driver? http://rethinkdb.com/community/cla/

mikemintz:
@danielmewes I just signed the CLA. Thanks everyone for resolving this so quickly.

danielmewes:
Thanks @mikemintz . I'm not sure if we can get it into 2.1, but we'll see.

neumino:
You can probably avoid creating a new object and just checking this.__proto__
That being said, it's a harmless change.

Issue #4526, opened 2015-07-10
JavaScript `noReplyWait` connection test timing out on select computers


larkost:
On two computers, on current `next` I am getting a failure one two of the subtests in `connections/connection.js`, specifically:
- `connection close does not wait if we want it not to`
- `connection reconnect does not wait if we want it not to`

The subtests time out, indicating that the post-call function is not being called. I disabled the setting of `BLUEBIRD_DEBUG` so this is probably not a repeat of #4496, but it could well be related to the recent update to `bluebird`.

I have a [smaller test](https://gist.github.com/larkost/99c526c1d171caa4096e) that demonstrates this, as well as an improved version of the test that adds more stable versions of these tests as well as new versions that demonstrate this:
- `connection close noreplyWait works even with outstanding noreplyWait calls`
- `connection reconnect noreplyWait works even with outstanding noreplyWait calls`

So far this only reproduces on my two computers (home and work), and only on `next` (so not `2.0.4`). Isolating the problem I found that it only occurs in `mocha` tests where the tests is inside (at least) two levels of `describe`. This leads me to believe that this is a very narrow timing issue, and is probably going to be really nasty to debug. I do think that this is a valid problem and could reach out and bite someone.

danielmewes:
I suspect that this might be a race condition in the driver. The test submits a query but doesn't wait until it has been sent over the wire. Then it tries to close the connection, but the `close` callback never gets called.

@deontologician could you double-check please that the logic in the driver for closing connections looks sane given these circumstances?

deontologician:
There is a change I can make to how callbacks are added to the existing promise. I'll make the change and maybe once Karl is back he can replicate it. I was unable to replicate (as expected) on newton, even with the improved tests.

deontologician:
Review is available in #4535

deontologician:
Previous PR didn't fix the issue. I'll try something else in a bit.

larkost:
A slightly improved test is in `next` with 1f07837. It separates the simultaneous `noreply` queries case from the outstanding one. However, this still only demos on my computers.

deontologician:
Ok, I'm going to try to replicate this on the MacInShelf, hopefully I can narrow it down

deontologician:
[...]

PR #4535, opened 2015-07-14
Leave the _closePromise in place

deontologician:
When a callback is chained onto an existing
already-resolved promise, the callback will be invoked
immediately. So we don't need to work around this
corner case.

Related to #4526


danielmewes:
@larkost have you had a chance to test this? Does it fix the problem?

larkost:
The change in `josh/4526` is still failing in the same way on my machine. Sorry.

Issue #4569, opened 2015-07-23
Optimize the python driver's serialization code.


sontek:
Currently if you want to work with 100,000 records it is extremely slow:

```
$ python foo.py 
count is 164717
int query took 4086ms
```
4 seconds to query 164,000 rows!  I dug into this a little bit using the python library `yappi` and the tool `kcachegrind` to see where it was spending the time.  Here is the python code I used:

```python
import rethinkdb as r
import random
import pytz
import time
import yappi
from datetime import datetime

UTC = pytz.utc

YAPPI = False
DO_CREATE = False
DO_INSERTS = False

conn = r.connect()

if DO_CREATE:
    r.table_create('test').run(conn)

START_BIGINT = 100000000000000000
END_BIGINT = 999999999999999999


def utc_now():
    now = datetime.utcnow()
    tz_now = now.replace(tzinfo=UTC)
    return tz_now


def get_rint(start=1000000, end=9999999):
    """
    Generate a very large integer
    :return:
    """
    return random.randint(start, end)


def get_bigint():
    """
    Generate a random BIGINT
    :return:
    """
    return get_rint(start=START_BIGINT, end=END_BIGINT)


if DO_INSERTS:
    for i in range(0, 300000):
        r.table('test').insert({
            'respondent_id': get_bigint(),
            'row_id': get_bigint(),
            'column_id': get_bigint(),
            'value_id': get_bigint(),
            'rid1': get_bigint(),
            'rid2': get_bigint(),
            'rid3': get_rint(),
            'now': utc_now()
        }).run(conn)

start = time.time()

if YAPPI:
    yappi.set_clock_type('cpu')
    yappi.start(builtins=True)

result = r.table('test').run(conn)
data = []
for row in result:
    data.append(row)

count = len(data)
end = time.time()

if YAPPI:
    stats = yappi.get_func_stats()
    stats.save('callgrind.out', type='callgrind')
    print('checkout callgrind.out')

print("count is %s" % count)
duration = int(1000 * (end - start))
print("int query took %sms" % duration)
```

and this showed that most of the time is spent in `recursively_convert_pseudo_types`, here is a screenshot of the kcachegrind:

![screenshot from 2015-07-23 13 06 13](https://cloud.githubusercontent.com/assets/151924/8861344/cba7a5cc-313e-11e5-8588-a630ee983175.png)

danielmewes:
Thank you for the script and data @sontek !
We'll look into it and see if we can optimize that function.

[...]

Issue #3506, opened 2015-01-06
Allow specifying a custom JSONDecoder class in the python driver


deontologician:
[Currently](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/rethinkdb/net.py#L61) we just use `json.loads` to deserialize json in the python driver. But it would be nice to allow some way for users to provide a custom [JSONDecoder](https://docs.python.org/2/library/json.html#json.JSONDecoder) subclass that allows customizing the deserialized result.  This would be useful, for example, for ORM writers who would like to instrument the objects coming out of the DB without having to reconvert or wrap them in a second step

deontologician:
I'm leaving open what the api for this should be, since it isn't obvious to me right now what's the best way

grandquista:
I have implemented an `r.json_parser` property that can be reassigned to override all serialization in the Lua driver. It seems like this would be useful to customize in any driver, not just Python.

deontologician:
Having a global like that seems the simplest way, but I think it might be better to specify it at the connection level

danielmewes:
We have to keep in mind though that the driver itself has to understand parts of the decoded JSON. Most importantly it has to be able to check the response type (error, cursor, single document).

deontologician:
That's why I think this should be more of a thing that's possible, vs. something exposed to regular users. ORM creators would be testing that it doesn't break the driver.

danielmewes:
I see, that makes sense. I imagine we could just make it an opt arg to `r.connect`?

danielmewes:
Actually that might be a bad idea because it would expose it a bit too prominantly.

One option might be to make the connection class such that an ORM developer can easily derive their own custom version of it, with some json decoding method overwritten.

deontologician:
that sounds like it might work

brianjpetersen:
@danielmewes

> One option might be to make the connection class such that an ORM developer can easily derive their own custom version of it, with some json decoding method overwritten.

If done thoughtfully, +1.  This would also address #4503.

danielmewes:
This is now possible with https://github.com/rethinkdb/rethinkdb/pull/4585

Issue #1016, opened 2013-06-14
ReQL driver proposal: `r.registerType`


coffeemug:
``` python
r.register_type(CustomType,
    # serializer
    lambda obj: r.json(json.dumps(obj)),
    # deserializer
    lambda dict: CustomType.from_dict(dict))
)
```

Then:

``` python
obj = CustomType()
r.table('foo').insert(obj)
r.table('foo').get(some_id) # returns CustomType
```

It's very easy to do the serializer, but harder to do the deserializer because it's unclear how to represent type tags properly. I'd suggest doing the serializer first and leaving the deserializer for later.

The use case for this is outlined in #241.

PR #4585, opened 2015-07-27
Create a JSON decoder subclass with a hook for pseudo types.

grandquista:
This is my proposed solution for #3506 and #4569

### DIFF
--- a/drivers/python/rethinkdb/ast.py
+++ b/drivers/python/rethinkdb/ast.py
@@ -693,7 +693,7 @@ class ReQLDecoder(py_json.JSONDecoder):
         py_json.JSONDecoder.__init__(self, object_hook=self.convert_pseudotype)
         self.reql_format_opts = reql_format_opts
 
-    def convert_type_time(self, obj):
+    def convert_time(self, obj):
         if 'epoch_time' not in obj:
             raise ReqlDriverError(('pseudo-type TIME object %s does not ' +
                                    'have expected field "epoch_time".')
--- a/drivers/python/rethinkdb/ast.py
+++ b/drivers/python/rethinkdb/ast.py
@@ -686,7 +686,9 @@ def recursively_make_hashable(obj):
 
 
 class ReQLDecoder(py_json.JSONDecoder):
-    """docstring for name"""
+    '''
+        Default JSONDecoder subclass to handle pseudo-type conversion.
+    '''
     def __init__(self, reql_format_opts={}):
         py_json.JSONDecoder.__init__(self, object_hook=self.convert_pseudotype)
         self.reql_format_opts = reql_format_opts
--- a/drivers/python/rethinkdb/ast.py
+++ b/drivers/python/rethinkdb/ast.py
@@ -670,19 +670,6 @@ class RqlTzinfo(datetime.tzinfo):
         return datetime.timedelta(0)
 
 
-def reql_type_time_to_datetime(obj):
-    if 'epoch_time' not in obj:
-        raise ReqlDriverError(('pseudo-type TIME object %s does not ' +
-                               'have expected field "epoch_time".')
-                              % py_json.dumps(obj))
-
-    if 'timezone' in obj:
-        return datetime.datetime.fromtimestamp(obj['epoch_time'],
-                                               RqlTzinfo(obj['timezone']))
-    else:
-        return datetime.datetime.utcfromtimestamp(obj['epoch_time'])
-
-
 # Python only allows immutable built-in types to be hashed, such as
 # for keys in a dict This means we can't use lists or dicts as keys in
 # grouped data objects, so we convert them to tuples and frozensets,
@@ -698,22 +685,6 @@ def recursively_make_hashable(obj):
     return obj
 
 
-def reql_type_grouped_data_to_object(obj):
-    if 'data' not in obj:
-        raise ReqlDriverError(('pseudo-type GROUPED_DATA object' +
-                               ' %s does not have the expected field "data".')
-                              % py_json.dumps(obj))
-    return dict([(recursively_make_hashable(k), v) for k, v in obj['data']])
-
-
-def reql_type_binary_to_bytes(obj):
-    if 'data' not in obj:
-        raise ReqlDriverError(('pseudo-type BINARY object %s does not have ' +
-                               'the expected field "data".')
-                              % py_json.dumps(obj))
-    return RqlBinary(base64.b64decode(obj['data'].encode('utf-8')))
-
-
 class ReQLDecoder(py_json.JSONDecoder):
     """docstring for name"""
     def __init__(self, reql_format_opts, *args, **kw):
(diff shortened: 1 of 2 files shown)


danielmewes:
Very cool. Thanks for the pull request @grandquista!
We're going to take a look at this soon.

mlucy:
@danielmewes --should this be in a milestone?

danielmewes:
2.1.x I think once we have that. Putting in 2.1-polish for now.

deontologician:
Tangentially related: #1016

grandquista:
#1016 looks like a duplicate of #3506. The proposal in #1016 seems to ask for #3506 to be done on the server.

danielmewes:
They are similar, but #1016 would provide a more convenient interface for specifying individual additional types, that would also work across drivers (not just on Python). I think the main work for #1016 would be on the client, but the server might need some minor additions as well to detect the fact that a certain object is actually a user-defined pseudo type.

danielmewes:
I'm getting "Unexpected response received" on some queries:
```py
Python 2.7.3 (default, Jun 22 2015, 19:33:41) 
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import rethinkdb as r
>>> r.connect().repl()
<rethinkdb.net.DefaultConnection object at 0x7f12a1b75a10>
>>> r.table('t1').insert({}).run()
  File "<stdin>", line 1, in <module>
  File "rethinkdb/ast.py", line 118, in run
    return c._start(self, **global_optargs)
  File "rethinkdb/net.py", line 565, in _start
    return self._instance.run_query(q, global_optargs.get('noreply', False))
  File "rethinkdb/net.py", line 426, in run_query
    res = self._read_response(query.token)
  File "rethinkdb/net.py", line 475, in _read_response
    raise RqlDriverError("Unexpected response received.")
rethinkdb.errors.RqlDriverError: Unexpected response received.
>>>
```

Any ideas @grandquista ?

drivers/python/rethinkdb/net.py
```diff
@@ -462,16 +458,18 @@ def _read_response(self, token, deadline=None):
                 self._parent.reconnect(noreply_wait=False)
                 raise ex
 
-            # Construct response
-            res = Response(res_token, res_buf)
+            res = None
 
-            cursor = self._cursor_cache.get(res.token)
+            cursor = self._cursor_cache.get(res_token)
             if cursor is not None:
```
danielmewes: This code needs to handle the case where the result is not for a cursor (e.g. `r.expr(1)`).

danielmewes:
Never mind, found the problem I think (I added a comment in the code).

danielmewes:
I hacked in a quick fix, and ran the benchmark from https://github.com/rethinkdb/rethinkdb/issues/4569#issue-96898802 .
Even with the default JSON deserializer, these changes make loading 300,000 documents more than twice as fast (7.1 seconds vs. 14.4 seconds)
This is really awesome @grandquista  :-)

@Tryneus I think this might be next after migration?

drivers/python/rethinkdb/ast.py
```diff
@@ -698,66 +685,72 @@ def recursively_make_hashable(obj):
...
-                                     % binary_format)
+class ReQLDecoder(py_json.JSONDecoder):
+    '''
+        Default JSONDecoder subclass to handle pseudo-type conversion.
+    '''
+    def __init__(self, reql_format_opts={}):
+        py_json.JSONDecoder.__init__(self, object_hook=self.convert_pseudotype)
+        self.reql_format_opts = reql_format_opts
+
+    def convert_type_time(self, obj):
```
danielmewes: This should be `convert_time` right? At least it's called by that name in line 728

danielmewes:
Actually my measurements were with a slightly modified version of the benchmark from https://github.com/rethinkdb/rethinkdb/issues/4569#issue-96898802 that used timestamps and not ReQL time objects in the table. With real time objects, your patch is still much faster though: ~18 seconds instead of ~25 seconds.

deontologician:
I'm excited, this is potentially fixes several longstanding issues. Nice work! (I've looked over the patch as well, and the only things I noticed Daniel already mentioned)

drivers/python/rethinkdb/ast.py
grandquista: Yes, that would be consistent with how I was naming the other convert methods.

drivers/python/rethinkdb/net.py
grandquista: I have a patch to handle the first call to `_read_response`.

Tryneus:
Ok, finally got around to this.  Looks good for the most part, although it doesn't change the other `net.py` implementations - I can take care of that when this is merged, though.

Tryneus:
Thanks for all the help, @grandquista, merged!

danielmewes:
@Tryneus once you are done transferring this to the other connection backends, please also backport this to `v2.1.x`.

Tryneus:
The changes were applied to the `v2.1.x` branch in commit 097b05b.

danielmewes:
For reference, here's an example for using UltraJSON in the driver (https://github.com/esnme/ultrajson):

```py
import rethinkdb as r
import ujson

class UjsonDecoder(object):
    def __init__(self, reql_format_opts={}):
        pass
    def decode(self, s):
        return ujson.loads(s)

def create_decoder(format_opts):
    return UjsonDecoder(format_opts)

conn = r.connect()
conn._get_json_decoder = create_decoder

# `conn` now uses UltraJSON to parse server responses
```

Note that UltraJSON doesn't currently allow registering a hook for processing pseudo types. So RethinkDB pseudo types (binary data, date/time values) will not be converted into native Python types in this case.

danielmewes:
The change was included in the latest Python driver release, "2.1.0.post1"

sontek:
@danielmewes Seems like an odd API to override a private method on the connection, could it be exposed as a public API?

coffeemug:
@sontek -- I think we should use the API you proposed originally:

```py
# Specify default encoder/decoder on the connection
r.connect(..., json_encoder=ujon.loads, json_encoder=ujson.dumps)

# Override for a particular query
query.run(..., json_encoder=jon.loads, json_encoder=json.dumps)
```

We didn't add it yet because the existing version is pretty simple, and we wanted to get it out fast. I can open a new issue for this if people think the API proposed above is reasonable.

danielmewes:
For reference: @grandquista put an implementation of the simplified API up as a pull request: https://github.com/rethinkdb/rethinkdb/pull/4825

We're planning to include that with RethinkDB 2.2.

drivers/python/rethinkdb/ast.py
```diff
@@ -698,66 +685,73 @@ def recursively_make_hashable(obj):
...
-            if binary_format is None or binary_format == 'native':
-                return reql_type_binary_to_bytes(obj)
-            elif binary_format != 'raw':
-                raise ReqlDriverError("Unknown binary_format run option \"%s\"."
-                                      % binary_format)
+class ReQLDecoder(py_json.JSONDecoder):
+    '''
+        Default JSONDecoder subclass to handle pseudo-type conversion.
+    '''
+    def __init__(self, reql_format_opts={}):
```
sontek: This is making a mutable global object at import time.  This should be something like this instead:

```python
def __init__(self, reql_format_opts=None):
    self.reql_form_opts = reql_format_opts or {}
```

sontek: cc @danielmewes @grandquista

sontek:
I was able to shave off 50% of my time using this API:

```
# standard json
$ python foo_rethinkdb.py 
count is 164717
int query took 2213ms

# using rapidjson
$ python foo_rethinkdb.py 
count is 164717
int query took 1332ms
```

drivers/python/rethinkdb/ast.py
sontek:
```python
>>> from rethinkdb.ast import ReQLDecoder
>>> one = ReQLDecoder()
>>> two = ReQLDecoder()
>>> one.reql_format_opts['boom'] = 'shaka'
>>> one.reql_format_opts
{'boom': 'shaka'}
>>> two.reql_format_opts
{'boom': 'shaka'}
```

danielmewes: Oh good catch! Do you want to open a pull request? (otherwise I'm just going to copy over your patch, if that's ok).

PR #4625, opened 2015-08-04
Avoid certain issues when using EventMachine and fibers

nviennot:
This avoid this error:

```
  RuntimeError: can't add a new key into hash during iteration
     # lib/net.rb:480:in `block in register_query'
     # lib/net.rb:476:in `register_query'
     # lib/net.rb:515:in `run'
     # lib/net.rb:341:in `em_run'
```

### DIFF
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -715,7 +715,7 @@ module RethinkDB
       @mon.synchronize {
         @opts.clear
         @data.clear
-        @waiters.each {|k,v|
+        @waiters.dup.each {|k,v|
           case v
           when QueryHandle
             v.handle_close
@@ -755,7 +755,7 @@ module RethinkDB
 
     def remove_em_waiters
       @mon.synchronize {
-        @waiters.each {|k,v|
+        @waiters.dup.each {|k,v|
           if v.is_a? QueryHandle
             v.handle_close
             @waiters.delete(k)


danielmewes:
@mlucy could you take a look at this one as well?

mlucy:
@nviennot -- what case is causing that error to occur?

I'm a little wary of this change.  If something is getting added to `@waiters` while we're iterating over it and closing everything, that suggests that with the new code something will get added to `@waiters` during the iteration and just never be told that the connection is closed.

nviennot:
If you run the following script, it will give you that exception:

```ruby
#!/usr/bin/env ruby
require 'rethinkdb'
require 'eventmachine'
include RethinkDB::Shortcuts

r.connect(:host => 'localhost', :port => 28015, :db => :test).repl

class UserHandler < RethinkDB::Handler
  def on_close
    r.table(:table).em_run(UserHandler)
  end
end

EM.run do
  r.table(:table).em_run(UserHandler)
  raise
end
```

Note that having on_close() issuing another query may seem weird, but it's not. Esp. if on_close() resumes another fiber, changing the context.

mlucy:
@nviennot -- the two lines you changed are in `Connection::close` and `Connection::remove_em_waiters`.  `Connection::close` stops listening on the socket, and `remove_em_waiters` is only called when EventMachine is shutting down, so in both cases I think the result of the change is that instead of throwing an exception you just silently start a query whose result you will never receive.

I think the root problem here is that there's no way to know in `on_close` whether or not the handler is being closed because the connection is going away or for some other reason, so there's no way to know whether or not it's safe to run another query.  (Of course there's also the separate problem that we run into a random Ruby error rather than producing our own error telling the user what's wrong.)

What do you think of changing the API and having `close` take an argument indicating whether or not the entire connection is closed?  (Or, alternately, having another overloaded function you can define `on_connection_close` or `on_force_close` that will handle those cases?)

nviennot:
I don't think we want to give more information to `on_close()`, because if `on_close()` simply resumes a fiber, it's not going to help, because it won't know what the resumed fiber might do.

However, if you don't want to allow new queries, maybe we could raise in `em_run()` with a EM shutdown error or something.

mlucy:
> maybe we could raise in em_run() with a EM shutdown error or something.

Raising an error if you try to run a query while EventMachine is shutting down or the connection is closing sounds pretty reasonable.  Let's do that.  Closing this pull request, I opened https://github.com/rethinkdb/rethinkdb/issues/4640 to track the other change.

PR #4626, opened 2015-08-04
Signal error before closing EM handles abruptly

nviennot:
Otherwise, there is no way to distinguish an empty stream, and a connection that has closed prematurely from the point of view of the handler.

### DIFF
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -163,6 +163,12 @@ module RethinkDB
         handle(:on_close)
       end
     end
+    def handle_force_close
+      if !@closed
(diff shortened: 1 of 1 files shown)


nviennot:
My earlier statement is false. By tracking `on_open` we can, but still, it would be better I'd think to provide an error of some sort when handlers get closed abruptly.

Also, it's unclear from https://github.com/rethinkdb/rethinkdb/blob/next/drivers/ruby/lib/net.rb#L253  what the on_close() would mean without a corresponding on_open or on_error.

mlucy:
Thanks @nviennot !  I agree this is probably better behavior.  Merging.

Issue #117, opened 2012-11-28
Support multiple pkey autogen schemes


coffeemug:
The uuid key autogen scheme we use isn't enough. We should allow picking an autogen scheme at table creation. Here are some alternative schemes we should support:
1. Approximately chronological (e.g. non-colliding timestamps, roughly ordered)
2. Approximately ordered (e.g. incrementing from zero, roughly ordered)

thelinuxlich:
A good pkey autogen scheme: https://github.com/ericelliott/cuid

mbrevda:
Perhaps considering additional flexability would be usefull. For example, we don't like the hyphens in our UUID's. It would be nice if it was possible to remove these

coffeemug:
:+1:

sinfex:
Hopefully this proves to be of some value... I find it to be a pretty elegant solution: [Sharding & IDs at Instagram](http://instagram-engineering.tumblr.com/post/10853187575/sharding-ids-at-instagram)

xfg:
+1 for human-readable ids

RubenKelevra:
wondering why we don't have a simple increasing int, I know this might hurt performance to do a global lock on the database, but it would be useful for some application types.

So we would need to run something like this atomically:

`r.db('nodes').table('basicdata').insert({'id': r.db('nodes').table('basicdata').max('id').getField('id')+1 })`

thelinuxlich:
The problem with autoincrementing ids is that RethinkDB uses range sharding, so you would need to run rebalance() on the table often.

RubenKelevra:
@thelinuxlich are you sure this is really an issue? I've imported a table with an auto-incrementing id from a MySQL-Server and this is the data-distribution:

![bildschirmfoto von 2017-08-02 12-27-03](https://user-images.githubusercontent.com/614929/28869821-46979c1a-777e-11e7-8660-597f50a423ca.png)

RubenKelevra:
Is there currently ANY possibility to run the following SQL statement in an analog fashion in ReQL?

```
Lock tables tbl write;
$maxid = SELECT MAX( id ) FROM tbl;
INSERT INTO tbl ( id, a, b ) VALUES ($maxid+1, 'abc', 'def');
unlock tables;
```

thelinuxlich:
I'm sure it's a issue

RubenKelevra:
@thelinuxlich forgive me the noobish question, but why is my table in this case perfectly balanced?

[...]

PR #4636, opened 2015-08-06
Name-based UUID from RQL

captainpete:
### Motivation

In RethinkDB one of the modelling constraints is that primary keys are limited to single attributes.
In [DV](https://en.wikipedia.org/wiki/Data_Vault_Modeling) modelling a similar constraint is present, and it's common practice to generate surrogate keys with hashing algorithms using one or more natural keys.

Hashed keys are useful for knowing the id of a record (that may not exist yet) if we have the natural key values (useful for parallel/partial loading). Primary surrogate keys derived from hashed natural keys are also useful in the situation when replacing or updating documents matching multiple values (a kind of soft uniqueness constraint for atomic updates by key vector).

All this hashing can be handled client-side, but ETL style insert-from-select queries are a lot faster and will run in parallel. Having a deterministic hashing function (something like `r.sha1`) in RQL allows for some pretty cool relational modelling based on existing natural keys.

### The patch

Instead of introducing new hashing terms into RQL, existing UUID code allows for generation of name-based UUIDs (RFC4122 4.3) using SHA1. I've modified the `r.uuid` term allowing an optional value to be passed in. This also matches the existing default primary key convention of a UUID.

### Possible improvements

- I've used a global base UUID, allow specifying the base (for example, a table name).
- Don't overload `r.uuid`, call it something else that produces a UUID.
- `is_deterministic` should return true if we're using a name.

Haven't written anything in C++ for many years but the patch is pretty straight-forward.


thelinuxlich:
+1000 for this, my system uses sha256 primary keys everywhere because of this constraint. Hashing the id directly on RethinkDB would let me cut some microservices running only for this purpose.

thelinuxlich:
Also, why not SHA-256? :)

lucaspalencia:
+1

gustavopalencia:
+1

guerreiro:
+1

rksasaki:
+1

feliperoberto:
+1

simaodeveloper:
+1

deontologician:
One thing to keep in mind is that `generated_keys` in the response will not return these values, like it will for default key generation. So to get the keys for a document you'll have to tell it to return changes like `r.table('foo').insert(..., return_changes=True).run(...)`

thelinuxlich:
Totally okay with that

mlucy:
Thanks for the pull request @captainpete !

Tagging this as a ReQL proposal since it changes the API and putting it into 2.2 since so many people seem to want it.  Generally the way we handle API changes is we have a discussion period at the start of every release (to avoid distracting people in the middle of a release).

danielmewes:
Using the existing `uuid` format for this sounds neat.
Let's give this a bit of discussion for 2.2. I definitely like the basic idea.

renanvaz:
+1

coffeemug:
My only concern is how this relates to #117. For example, if we add `r.uuid(type='scheme1 | scheme2 | etc')`, passing a name to `r.uuid` won't make sense in all cases.

The options I can think of:

- Just merge the patch, and then add another term for other types of key generation later.
- Adjust the patch to make the string a named optarg.
- Merge the patch; when we add other schemes we can throw if the user passes a string.
- merge the patch; when we add other schemes, break the API and make the string a named optarg then.

@mlucy @danielmewes -- what do you think?

danielmewes:
I think it might be better to add new terms for schemes that are very different.
"uuid" is usually associated with a specific set standards for ID generation (https://en.wikipedia.org/wiki/Universally_unique_identifier#Standards), so I think things like the 'ordered' UUID that was proposed in https://github.com/rethinkdb/rethinkdb/issues/117#issuecomment-55217480 might not be a great fit for the `r.uuid` term.

There is another part to #117 which is supporting a more compact representation of UUIDs (i.e. not as a human-readable string, but something shorter). That part would still work well with this patch, assuming the representation will be changed through an optarg.

It's a bit hard to say given that #117 is still in flux, but as things are now, I have a slight preference for keeping the patch as it is. It feels nicer to write `r.uuid(name)` than `r.uuid({name: name})`.

thelinuxlich:
r.uuid(string, optional_optarg_with_scheme) would be nice

mlucy:
@coffeemug -- I think we should merge it, in the worst case where we decide that we want `r.uuid` to support a generation scheme that doesn't make sense when passed a constant we can just error in that case.

marshall007:
This is pretty minor, but could we have the server automatically coerce the `name` argument to a string? The motivation for this PR implies that `r.uuid(<array>)` and `r.uuid(<object>)` are probably the most useful cases, so not having to chain that extra `.coerceTo('string')` on the argument would be great!

mlucy:
If it's just data that we're using as a seed, I don't see any reason why we can't accept arbitrary data as a seed instead of just strings.  We should try to have the property that if two values compare unequal in ReQL they count as different seeds.  (That might be too complicated for an initial implementation though.)

deontologician:
With using arbitrary objects as the seed, we probably need some predetermined and unchanging way to generate the final string that gets hashed to a uuid. So we'd need to provide a contract along the lines of "The json serialized version of the object, with keys alphabetized and pseudotypes serialized like so...".

Alternately, we could define it as however they're serialized now when we order things lexicographically for indexes.

mlucy:
We currently don't allow objects as keys, but I don't think it would be too hard to define a deterministic object -> string transformation (the only worry would be that we could never change the field ordering even if RethinkDB later becomes more Unicode-aware).

deontologician:
Yeah, it's the "whatever we decide now is locked in forever" aspect that is troubling. I'd rather not throw it in as an add-on even though it would be super convenient. For now users can do something like `r.uuid(blah.coerceTo('string'))` if they need this functionality

marshall007:
Is `<object>.coerceTo('string')` not already deterministic? It appears to already sort keys lexicographically before serializing to JSON, at least.

deontologician:
It is, but I don't believe there's any guarantee it will never change. If we're saying "generate your primary keys with this" we're making a commitment that if you query with `table.get(r.uuid(my_real_query_object))` on your data in the future, it will not suddenly stop working because we changed what fields we add to pseudo-types or something

danielmewes:
@marshall007 It is, but we have changed our JSON encoder in the past. So this limits our options for changing those details in the future.

Another problem that's specific to allowing objects is that it could clash with later adding optargs to `r.uuid`.

coffeemug:
Ok, I remove my objection.

thelinuxlich:
As long as I can replace all those "crypto.createHash()/hash.update(key)/hash.digest('hex')" in my code, manually coercing is nothing :)

mlucy:
Alright, since there are complications let's just support it for strings in the first version.

danielmewes:
Marking as settled from the API perspective, as implemented in the PR (i.e. `r.uuid` accepts an optional string argument).

coffeemug:
@danielmewes -- do we still plan to merge this PR for 2.2? Is there a reason we haven't merged it yet (other than test time/scheduling)?

danielmewes:
Sorry for the delay. The patch looks good, merging this now.
It will ship with RethinkDB 2.2.0.

Issue #4559, opened 2015-07-22
Implement changed error hierarchy from #4544


danielmewes:
See https://github.com/rethinkdb/rethinkdb/issues/4544#issuecomment-123514164

mlucy:
@danielmewes -- this isn't assigned to anyone.  Are we going to have the person in charge of each driver do this, or should one person just sit down and do all of them do you think?

danielmewes:
@mlucy I don't really care, though I think it will be easier if one person updates all the drivers at once. 
Since all of your 2.1 issues are now in code review or waiting for test fixes (or for this), would you mind picking this one?

larkost:
I just checked in some changes to the testing system to make the polyglot test system actually work for this type of test, and it looks like there is some issue in how the JavaScript driver is picking up error types. For instance this is coming up in the `polyglot/changefeeds/edge.js` a few times where one of the new subclasses is not being recorded but rather `RqlRuntimeError` being returned.

mlucy:
@larkost -- I broke that out into a separate issue so it won't get lost.

deontologician:
Should the TimeoutError be in new drivers? The comment next to it seems to indicate it should only be there for backwards compatibility

danielmewes:
@deontologician I think it's what the Python driver throws if waiting on a cursor times out (http://rethinkdb.com/api/python/next/). If there's a native error type for timeouts in the language that can be used instead. Are you asking because of the Java driver?

deontologician:
yeah, I added some timeout code yesterday, and didn't create a ReqlTimeoutError based on the note next to it, I just wanted to clarify that was the intention

deontologician:
OK, new query: what happens to the CLIENT_ERROR responsetype? Is it renamed something else? The previous thread didn't mention it

danielmewes:
Good question. It should probably just be another error type `ReqlClientError` under `ReqlError`.

mlucy:
I made the appropriate changes to the protobuf file, the Ruby driver and the tests in the `mlucy_4559` branch.  @deontologician, @grey -- could I talk you into doing the changes for the JS driver and the Python driver respectively?  I can do it if you guys don't have time before the release, but there were a few small subtleties in the Ruby driver changes (where to detect `AuthError`, how to do the aliasing for backward-compatibility) that I don't quite trust myself to do properly in the drivers I don't understand as well.

mlucy:
Er, I mean @Tryneus .

[...]

Issue #4646, opened 2015-08-08
JavaScript expected arguments errors should be RqlCompileErrors


larkost:
As exemplified in the `polyglot/arity` test the JavaScript driver is reporting a number of errors as `RqlDriverError`s when the other drivers report them as `RqlCompileError`s. The reasons for the current state are internal to how the drivers work, and so not ones that would matter for customers. I would rather we have consistency on this.

Probably every `RqlDriverError` line in `polyglot/arity` should change after this.

deontologician:
Since I'm in the code now and we're changing the errors etc, @danielmewes any opposition to making this change for 2.1?

danielmewes:
It's not necessary, but if it's a quick fix and you make sure that all tests are updated go for it. We will freeze 2.1 once the Python driver error changes are in today though.

deontologician:
OK, I did this the easy way and it's not quite as simple as I'd expected. In order to unify the error specification in the polyglot tests, we'd also have to change how many arguments are reported by the error message. Right now there are numerous that say "expected 0 but found 1", whereas the other drivers say "expected 1 but found 2". I think that's a large enough change not to try to slip it in at the last moment, so I'm not doing this for now.

deontologician:
@larkost if you pick this up later and are interested, the little bit I did is on josh/4646

danielmewes:
Seems like there are also a few cases like this in the Ruby driver: https://github.com/rethinkdb/rethinkdb/pull/4652/files#r3668721

danielmewes:
Was fixed as part of 4669.

PR #4652, opened 2015-08-10
Ruby + test changes to error hierarchy

danielmewes:
To speed things up a little, I went ahead and created this pull request from @mlucy's branch `mlucy_4559`, see https://github.com/rethinkdb/rethinkdb/issues/4559 .

I'm going to do a pass over this.


test/rql_test/src/times/constructors.yaml
```diff
@@ -4,31 +4,31 @@ tests:
 # Note that JS Date comparison doesn't work, these are implemented in the connection tests
   - cd: r.expr(r.epoch_time(896571240))
     ot:
-      py: datetime.fromtimestamp(896571240, r.ast.RqlTzinfo('00:00'))
```
danielmewes: These shouldn't have been renamed as part of this.
We can still merge this though, since @VeXocide reverted these test changes in his Python driver change.

danielmewes:
The test and server changes look fine.

@deontologician could you check the Ruby driver changes in this please?

drivers/ruby/lib/func.rb
```diff
@@ -151,7 +151,7 @@ def ==(rhs)
     def do(*args, &b)
       a = ((@body != RQL) ? [self] : []) + args.dup
       if a == [] && !b
-        raise RqlDriverError, "Expected 1 or more arguments but found 0."
+        raise ReqlDriverError, "Expected 1 or more arguments but found 0."
```
deontologician: Out of curiosity, aren't arity things like this compile errors?

danielmewes: This is the same problem as in the JS driver https://github.com/rethinkdb/rethinkdb/issues/4646
Changing this is a separate issue.

deontologician:
I don't see any issues in the Ruby driver

deontologician:
:+1:

danielmewes:
Merging! :)

PR #4726, opened 2015-08-19
Possible missing break

lenstr:


danielmewes:
I think this is the first pull request for the Java driver. :-)
Pinging @deontologician

deontologician:
! nice catch. Obviously, lots of work still being done, but this wasn't on my radar to revisit. Thanks @lenstr

PR #4757, opened 2015-08-25
The correct spelling: occurred

alexsalas:


danielmewes:
Thanks for the PR @alexsalas . Would you mind taking out the changes that affect the `external` path (i.e. the v8 changes)?

That code comes from Google. I'd recommend submitting that patch directly to https://code.google.com/p/v8/ .

Have you signed our CLA yet? http://rethinkdb.com/community/cla/

alexsalas:
Already signed the CLA :+1: @danielmewes

danielmewes:
Thanks @alexsalas. So the only remaining thing is taking out the V8 patches, and I think then we can merge.

danielmewes:
@alexsalas Do you want to update the PR to take out the V8 files or would you prefer us to pick it up from here?

larkost:
Since this pull request has gone stale I decided to re-do it, and it is now checked into `next` with 3c988775. @alexsalas thanks again for finding the misspelling!

PR #4783, opened 2015-09-01
This is my attempt at documenting how to get started hacking on the driver

sontek:


drivers/python/README.md
```diff
@@ -1,3 +1,40 @@
...
+
+```bash
+git clone https://github.com/rethinkdb/rethinkdb.git
+./configure --allow-fetch
+cd drivers/python
+make
+pip install -e .
+```
+
+The `make` command will take awhile, if you have a lot of cores you can speed
```
danielmewes: I think running `make` in `drivers/python` should be very fast, so this paragraph is probably unnecessary.

danielmewes: That being said, the tests will require the full server. So I guess there should be an extra step about running `make` in the rethinkdb root directory?

test/rql_test/test-runner
```diff
@@ -30,7 +30,7 @@ try:
     openFileLimit = int(os.getenv('OPEN_FILE_LIMIT', 512))
     if resource.getrlimit(resource.RLIMIT_NOFILE)[0] < openFileLimit:
         resource.setrlimit(resource.RLIMIT_NOFILE, (openFileLimit, openFileLimit))
-except Exception, e:
+except Exception as e:
```
danielmewes: This fixed a problem with certain Python versions right?

danielmewes:
Thanks for the PR @sontek !

@larkost What's your opinion on putting the `tox.ini` into the repository?

danielmewes:
@sontek If you haven't done this yet, would you mind taking a look at our CLA http://rethinkdb.com/community/cla/ and signing it online? We need this for legal reasons before we can merge a PR.

test/rql_test/test-runner
deontologician: this syntax is accepted in Python 3 and 2.6+, whereas the comma is only accepted in 2.x

drivers/python/README.md
sontek: running `make` in the python directory does the full `make` if necessary which is why its takes a long time the first time

test/rql_test/test-runner
sontek: This was in preparation for `tox -e py34` but this file isn't quite py34 compatible but I'll send a separate PR for that.

drivers/python/README.md
danielmewes: That's weird, it doesn't for me. Are you sure you didn't accidentally run it in `drivers` rather than `drivers/python`?

sontek: This is a valid concern that I'm not sure of ;)   I'll re-run the instructions from scratch to verify when I get home tonight

drivers/python/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+PyOpenSSL
+PyYAML
```
larkost: The `PyYAML` requirement is only on the test system, and there are plans to remove it (#4787)

drivers/python/tox.ini
```diff
@@ -0,0 +1,19 @@
...
+pip_pre=False
+commands =
+    {[base]commands}
+    {toxinidir}/../../test/rql_test/test-runner -i py2.7 {posargs}
+
+[testenv:py34]
+pip_pre=False
+commands =
+    {[base]commands}
+    {toxinidir}/../../test/rql_test/test-runner -i py3 {posargs}
```
larkost: This should be `-i py3.4` if you are specifying Python3.4. Otherwise you will get Python3.0 if it is installed. With `twisted` not being supported on Python3.0 (grumble, grumble), that would defeat some of the purpose.

Also, why not include versions for 2.6, 3.0, 3.1, 3.2, and 3.3? That way anyone who wants to can use if for any version.

drivers/python/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
...
+twisted
+tornado
```
larkost: Why not just include these as [`deps`](https://testrun.org/tox/latest/config.html#confval-deps=MULTI-LINE-LIST)? That would even allow for not including things like `twisted` where they are not compatible.

sontek:
@danielmewes @larkost I made some changes that addressed most comments.

sontek:
@danielmewes I left the information about using `make -j8` because of having to run make on the whole database.

larkost:
I am with @danielmewes about `README.md`: you do not need to `make` the whole project in order to build the driver (`configure` is a different story). I just tested that to make sure.

However, you do need to have a RethinkDB server executable to do any testing, and we default to using the latest built version. However, setting a `RDB_EXE_PATH` environmental variable should make the testing system use an existing install, but I have not audited everything to make sure that there are not holdouts that would ignore that.

larkost:
The `PyYAML` requirement is now gone. We no longer use it at all with the merging of #4787.

sontek:
@larkost Running a full make seems to be required at this point but I'd be fine with updating the docs to use RDB_EXE_PATH when the tests actually allow it:

```
(rethinkdb_env)Johns-MBP:python sontek$ export RDB_EXE_PATH=`which rethinkdb`
(rethinkdb_env)Johns-MBP:python sontek$ tox -e py27
py27 installed: backports-abc==0.4,backports.ssl-match-hostname==3.4.0.2,certifi==2015.9.6.2,cffi==1.3.0,cryptography==1.1,enum34==1.0.4,idna==2.0,ipaddress==1.0.15,pyasn1==0.1.9,pycparser==2.14,pyOpenSSL==0.15.1,rethinkdb==2.2.0.post0,singledispatch==3.4.0.3,six==1.10.0,tornado==4.3,Twisted==15.4.0,wheel==0.24.0,zope.interface==4.1.3
py27 runtests: PYTHONHASHSEED='1344471790'
py27 runtests: commands[0] | pip install -e .
Obtaining file:///Users/sontek/code/rethink/rethinkdb/drivers/python
Installing collected packages: rethinkdb
  Running setup.py develop for rethinkdb
Successfully installed rethinkdb-2.2.0.post0
py27 runtests: commands[1] | /Users/sontek/code/rethink/rethinkdb/drivers/python/../../test/rql_test/test-runner -i py2.7
An item was not built: no version of this project has yet been built
ERROR: InvocationError: '/Users/sontek/code/rethink/rethinkdb/drivers/python/../../test/rql_test/test-runner -i py2.7'
_________________________________________________________________________ summary __________________________________________________________________________
ERROR:   py27: commands failed
```

sontek:
I removed the requirement for PyYAML now but I do believe the rest of the instructions are necessary since the tests wont run unless you have a `build` directory.

sontek:
@larkost @danielmewes Does this look good?

drivers/python/tox.ini
```diff
@@ -0,0 +1,26 @@
+[tox]
+skipsdist = True
+envlist = py27, py34
+
+[base]
+deps =
+    PyOpenSSL
```
larkost: `PyOpenSSL` is not a requirement, rather`twisted.internet.ssl` has a requirement on the `OpenSSL` module. So from our standpoint `twisted` is a requirement for whatever `/usr/bin/env python` comes back with. Additionally it is required to test the `twisted` flavor of the driver, but that does not work on Python3.0 through about Python3.2 (except on some versions of 3.2... shrug).

So `PyOpenSSL` should be removed, and `twisted` added (at least for non-3.0-3.2).

drivers/python/README.md
```diff
@@ -1,3 +1,46 @@
...
+git clone https://github.com/rethinkdb/rethinkdb.git
+cd rethinkdb
+./configure --allow-fetch
+make
+cd drivers/python
+make
+pip install -e .
+```
+
+Since the python driver requires you to run the `make` command for the whole
```
danielmewes: I suggest changing this to "Since *testing* the Python driver..."

danielmewes:
@sontek thanks for updating the PR.
I added one more comment, and there's the dependency detail that @larkost mentioned.
Everything else looks good to me.

(also please let us know if you'd prefer us to take this over and fix the remaining minor issues)

drivers/python/tox.ini
```diff
@@ -0,0 +1,26 @@
+[tox]
+skipsdist = True
```
larkost: I don't think that `skipdist` should be set. `sdist` is very cheap for us, and with some recent changes that would make it even better to test with `sdist`-trimmed drivers.

drivers/python/tox.ini
```diff
@@ -0,0 +1,26 @@
...
+envlist = py27, py34
```
larkost: This list should include everything from Python2.6 through Python3.5. Adding `skip_missing_interpreters = True` would keep if from being a problem.

drivers/python/README.md
```diff
@@ -1,3 +1,46 @@
...
+`8` meaning use 8 cores to compile this.
+
+After you have everything installed you can test it by running this command:
+
+```bash
+python -c 'import rethinkdb'
+```
+
+*Note*: The make for `rethinkdb` itself (not the python driver) requires python2
+so make sure you do not have a python3 virtualenv activated when running it.
```
larkost: I have not been able to find anything that would break here, and am only aware of one place where we use Python in the build system, and that seems to work fine for Python3.x. If you can tell me where this is, I will fix it.

drivers/python/README.md
larkost: And for completeness, you can run testing using an existing install using the `--server-binary` option. Not sure how that is going to play here.

sontek:
Its been a few years, going to close this

PR #4784, opened 2015-09-02
JS Driver: promise friendly `cursor.eachAsync()`

### LINKED ISSUE #4780: (Javascript) Make `each()` return a Promise
Each iteration of `each()` is done asynchronously. In the following code, close would be called before the second iteration of each:

```javascript
cursor.each(async (err, data) => {
  if(err) throw err;
  await dosomethingasynchronouslywithdata();
});
await conn.close();
```

Almost all methods offer the choice between a callback and a Promise. `each` is an exception: it only offers an `onFinished` callback, but it doesn't return a `Promise`. If this were done, the code would look like:

```javascript
await cursor.each(async (err, data) => {
  if(err) throw err;
  await dosomethingasynchronouslywithdata();
});
await conn.close();
```

marshall007:
Fixes #4780. Unfortunately, there was no way to implement this as an overload to the existing `cursor.each()` since there's no way to disambiguate the argument style.

Here's some example usage:

```js
cursor.eachAsync(function (row) {
  if (/* some condition */) {
    // if you return a promise, it will be awaited
    // before the cursor continues iteration
    return asyncRowHandler(row.id);
  }
  else {
    // iteration can be stopped early by returning a
    // promise that is later rejected or simply by throw'ing
    return Promise.reject('Stop iteration early.')
  }
})
.then(function () {
  // `cursor.eachAsync()` returns a promise that is resolved
  // once all rows have been read from the cursor
});
```

### DIFF
--- a/drivers/javascript/cursor.coffee
+++ b/drivers/javascript/cursor.coffee
@@ -240,30 +240,11 @@ class IterableResult
         return nextCb()
 
     toArray: varar 0, 1, (cb) ->
-        fn = (cb) =>
-            arr = []
-            eachCb = (err, row) =>
-                if err?
-                    cb err
-                else
-                    arr.push(row)
-
-            onFinish = (err, ar) =>
-                cb null, arr
-
-            @each eachCb, onFinish
-
         if cb? and typeof cb isnt 'function'
             throw new err.ReqlDriverError "First argument to `toArray` must be a function or undefined."
 
-        new Promise( (resolve, reject) =>
-            toArrayCb = (err, result) ->
-                if err?
-                    reject(err)
-                else
-                    resolve(result)
-            fn(toArrayCb)
-        ).nodeify cb
+        results = []
+        return @eachAsync(results.push.bind(results)).return(results).nodeify(cb)
 
     _makeEmitter: ->
         @emitter = new EventEmitter
--- a/drivers/javascript/cursor.coffee
+++ b/drivers/javascript/cursor.coffee
@@ -217,15 +217,12 @@ class IterableResult
         if onFinished? and typeof onFinished isnt 'function'
             throw new err.ReqlDriverError "Optional second argument to each must be a function."
 
-        stopFlag = false
         self = @
         nextCb = (err, data) =>
-            if stopFlag isnt true
-                if err?.message is 'No more rows in the cursor.'
-                    onFinished?()
-                else
-                    stopFlag = cb(err, data) is false
-                    @_next nextCb
+            if err?.message is 'No more rows in the cursor.'
+                onFinished?()
+            else if cb(err, data) isnt false
+                @_next nextCb
             else
                 onFinished?()
         @_next nextCb
--- a/drivers/javascript/cursor.coffee
+++ b/drivers/javascript/cursor.coffee
@@ -39,6 +39,7 @@ class IterableResult
 
         @next = @_next
         @each = @_each
+        @eachAsync = @_eachAsync
 
     _addResponse: (response) ->
         if response.t is @_type or response.t is protoResponseType.SUCCESS_SEQUENCE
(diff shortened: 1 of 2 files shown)


jorrit:
If I read the coffeescript correctly this method will sequentialize all iterations by invoking the next callback when the previous one is resolved. Isn't it possible to execute all iterations in parallel with a `Promise.all`-like approach?

marshall007:
@jorrit this would be easy to do now, at least:

```js
var results = [];

cursor.eachAsync(function (row) {
  results.push(asyncRowHandler(row.id));
  // as long as we don't return the promise
  // the next row will be read immediately
})
.then(function () {
  return Promise.all(results).then(function (data) { /* ... */ });
});
```

jorrit:
Personally I see no benefit in eachAsync supporting a callback that returns Promises when it serializes these promises. Maybe make it configurable with an options parameter?

deontologician:
はい

larkost:
I would make me (the testing guy) particularly happy if you added appropriate tests to `test/rql_test/connections/cursor.mocha`

drivers/javascript/cursor.coffee
```diff
@@ -160,23 +161,14 @@ class IterableResult
...
-                cb = (err, result) ->
-                    if (err)
-                        reject(err)
-                    else
-                        resolve(result)
-                fn(cb)
-            return p
-        else
-            throw new err.ReqlDriverError "First argument to `next` must be a function or undefined."
+        return Promise.fromNode(fn).nodeify(cb)
```
deontologician: Damn, that is a nice cleanup. I need to do that in the other places we do this manually...

marshall007:
@larkost I already copied the `cursor.each` tests and made them work for `cursor.eachAsync`, but I'll definitely add a couple more since those are fairly sparse.

deontologician:
>  Maybe make it configurable with an options parameter?

I like this interface. It seems to make the common case easy. When I think of iteration with each, I think of them executing in order. Given that you can just collect them and run them asynchronously if you want, it seems ok not to provide an option.

It looks good to me overall. I've played with the code a bit and it works like I'd expect. Since this is an api change, I'd like to let it breath for a bit before merging it. Maybe get some opinions from other js guys that hang around here like @neumino and @joaojeronimo ...

drivers/javascript/cursor.coffee
marshall007: @deontologician thanks! In my last commit I went ahead and refactored `cursor.toArray` into a one-liner as well.

marshall007:
@larkost just added tests covering all the cases outlined in my opening comment and added a "stop iteration early" test to `cursor.each` as well.

@deontologician somewhat unrelated to this PR, the original `cursor.each` implementation appears to read an extra row off the cursor after iteration has already been stopped (but doesn't fire the callback). I added a fix for that and the tests still pass, but I wanted to make sure I wasn't overlooking something intentional.

deontologician:
Looks like it was part of removing `hasNext()` in commit 190c8c262a8be937fa8a4bac1148c58e68ac46d3 . It's possible reading one more is for ensuring the cursor has retrieved everything there is off the server, but I haven't looked to closely yet at the code.

marshall007:
Ok, that makes me pretty certain it was a bug. We weren't even checking the result of that last (unintentional) iteration, so I don't see how it could impact anything.

neumino:
@deontologician -- Adding a new method seems reasonable to me and I can see how useful it can be.

Maybe there's a better name than `eachAsync` since people may thing that `each` is synchronous, but I  have to admit that I can't come up with a better name.
Bluebird's doc calls `map(... {concurrency: 1})` a static map, but it doesn't sound good :)

deontologician:
Ok, I'm ready to merge this, but I was wondering what people thought of the name "eachPromise" for the name. I think I agree with @neumino that since both each implementations are async, it might be misleading to call it eachAsync. @marshall007 thoughts? (or a better name)

marshall007:
@deontologician I think at the next opportunity to introduce breaking changes (i.e. 2.2) we should consider changing the current signature of `cursor.each` so that it can once again support both styles:

```js
cursor.each(function(err, row), function()) // current

cursor.each(function(row, cb), function(err)) // new callback-style
cursor.each(function(row)) // new promise-style
```

I wouldn't mind waiting to merge all this at once under `cursor.each`. If you want to go ahead and get it released, I think the `-Async` suffix is fairly standard and I definitely prefer that over `eachPromise`.

https://github.com/petkaantonov/bluebird/blob/master/API.md#promisification

Notice examples like `fs.readFileAsync` in their docs even though `fs.readFile` is asynchronous as well.

deontologician:
Hmm. Alright, in that case I think we should just merge this as-is then. Breaking backwards compatibility in something this core is something I would expect in a major release (like 3.0). I think adding the extra method is a good compromise in the meantime.

Issue #544, opened 2013-03-28
have server log client requests


taybin:
It would be good if there was a more verbose logging option. Right now, I haven't seen rethinkdb print out anything after starting up.

flockonus:
+1 fww - my motivation to use this feature is to debug development code. The JS library doesn't seem to have a way to log queries.

Tryneus:
Now that we have the `rethinkdb.jobs` table, we can add that information to it with #3636.  Then, if we make changefeeds on the `jobs` table not poll (and potentially miss short-running queries), it would be trivial to implement a query log in an arbitrary client using:

```py
r.db('rethinkdb').table('jobs').filter(r.row['id'][0].eq('query')).changes(squash=False)
```

We could even integrate it into the Web UI or data explorer.  Of course, we might want to filter out the web UI's requests to avoid spam.

danielmewes:
Yeah that would be pretty cool. We'll also need a query pretty printer, probably based on https://github.com/rethinkdb/rethinkdb/issues/3636

danielmewes:
I'd like to consider this for 2.2, since it would be very useful in combination with access permissions (https://github.com/rethinkdb/rethinkdb/issues/4519 ) for monitoring / auditing purposes.

coffeemug:
FYI, talked to a user today in a regulated industry; audit trails require logging actual data changes, along with the user information.

marshall007:
Is it currently possible to filter the output of `r.db('rethinkdb').table('jobs').changes()` to only include jobs from the current client or do we have to wait on something like #4519?

The `info.client_port` property appears to be unique across clients, but is it possible to obtain that information from the drivers?

coffeemug:
Unfortunately I can't think of a way of doing that right now (AFAIK the server doesn't send the `client_port` back to the drivers). We should definitely make this information available in the drivers, it would make the property in the `jobs` table dramatically more useful.

danielmewes:
I opened https://github.com/rethinkdb/rethinkdb/issues/4620 for the `client_port` suggestion.

Omeryl:
This would be really, really helpful :).

shinzui:
I am curious how to debug bad clients without any logging. I tried using a changefeed filtered to my database but that was not very helpful. This issues is three years old so there must be a workaround that I am missing.

mxstbr:
Has anybody found a workaround for this yet? It's been a couple of years now... :wink:

[...]

Issue #4084, opened 2015-04-20
Setup SSL tests for client drivers


larkost:
SSL support is being added to the drivers for use at Compose.io (but not the server at this point). We need to add tests to make sure that we don't introduce regressions in that behavior. The easiest way of doing this is probably to expand on the SSL support in `test/rql_test/connections/http_support/server.py` and make a simple proxy to a server that we spin up for the purpose.

An example of a socket proxy can be found at: http://voorloopnul.com/blog/a-python-proxy-in-less-than-100-lines-of-code/

danielmewes:
Turns out we actually had a bug in the JS driver that we didn't notice. I would feel better with this test in place.
Scheduling for 2.2

PR #4796, opened 2015-09-04
Drivers: expose client port and address

### LINKED ISSUE #4620: Add a function to driver to get outgoing port
See the comment here https://github.com/rethinkdb/rethinkdb/issues/544#issuecomment-127435975

It might be useful to add a method to the connection objects that returns the outgoing IP port. This is probably a pretty simple client-side addition.

marshall007:
Fixes #4620. The original use-case was to filter the `jobs` table to only include jobs initiated by the current client ([see #544 comment](https://github.com/rethinkdb/rethinkdb/issues/544#issuecomment-127435975)).

I will also update the Ruby driver if the API looks ok to everyone.

### DIFF
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -524,10 +524,10 @@ class Connection(object):
         self._next_token = 0
 
     def client_port(self):
-        self._instance._socket.client_port()
+        return self._instance._socket.client_port()
 
     def client_address(self):
-        self._instance._socket.client_address()
+        return self._instance._socket.client_address()
 
     def reconnect(self, noreply_wait=True, timeout=None):
         if timeout is None:
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -317,11 +317,19 @@ class SocketWrapper(object):
                 raise ReqlDriverError("Server dropped connection with message: \"%s\"" %
                                       (message, ))
 
+    def client_port(self):
+        if self.is_open():
+            return self._socket.getsockname()[1]
+
+    def client_address(self):
+        if self.is_open():
+            return self._socket.getsockname()[0]
+
     def is_open(self):
         return self._socket is not None
 
     def close(self):
-        if self._socket is not None:
+        if self.is_open():
             try:
                 self._socket.shutdown(socket.SHUT_RDWR)
                 self._socket.close()
@@ -514,6 +522,12 @@ class Connection(object):
         self._child_kwargs = kwargs
         self._instance = None
         self._next_token = 0
+
+    def client_port(self):
+        self._instance._socket.client_port()
+
+    def client_address(self):
+        self._instance._socket.client_address()
 
     def reconnect(self, noreply_wait=True, timeout=None):
         if timeout is None:
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -317,14 +317,6 @@ class SocketWrapper(object):
                 raise ReqlDriverError("Server dropped connection with message: \"%s\"" %
                                       (message, ))
 
-    def client_port(self):
-        if self.is_open():
-            return self._socket.getsockname()[1]
-
-    def client_address(self):
-        if self.is_open():
-            return self._socket.getsockname()[0]
-
     def is_open(self):
         return self._socket is not None
 
@@ -418,6 +410,14 @@ class ConnectionInstance(object):
         self._header_in_progress = None
         self._socket = None
         self._closing = False
+
+    def client_port(self):
+        if self.is_open():
+            return self._socket._socket.getsockname()[1]
+
+    def client_address(self):
+        if self.is_open():
+            return self._socket._socket.getsockname()[0]
 
     def connect(self, timeout):
         self._socket = SocketWrapper(self, timeout)
@@ -524,10 +524,12 @@ class Connection(object):
         self._next_token = 0
 
     def client_port(self):
-        return self._instance._socket.client_port()
+        if self.is_open():
+            return self._instance.client_port()
 
     def client_address(self):
-        return self._instance._socket.client_address()
+        if self.is_open():
+            return self._instance.client_address()
 
     def reconnect(self, noreply_wait=True, timeout=None):
         if timeout is None:
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -482,6 +482,13 @@ module RethinkDB
     end
     attr_reader :host, :port, :default_db, :conn_id
 
+    def client_port
+      is_open() ? @socket.addr[1] : nil
+    end
+    def client_address
+      is_open() ? @socket.addr[0] : nil
+    end
+
     def new_token
       @token_cnt_mutex.synchronize{@token_cnt += 1}
     end
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -486,7 +486,7 @@ module RethinkDB
       is_open() ? @socket.addr[1] : nil
     end
     def client_address
-      is_open() ? @socket.addr[0] : nil
+      is_open() ? @socket.addr[3] : nil
     end
 
     def new_token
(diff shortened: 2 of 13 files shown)


danielmewes:
Thanks @marshall007!

I think the API looks good.

@deontologician and @Tryneus could you review the JS and Python implementations respectively when you have a chance please?

One thing we should test is if these methods still work with SSL connections. They wouldn't be very useful at the moment, since whatever acts as an SSL proxy between the clients and the server is going to have a different port for its own outgoing connection to the server (this will change when we integrate SSL directly into the server). However we should make sure that they still return something reasonable and don't throw an error.
Until we have https://github.com/rethinkdb/rethinkdb/issues/4084, this might be a bit of a hassle to set up. I could probably launch a test instance on Compose and test it against that...

deontologician:
JS looks good :+1:

drivers/python/rethinkdb/net.py
```diff
@@ -515,6 +523,12 @@ def __init__(self, conn_type, host, port, db, auth_key, timeout, ssl, **kwargs):
         self._instance = None
         self._next_token = 0
 
+    def client_port(self):
+        return self._instance._socket.client_port()
```
larkost: I think it would be slightly better to have this be a properly. We can get a read-only property by just putting @property on the line before the `def`. `client_address` should probably get this as well. I am ambivalent about those on `SocketWrapper` since that is not deliberately exposed.

Additionally I am slightly concerned about what happens if `self._instance` or `self._instance._socket` are `None` when this is called (say when a `Connection` is `closed()`).

marshall007: > I think it would be slightly better to have this be a property.

I was thinking this too, actually. @deontologician should we make these read-only properties in the JS driver as well?

Tryneus:
Looks good, although it won't work for the Python async backends (asyncio, Tornado, and Twisted).

In default `net.py` `ConnectionInstance` has a member `_socket`, corresponding to a `SocketWrapper`, but the behaviors are different for the other backends:

* `net_asyncio.py`: `ConnectionInstance` has a member `_streamwriter`, and you can get the underlying socket via `_streamwriter.get_extra_info('socket')`
* `net_twisted.py`: `ConnectionInstance` has a member `_connection`, which is a `DatabaseProtocol` object, and you can get the local connection info via `_connection.transport.getHost()`, I believe.
* `net_tornado.py`: `ConnectionInstance` has a member `_socket`, of the type `socket.socket`.

Because these behaviors are so different on each `ConnectionInstance` implementation, the `ConnectionInstance` itself should provide methods for reading these properties.

@marshall007, I understand if you don't want to spend the resources to implement and test this on the other Python backends, and if you prefer I can open an issue and add the other implementations myself once this PR has been merged.

drivers/python/rethinkdb/net.py
deontologician: Eh... we don't use properties anywhere currently in the js driver, and I don't think this is important enough to warrant it. Generally I think properties are useful if you need to be backwards compatible with something that was previously an attribute. I think it being a function works fine in js. (Incidentally, I think making it a read-only property is overkill in python too, given python's "we're all adults here" ethos)

marshall007:
@Tryneus thanks, that seems pretty straightforward. I'll update this PR to include as much of that as I can and let you know if I don't get around to some of it.

Also, can someone point me to the Ruby connection tests if they exist? I don't see any under `rql_test/connections`.

danielmewes:
There currently isn't a direct Ruby equivalent of the connection tests for the other drivers. There are however a few tests for the EventMachine interface in test/rql_test` (such as `em.rb`). You could add a similar test for your client port function.

marshall007:
@Tryneus wasn't too bad after all. I added support for this to the remaining connection types with tests for each. It's ready for another look whenever you get the chance.

drivers/python/rethinkdb/net.py
marshall007: @larkost latest commit addresses the possible null reference you pointed out. All the tests now ensure that these two methods return `None` when called on a closed connection.

Tryneus:
Awesome, looks good, @marshall007!

marshall007:
@danielmewes I added support for the Ruby driver with a test under `connections/client_info.rb`. Is that pretty much what you were saying I should do?

Aside from setting up explicit tests over SSL connections, I think that wraps it up.

danielmewes:
Nice, I'll try to get the SSL testing done today or tomorrow.

danielmewes:
@marshall007 the test is about what I had in mind. Your Ruby changes don't seem to be in this PR yet btw.

danielmewes:
Oh wait, never mind. They're there. I must have been looking at an old version of the page.

marshall007:
@danielmewes haha, I just saw the same thing and was also confused.

drivers/ruby/lib/net.rb
```diff
@@ -482,6 +482,13 @@ def initialize(opts={})
     end
     attr_reader :host, :port, :default_db, :conn_id
 
+    def client_port
+      is_open() ? @socket.addr[1] : nil
+    end
+    def client_address
+      is_open() ? @socket.addr[0] : nil
```
danielmewes: This returns "AF_INET" or "AF_INET6" for me, not the actual address.

danielmewes:
Just tested all drivers with an SSL connection, and the `client_port`/`client_address` methods all work fine (except for the general problem with `client_address` in the Ruby implementation).

drivers/ruby/lib/net.rb
marshall007: Good catch, I misread the docs. I think it should be `@socket.addr[2]`.

marshall007:
@danielmewes by the way, is it possible that the hostname we get back from all these different socket implementations could differ from the `client_address` in the `jobs` table? For example, does the server resolve hostnames to a numeric address or anything like that?

danielmewes:
@marshall007 The server will always show numeric addresses I believe. It looks like the driver implementations (at least JS and Python) do the same, don't they?

However any NAT router or other higher-level proxy will of course cause mismatches in those addresses.
I'm not sure, but it's also possible that IPv4 addresses will be represented by an IPv6 address in some contexts.

That's one of the reasons for why an explicit `client_id` for the `jobs` table would be a lot nicer medium term...

marshall007:
@danielmewes I think in Node and Ruby we'll always get a numeric address back. The [Python docs](https://docs.python.org/2/library/socket.html), however, seem to imply we could get back a hostname or address:

> A pair (host, port) is used for the `AF_INET` address family, where host is a string representing either a hostname in Internet domain notation like `daring.cwi.nl` or an IPv4 address like `100.50.200.5`, and port is an integer.

I don't think we need to worry about this right now, do you? You only really need `client_port` to filter the `jobs` table and `client_id` will be the more robust option once it's implemented anyway.

danielmewes:
@marshall007 I agree that we shouldn't worry about it too much, since as you say the `client_id` will solve the issue better relatively soon.
If there are multiple client servers, the `client_port` is not going to be sufficient actually, since different client servers can assign the same client port to one of their connections.

In my brief test the Python driver actually did give me a numeric address though. I'm not sure under which circumstances it will report a hostname instead.
It sounds like `socket.getaddrinfo` is the way to get a numeric address from an arbitrary hostname or address string: https://docs.python.org/2/library/socket.html#socket.getaddrinfo
Do you think it would make sense to call that as part of the `client_address()` implementation for the Python driver, to make sure that we resolve any hostnames that might appear before returning the address to the user?

marshall007:
@danielmewes oh yea, I wasn't thinking about connecting to remote instances.

After looking into it more I think you only get back a hostname in Python if you construct the socket instance with a hostname *and* DNS resolution fails. If that is indeed true then I think we're fine not doing anything.

danielmewes:
Self-assigning for making a final check, so we can merge this on time before 2.3.

nighelles:
Merged

deontologician:
This was before the Java driver, so there should probably be a corresponding change there

danielmewes:
@deontologician Good call. Opened https://github.com/rethinkdb/rethinkdb/issues/5571

PR #4818, opened 2015-09-09
Implement a custom encoder to match the decode route.

grandquista:
This makes the query encoding recurse one fewer times like the response improvement.


danielmewes:
Awesome, thanks for another great pull request @grandquista ! :)

@Tryneus when you have some available time, could you take a look at this please?
(I think this lower priority than some of the other changes for 2.2, but would be nice to get in of course)

Tryneus:
Looks good to me, merging.  Thanks, @grandquista!

PR #4825, opened 2015-09-10
Simple JSON override interface

grandquista:
This depends on #4818

### DIFF
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -149,6 +149,7 @@ class Cursor(object):
         self.outstanding_requests = 0
         self.threshold = 1
         self.error = None
+        self._json_decoder = self.conn._get_json_decoder(self.query)
 
         self.conn._cursor_cache[self.query.token] = self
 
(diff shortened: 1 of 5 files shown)


danielmewes:
You're awesome @grandquista ! :)

(mentioning @Tryneus to put this on your screen. I'm going to move #4818 and this one into the 2.2 milestone)

drivers/python/rethinkdb/net.py
```diff
@@ -401,12 +403,14 @@ def _get_ssl_context(self, ca_certs):
 
 
 class ConnectionInstance(object):
-    def __init__(self, parent):
+    def __init__(self, parent, json_encoder=None, json_decoder=None):
```
Tryneus: The `json_encoder` and `json_decoder` code appears to be repeated verbatim for each type of connection.  This code should probably be put straight into the `Connection` class, which is used as the base class for the different types - `DefaultConnection`, `net_twisted.Connection`, `net_tornado.Connection`, and `net_asyncio.Connection`.

Since this code is common to all the connection types (and doesn't really rely on any ConnectionInstance variables), `Connection` should just pull `json_encoder` and `json_decorder` out of the kwargs, and provide the `_get_json_decoder` and `_get_json_encoder` functions that the child classes will inherit.

Tryneus:
@grandquista - looks good, but I had a comment to try and reduce the code duplication.  Thanks for submitting this!

drivers/python/rethinkdb/net.py
grandquista: I have changed the solution to use `Connection` properties and reduce code.

Tryneus:
Awesome, merging.  Thanks again for the help, @grandquista!

PR #4834, opened 2015-09-12
added systemd-hint on pid-file parameter

RubenKelevra:


packaging/assets/config/default.conf.sample
```diff
@@ -21,6 +21,7 @@
 ## Stash the pid in this file when the process is running
 ## Command line default: none
 ## Init script default: /var/run/rethinkdb/<name>/pid_file (where <name> is the name of this config file without the extension)
+## Systemd default: Uses internal mechanism, do not set this paramtert.
```
danielmewes: I think rather than "Systemd default:", this should say something like "Note for systemd users: Systemd uses its own internal mechanism, do not set this parameter.".
There's also a typo in "parameter".

(in general the default for systemd is the same as the command line default, but that's not specific to this option)

danielmewes:
Thanks for the PR @RubenKelevra .
I added a comment. Looks good otherwise.

Have you signed our CLA yet? http://rethinkdb.com/community/cla/

danielmewes:
@RubenKelevra Just checking in: Do you prefer updating your PR yourself or do you want us to make the changes instead?

RubenKelevra:
@danielmewes feel free to edit :)

danielmewes:
I committed something inspired by your PR in https://github.com/rethinkdb/rethinkdb/commit/86cb86ee41cecddf146aa5ae3c20509697bd1be8 .

PR #4842, opened 2015-09-16
2.1.4 Release notes draft

danielmewes:
@mglukhovsky or @dalanmiller could you review please?

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,11 +1,11 @@
 # Release 2.1.4 (Forbidden Planet)
 
-Released on 2015-10-15
+Released on 2015-09-16
 
 Bug fix release
 
-* Fixed a second data corruption bug in the b-tree implementation (#4769)
-* The web UI now displays table configuration errors correctly (#4811)
+* Fixed a data corruption bug that could occur when deleting documents (#4769)
+* The web UI no longer ignores errors during table configuration (#4811)
 * Added a check in case `reconfigure` is called with a non-existent server tag (#4840)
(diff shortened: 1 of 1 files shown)


NOTES.md
```diff
@@ -1,3 +1,19 @@
+# Release 2.1.4 (Forbidden Planet)
+
+Released on 2015-10-15
```
danielmewes: (need to fix the date)

NOTES.md
```diff
@@ -1,3 +1,19 @@
...
+
+Bug fix release
+
+* Fixed a second data corruption bug in the b-tree implementation (#4769)
```
mglukhovsky: Why is it important to mention that it's a second bug?

B-tree should be capitalized.

mglukhovsky: Could you add clarity around what kind of bug it is?

dalanmiller: "Fixed a data corruption..."

NOTES.md
```diff
@@ -1,3 +1,19 @@
...
+* The web UI now displays table configuration errors correctly (#4811)
```
dalanmiller: Given what former state when it wouldn't?

NOTES.md
danielmewes: I wrote "second" because it's the same issue number as a fix with the same description in the 2.1.3 release.
I'll try to add more details instead to distinguish it.

danielmewes:
Thanks @mglukhovsky and @dalanmiller . I updated the draft.

mglukhovsky:
Looks great!

PR #4846, opened 2015-09-16
Drivers: support top-level aggregation functions

### LINKED ISSUE #4594: r.min(a, b) / r.max(a, b) variants
Currently you have to write these as
```js
r.expr([a, b]).min()
r.expr([a, b]).max()
```

It would be nice if you could just write
```js
r.min(a, b)
r.max(a, b)
```

probably these should be variadic and allow for an arbitrary number of arguments (at least 2 to distinguish them from the other case).

Suggested by @weshoke

marshall007:
Fixes #4594. The following are now exposed as top-level terms:

- `sum`
- `avg`
- `min`
- `max`
- `distinct`

Note as per discussion in #4594, these terms were not overloaded. Meaning `r.min(<expr>, <expr>)` still won't work but `r.min([<expr>, <expr>])` now will.


marshall007:
Since we're not overloading, I'm tempted to go ahead and support this for the remaining aggregation terms (`reduce`, `count`, `contains`, even `group`?). Thoughts?

There's no obvious reason why you'd ever want to use the infix `r.count([ ... ])`, but I think it's slightly better that you *could* if you wanted to rather than us making arbitrary distinctions based on what we think people might actually use.

danielmewes:
Thanks @marshall007 !

I'm not sure if we want to do this for all aggregation terms, though I don't have any particular objection to it, and I can see the added consistency in that.

We usually try to do all discussion of ReQL related proposals in a certain time frame early during the release planning, and it's now a bit too late for squeezing that into 2.2.

I don't think there's any controversy about the terms you currently changed in this PR, so I would be happy to merge those into 2.2 and leave the discussion of the wider change that includes `reduce`, `count`, `group` etc. for 2.3.

Would it be ok if we did both together for 2.3, or would you prefer to have us merge what you've done so far into 2.2, and then have a second discussion for 2.3?

marshall007:
@danielmewes totally up to you guys. Merging everything at once for 2.3 sounds good to me. I just happened to run into this yesterday; it's not something I encounter very often.

Or if, before the 2.2 release, someone comes up with good reasons not never include `reduce`, `count`, and `group` then we could go ahead and merge this as-is.

danielmewes:
Ok I'm going to move this into subsequent / 2.3, and we'll discuss the details of it then.

danielmewes:
@mlucy do you have any opinions on `r.reduce`, `r.count` and `r.group` as top-level terms?

I have no objections.

mlucy:
Giving those top-level variants seems good to me.  Personally I would add top-level variants for everything the same way we do in Ruby.

marshall007:
@danielmewes sounds like we're all in agreement. Should I go ahead and add the other terms?

danielmewes:
@marshall007 Sounds good :+1:

marshall007:
@danielmewes I added the remaining terms, as discussed. Also noticed while testing `r.group` that the arguments needed to be wrapped for all these terms, so I did that and added additional polyglot tests to cover it.

deontologician:
Putting a reminder here that I need to add these for the Java driver. Normally you could just add a line to generate a top-level method, but right now there's a bug that prevents it

danielmewes:
Thanks @marshall007 
@deontologician could you please review this when you're back (and add the Java implementation)?

deontologician:
Reviewed. It's pretty straightforward, I just made one small formatting change.

The java changes are in as well, the bug was that it wasn't looking at changes in the json file to determine if it should re-render (it was only looking at the age of the templates and the rendering script itself).

PR #4847, opened 2015-09-16
Fix for 4845

eric-poitras:
Hi,

here is a fix to the bug 4845 I just entered. I also upgraded the test class to allow "override" of the configuration without having to modify any code.

The idea is to include a file in the root of the java driver called "test-config-override.properties" that can contains the hostName and port parameters. This file is git-ignored so nobody commit it to the repo.

If not defined, those parameters default to "newton" and 31157 which are Josh's parameters, I guess.

Have a nice day !

### DIFF
--- a/drivers/java/.gitignore
+++ b/drivers/java/.gitignore
@@ -3,3 +3,4 @@ build/
 .#*
 *.iml
 .idea/
+test-config-override.properties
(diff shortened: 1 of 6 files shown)


deontologician:
Wow, this is awesome. Have you signed the CLAalready by chance? (http://rethinkdb.com/community/cla/) ?

eric-poitras:
Yep ! It is signed.

PR #4889, opened 2015-09-26
Support for insert and select operations with POJOs in Java driver

igorlukanin:
Added support for select and insert operations on Java bean classes, not only on boxed primitive types, lists and maps. You can select one typed object:

```java
Pojo pojo = r.db("db").table("table").get("reql-expr").run(conn, Pojo.class);
```
...or a list of objects:
```java
List<Pojo> pojoList = r.db("db").table("table").get("reql-expr").runList(conn, Pojo.class);
```

Objects are constructed from ```Map<String, Object>``` instances selected from RethinkDB using reflection. Insertion of typed objects employs the same technique:

```java
Pojo pojo = new Pojo("foo", new NestedPojo(42L, true));
r.db("db").table("table").insert(pojo).run(conn);
```

Merging this code to Java driver would be extremely useful for me and, I presume, to the community, because working with maps and primitives is quite painful in Java.

I bet my code removes a portion of that pain :smile:

### DIFF
--- a/drivers/java/build.gradle
+++ b/drivers/java/build.gradle
@@ -2,6 +2,10 @@ apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'signing'
 
+tasks.withType(JavaCompile) {
+    options.compilerArgs << "-parameters"
+}
+
 version = '2.2-b1'
 ext.isReleaseVersion = !version.endsWith("b1")
 group = "com.rethinkdb"
--- a/drivers/java/src/main/java/com/rethinkdb/ast/ReqlAst.java
+++ b/drivers/java/src/main/java/com/rethinkdb/ast/ReqlAst.java
@@ -83,9 +83,11 @@ public class ReqlAst {
     /**
      * Runs this query via connection {@code conn} with default options and returns an atom result
      * or a sequence result as a cursor. The atom result representing a JSON object is converted
-     * to an object of type {@code Class<P>}. The cursor is a {@code com.rethinkdb.net.Cursor}
-     * which may be iterated to get a sequence of atom results of type {@code Class<P>}
+     * to an object of type {@code Class<P>} specified with {@code pojoClass}. The cursor
+     * is a {@code com.rethinkdb.net.Cursor} which may be iterated to get a sequence of atom results
+     * of type {@code Class<P>}
      * @param conn The connection to run this query
+     * @param pojoClass The class of POJO to convert to
      * @param <T> The type of result
      * @param <P> The type of POJO to convert to
      * @return The result of this query (either a {@code P or a Cursor<P>}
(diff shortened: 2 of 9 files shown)


igorlukanin:
I guess @deontologician is an appropriate assignee for this PR.

deontologician:
Whoah! This is really beautiful. I will check it out

igorlukanin:
Before last commit, this code wasn't supposed to work:
```java
public class Pojo  {
    private Integer integerProperty;
    public Pojo(Integer integerProperty) { this.integerProperty = integerProperty; }
}
Pojo pojo = r.db("db").table("table").get("reql-expr").run(conn, Pojo.class);
```

A public parameterless constrcutor and a setter were required:
```java
public class Pojo  {
    private Integer integerProperty;
    public Pojo() {}
    public Pojo(Integer integerProperty) { this.integerProperty = integerProperty; }
    public void setIntegerProperty(Integer integerProperty) { this.integerProperty = integerProperty; }
}
Pojo pojo = r.db("db").table("table").get("reql-expr").run(conn, Pojo.class);
```

Last commit adds support for beans with no parameterless constructors. Now if a bean has a constructor with several parameters matching both by names and types with the map of values retrieved from RethinkDB, it can still be instantiated and populated with values.

Names of parameters are only available in Java 8 for code compiled with ```javac -parameters``` (see https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html), but that should be acceptable for lots of cases.

So, in Java we may stick to this short POJO definitions:
```java
public class Pojo  {
    private Integer integerProperty;
    public Pojo(Integer integerProperty) { this.integerProperty = integerProperty; }
}
```

And in other JVM languages, like Kotlin (see http://kotlinlang.org/docs/reference/data-classes.html), we may use even more concise version:
```kotlin
data class Pojo(val integerProperty: Integer)
```

deontologician:
Ok this is next on my list, once the java branch is merged to next I'm going to adapt this to work on the current version of the driver. Primarily I want to collapse the runList into one that infers the list based on the result type, which should also work for cursors

deontologician:
Hey @igorlukanin so this is definitely something we should do, but there's a couple of issues with the PR as it is now:

1. It has conflicts (this is kinda my fault since it's been so long since you made the PR)
2. It doesn't handle cursors
3. I think the .runList should just happen through .run

The strategy I'm thinking of is passing the .class object down through to `runQuery` in the `Connection` class, and having the cursor be aware of what class it's serializing to, as well as mapping over List when it's returned. This way you can do `.run(conn, options, Foo.class)` and get a `Cursor<Foo>` back, or a `List<Foo>` depending on what the server returns.

Thoughts?

igorlukanin:
@deontologician Hi Josh! Sounds good. I will rebase this PR on top of this branch and have a look at making it work for cursors. Just give me some time, I'll get in touch.

igorlukanin:
@deontologician Josh, now this PR should probably be OK. Would you have a look at last 5 commits?

igorlukanin:
Though I'm not sure whether it's OK to throw a raw `ClassCastException` when the class is not specified. It definitely may be wrapped with a proper `ReqlError` ancestor.

You can see an example of code throwing an exception in [tests](https://github.com/igorlukanin/rethinkdb/commit/fe70375de07ae97045328187d19a00c40a70cfab#diff-5d1cefb097beeae2990c82eff92392daR261).

deontologician:
I think the classCastException is fine, it doesn't have an analogue in the other drivers and it's probably what the user expects to happen if the class doesn't work.

All of the tests are passing, I think this looks good! I'm going to merge it. Thanks again @igorlukanin

igorlukanin:
> All of the tests are passing

@deontologician Do you mean 'tests from the default package'? Because these tests from the 'gen' package fail for me, though I don't think it's related to this PR:

`ChangefeedsIncludeStates. test`
`DatumArray. test`
`DatumObject. test`
`DatumUuid. test`
`GeoPrimitives. test`
`MetaTable. test`
`Range. test`
`SindexNullsinstrings. test`

deontologician:
There are several tests that rely on the version of the server currently in next. I just meant the RethinkDBTest

PR #4914, opened 2015-10-03
Date precision loss in Java driver

igorlukanin:
Upstream implementation of Date-to-ReqlAst conversion causes a precision loss: seconds and milliseconds are gone. The conversion of value returned from database also causes milliseconds to be lost, because the double value is truncated before it's multiplied by 1000.

Fixed both issues and added a test.

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/ast/Util.java
+++ b/drivers/java/src/main/java/com/rethinkdb/ast/Util.java
@@ -77,7 +77,7 @@ public class Util {
 
         if (val instanceof Date) {
             TimeZone tz = TimeZone.getTimeZone("UTC");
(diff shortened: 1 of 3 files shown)


igorlukanin:
@deontologician, will you, please, have a look at this PR and merge it to upstream?

deontologician:
Yep, looks good, thanks for the PR

mlucy:
@deontologician -- just to double check, has @igorlukanin signed the CLA before?

igorlukanin:
@mlucy Yes, I did :)

deontologician:
Yeah, he made a previous contribution to the Java driver, so I knew it wasn't an issue

Issue #4902, opened 2015-09-30
allow change feed notifications for raft related data


mbroadst:
My immediate use case is to be informed of whether a given rethinkdb instance has become the leader, however this could potentially be useful for other cases. Temporarily I can set up a change feed on the logs table and filter for the term "Raft", however a more direct means would be ideal. 

Though this was discussed briefly on the mailing list, no actual proposal was put forth. If this was something as simple as exposing a key like "raft_role" maybe on the `server_status` table (it looks like here: https://github.com/rethinkdb/rethinkdb/blob/next/src/clustering/administration/servers/server_status.cc#L40), then I could totally put together a PR for that, however I wanted to open up the issue here first to discuss potentially better alternatives (or interest in the first place!).

coffeemug:
/cc @danielmewes 

Daniel, what do you think about this?

danielmewes:
A quick clarification ahead: RethinkDB uses one raft cluster per table. That means that for each table, a different server can be the Raft leader.

How about exposing a `raft_leader` field in `table_status`, which gets set to the server that's currently the Raft leader (or `null` if there is no known leader)?
I quite like that idea.

Also pinging @VeXocide for any thoughts on this.

mbroadst:
@danielmewes ah right, I did fail to notice that this is per-table! Other than that, `raft_leader` sounds like a great idea +1

mbroadst:
In the meantime, I've ended up with this reql to watch for leadership changes on a given table (`r.db('test').table('changing_thing')` in this case). The query currently takes ~350ms on my localhost, can you think of a way to speed it up? Unfortunately it doesn't seem that the logs provide any relevant index for me to key on, and the table id is a string I need to match for.

```
r.db('test').table('changing_thing').config()
  .do(function(config) {
    return r.db('rethinkdb').table('logs')
      .filter(function(log) {
        return log('message').match(config('id')).and(log('message').match('Raft leader'));
      });
  });
```

mbroadst:
strangely enough if I change the match line from:
```
return log('message').match(config('id')).and(log('message').match('Raft leader'));
```
to
```
return log('message').match(config('id').add(': This server is Raft leader'));
```

I get a slightly slower query (but this is all based on Data Explorer, so I imagine the numbers are close enough)

danielmewes:
I think reading the full log table might be what's slow here. It's not parsed in the most efficient way at the moment. You can make it a bit faster by stopping once you've found the first match:
[...]

PR #4942, opened 2015-10-11
feat(raft_leader): display raft leader for table in table_status

mbroadst:
This adds support for tracking the raft leader for each table and
displaying that information per-table in the table_status
artificial table.

relates to issue https://github.com/rethinkdb/rethinkdb/issues/4902

I know there probably is a lot to review here, as I'm clearly not very familiar with your code base. Off the top of my head I have the following comments/concerns:

  * The serialized form of `table_status_response_t` has changed, making it incompatible with older versions of the database. I had considered packing this information into `table_raft_state_t` but wasn't sure if that would even make a difference, so it would be helpful if someone could advise here (I tacked it on to the end of the serialization in case order mattered here).

  * The location of the data itself is just set as `raft_leader` on the top level of the `table_status` documents, this is easily changeable and I think it's just a matter of consensus on your side of where it belongs (`raft_leader` or `raft.leader` etc)

  * Currently the raft_leader information piggybacks a request for `want_raft_state` which may be misleading, but let's me get away with not having to change the request format.

  * The way I'm currently getting the server name seems a bit unorthodox, there are comments in your code describing its flakey nature in production

  * In order to determine if a table is a raft leader I'm statically casting the `leader` member of a table manager, and before that process I use the leader rw locker. In other parts of the codebase where you something similar you allow for the possibility of it already being locked (sort of looks like a try_lock) due to leadership changes - this might be trouble if I don't do it here

Anyway, I'm sure there's more but it's a start.

### DIFF
--- a/src/clustering/administration/tables/calculate_status.cc
+++ b/src/clustering/administration/tables/calculate_status.cc
@@ -79,6 +79,9 @@ void get_table_status(
     status_out->total_loss = false;
     status_out->config = config;
 
+    /* Determine raft leader for this table */
+    table_meta_client->get_raft_leader(table_id, interruptor, &status_out->raft_leader);
+
     /* Send the status query to every server for the table. */
     bool all_replicas_ready;
     try {
--- a/src/clustering/administration/tables/calculate_status.hpp
+++ b/src/clustering/administration/tables/calculate_status.hpp
@@ -57,6 +57,7 @@ public:
     table_config_and_shards_t config;
     std::map<server_id_t, range_map_t<key_range_t::right_bound_t,
         table_shard_status_t> > server_shards;
+    boost::optional<server_id_t> raft_leader;
     std::set<server_id_t> disconnected;
     server_name_map_t server_names;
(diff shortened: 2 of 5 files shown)


mbroadst:
hmm yeah after experimenting with modifying `table_raft_state_t` instead, we'll run into the same serialization issues, so either way this change breaks compatibility

danielmewes:
Awesome, thanks for the PR @mbroadst.

As mentioned on IRC, changing the serialization of `table_status_response_t` is fine. I think changing `table_raft_state_t` actually isn't an option here, because that's the state that's maintained *through* Raft (i.e. the table configuration etc.).

@VeXocide Could you review the code please when you have some time?

VeXocide:
Over IRC @mbroadst and I found a route that doesn't require changing the `table_status_response_t`, it uses the data available in the `table_manager_directory` in the `table_meta_client_t`.

VeXocide:
Thanks again @mbroadst!

Issue #3133, opened 2014-10-01
Make coerceTo("BOOL") work on all types?


danielmewes:
Currently we have:
```
r.expr(null).coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce NULL to BOOL in:
r.expr(0).coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce NUMBER to BOOL in:
r.expr("false").coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce STRING to BOOL in:
r.expr("foo").coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce STRING to BOOL in:
r.expr([]).coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce ARRAY to BOOL in:
r.expr({}).coerceTo('BOOL')
    RqlRuntimeError: Cannot coerce OBJECT to BOOL in:
```

It seems that `coerceTo('BOOL')` basically doesn't do anything at all. Is that correct?

Here are three options of what we could do about it:

1. Leave it as it is now, without a real use
2. Make it consistent with `r.branch(x, true, false)`, i.e.
```
r.expr(null).coerceTo('BOOL')
    false
r.expr(0).coerceTo('BOOL')
    true
r.expr("false").coerceTo('BOOL')
    true
r.expr("foo").coerceTo('BOOL')
    true
r.expr([]).coerceTo('BOOL')
    true
r.expr({}).coerceTo('BOOL')
    true
```
'3. Try to be a bit "smarter" with respect to numbers and strings:
```
r.expr(null).coerceTo('BOOL')
    false
r.expr(0).coerceTo('BOOL')
    false
r.expr(1).coerceTo('BOOL')
    true
r.expr(50).coerceTo('BOOL')
    ERROR: Only 0 and 1 can be coerced to BOOL.
r.expr("").coerceTo('BOOL')
    false (?)
r.expr("false").coerceTo('BOOL')
    false
r.expr("true").coerceTo('BOOL')
    true
r.expr("foo").coerceTo('BOOL')
    ERROR: Only "true" and "false" can be coerced to BOOL.
r.expr([]).coerceTo('BOOL')
    true
r.expr({}).coerceTo('BOOL')
    true
```

mlucy:
I agree we should go for consistency with `r.branch`.

VeXocide:
I'm not sold on how `r.branch` currently handles it, but I agree we should go for consistency.

[...]

PR #4996, opened 2015-10-25
Fix for #3133

eliangidoni:


danielmewes:
Thanks, the patch looks good to me.

Are you familiar with the tests in `test/rql_test`? You can run them through the `test-runner` script in that folder. The test definitions are in `src` (`test-runner` refers to those tests as `polyglot` tests). It would be great to have a test for the new behavior.
If you don't want to deal with that, we can also add the tests later on our side.

eliangidoni:
@danielmewes  I've just updated the tests, I'm not familiar with them though, I hope they're ok.

danielmewes:
Looks good! Thanks for adding the test.

PR #5052, opened 2015-11-09
MakeObject.tt should be MAKE_OBJ not MAKE_OBJECT

mikemintz:
This fixes an incorrect tt field in MakeObject for the JavaScript driver's AST.

This doesn't matter for typical use cases since objects are JSON encoded without a wrapping array and term type. But this makes introspection of AST terms more accurate.


test/rql_test/importRethinkDB.js
```diff
@@ -9,6 +9,8 @@ if (fs.existsSync(path.resolve(rethinkdbLocation, 'rethinkdb.js')) == false) {
     process.exit(1);
 }
 var r = require(path.resolve(rethinkdbLocation, 'rethinkdb'));
+var protodef = require(path.resolve(rethinkdbLocation, 'proto-def'));
```
danielmewes: Pinging @larkost about this change

danielmewes:
Good catch! The fix and test look good to me.

Waiting on @larkost to give his ok on the small testing system change.

danielmewes:
Merged into `next`. This will ship with RethinkDB 2.3. Thanks again @mikemintz

Issue #3930, opened 2015-03-13
Create official Java driver


deontologician:
https://github.com/npiv/rethink-java-driver might be a good jumping off point. It seems both of the 3rd party drivers for Java are abandoned or very out of date

danielmewes:
I think we should try to get a beta release out by 2.1, especially since we might abandon protocol buffer support with that release and @npiv's driver is still using it.

Later we'll need to integrate it into our polyglot testing framework.

Another possible starting point for a driver is @dkhenry's Java driver. I've already ported it to the JSON protocol for benchmarking purposes in https://github.com/danielmewes/rethinkjava , but in contrast to @npic's driver it's missing support for lambdas.

deontologician:
Most likely the serialization stuff is easier to port from one to the other. There are some language design issues to consider as well though, I think the default should return a cursor, rather than needing runForCursor. So in any case, there would be a lot of fiddling to get iit into shape

kclay:
I've been meaning to add more javaness to my Scala driver(   https://github.com/kclay/rethink-scala) since it enforces type safety and provides methods for casting, blocking and async modes etc. On @deontologician note about the need for `runForCursor`. The driver is built in a way that its able to determine if the result should be a single value or an cursor. Pretty sure I can refactor some of this to work with java since some parts relying strongly on Scala implicits. Just a thought.

stephanos:
@deontologician 
Will the new driver support Java 7?

stephanos:
@danielmewes 
I can't even get your fork to compile! After some fixing I am stuck with `Error:(296, 37) java: incompatible types: long cannot be converted to com.google.protobuf.CodedInputStream` (RqlConnection.java) @ https://github.com/stephanos/rethinkjava

danielmewes:
@stephanos I saw you did some patches in your branch, did you get it to compile eventually? I'm not sure why it's not compiling (though I'm not surprised, I've only ever used it on one type of system for a single purpose). My guess would be an incompatible protocol buffer library or compiler.

The official driver will not be based on that code base by the way.

stephanos:
@danielmewes no, I didn't :( -will have to wait for the official driver.

danielmewes:
@stephanos another driver you can try until then is https://github.com/dropbox/rethink-java-driver

deontologician:
[...]

PR #5053, opened 2015-11-09
Added a test for the '.filter(...)' function in the Java driver

igorlukanin:
I wrote this test as a tiny workaround to the lack of Java driver docs mentioned [here](https://github.com/rethinkdb/rethinkdb/issues/3930#issuecomment-154913718). I hope it's worth adding to the driver code base.

I think that @deontologician is the appropriate assignee for this PR.

### DIFF
--- a/drivers/java/src/test/java/RethinkDBTest.java
+++ b/drivers/java/src/test/java/RethinkDBTest.java
@@ -11,9 +11,7 @@ import org.junit.*;
 import org.junit.rules.ExpectedException;
 
 import java.time.OffsetDateTime;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks for the PR @igorlukanin .

We actually have some tests that are automatically generated from the yaml files in `test/rql_test/src`, including a wide variety of `filter` tests. @deontologician can tell you how to run them.

@deontologician Do you think it makes sense to add a manual test like this anyway? I think it's probably not that useful, given that we already have tests for this.
I can see a use as a code example, but we will have full API documentation for the Java driver including `filter` examples done very soon.

deontologician:
I don't think we have any autogenerated filter tests currently since they likely use r.row somewhere in the file

danielmewes:
Oh I see. I didn't consider that.

deontologician:
I'll merge it, since it can't hurt until the other tests are convertible. I think ultimately the RethinkDBTest file may end up the beginning of the connection-like tests

PR #5075, opened 2015-11-11
2.2.0 release notes

danielmewes:
@mglukhovsky @AtnNn could you review please?


NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+
+### API-breaking changes ###
+
+* Changefeeds on `.orderBy.limit` as well as `.get` queries previously provided
+  initial results by default. You now need to pass the `includeInitial: true` optarg to
+  `.changes` to achieve the same behavior.
+* The protocol buffer driver protocol is no longer supported. The JSON-based protocol is
+  now the only supported driver protocol. Certain older drivers (e.g. the Java drivers by
+  @dkhenry and by @npiv) no longer work with RethinkDB 2.2.0. See the [drivers][drivers]
+  list for up-to-date drivers.
```
AtnNn: It seems wrong to explicitly mention that those Java drivers no longer work without mentioning the upcoming official driver.

AtnNn:
The contributors list seems smaller than usual. Does it include contributors to the docs repo?

NOTES.md
```diff
@@ -1,3 +1,143 @@
+# Release 2.2.0 (Modern Times)
+
+Released on 2015-11-XX
+
+Release highlights:
+
+* Atomic changefeeds
+* Performance & Scalability improvements
```
mglukhovsky: >  Performance and scalability improvements

danielmewes:
@AtnNn Good point, these are only distributors to the core repo. I wasn't sure if we usually included docs contributors. I'll add them.

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+
+Read the [blog post][2.2-release] for more details.
+
+[2.2-release]: http://rethinkdb.com/blog/2.2-release/
+
+## Compatibility ##
+
+Data files from RethinkDB versions 1.16.0 onward will be automatically migrated.
```
mglukhovsky: Can we just say "versions 1.16 onward"?

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+As with any major release, back up your data files before performing the upgrade.
+
+If you're upgrading from RethinkDB 1.14.x or 1.15.x, you need to migrate your secondary
+indexes first. You can do this by following these steps:
+* install RethinkDB 2.0.5
```
mglukhovsky: These should be sentences:

- Install RethinkDB 2.0.5.
- ...
- Run `rethinkdb index-rebuild`.

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+* run `rethinkdb index-rebuild`
+
+Afterwards, you can install RethinkDB 2.2 and start it on the existing data files.
+
+If you're upgrading directly from RethinkDB 1.13 or earlier, you will need to manually
+upgrade using `rethinkdb dump`.
+
+### API-breaking changes ###
+
+* Changefeeds on `.orderBy.limit` as well as `.get` queries previously provided
```
mglukhovsky: > You now need to include the optional argument `includeInitial: true` to `.changes` to achieve the same behavior.

NOTES.md
mglukhovsky: Agreed with Etienne. It would be good to not call out two specific drivers, especially because this implies that there is now no way to use RethinkDB and Java.

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+
+[drivers]: http://rethinkdb.com/docs/install-drivers/
+[issue-4669]: https://github.com/rethinkdb/rethinkdb/issues/4669#issue-100428248
+
+### Supported distributions ###
+
+RethinkDB 2.2.0 now comes with official packages for Ubuntu 15.10 (Wily Werewolf) and
+CentOS 7.
+
+We no longer provide packages for Ubuntu 10.04 (Lucid Lynx).
```
mglukhovsky: I'd mention, "..., which has reached end of life."

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+
+* Added full support for atomic changefeeds through the `include_initial` optarg (#3579)
+* Added a `values` command to obtain the values of an object as an array (#2945)
+* Added a `conn.server` command to identify the server for a given connection (#3934)
+* Extended `r.uuid` to accept a string and work as a hash function (#4636)
+
+## Improvements ##
+
+* Server
+ * Improved the scalability of many range queries on sharded tables (#4343)
```
mglukhovsky: Remove "many", it makes the sentence ambiguous (many queries at one time, or many common range queries?)

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+ * Improved the performance of secondary-index-based `between` queries (#4862)
```
mglukhovsky: Improved the performance of `between` queries based on secondary indexes.

NOTES.md
danielmewes: >  especially because this implies that there is now no way to use RethinkDB and Java.

... which is correct. That's the core implication that I want to make here. "Java users please don't upgrade yet."

Mentioning the upcoming official driver sounds like a good idea though. I might also include a link to its current snapshot.

mglukhovsky: Then I'd say something like:

> - The deprecated protocol buffer driver protocol is no longer supported. The newer JSON protocol is now the only supported driver protocol. Older drivers using the deprecated protocol no longer work with RethinkDB 2.2.0. See the [drivers][drivers]  list for up-to-date drivers.
  * If you're using Java, please note that at this time of writing, existing community drivers have not been updated to use the newer JSON protocol. However, an [official Java driver] is in active development and will be available soon.

In its current form it reads like it's calling out two drivers (and driver developers) specifically, and implies that we're abandoning Java (which isn't the case).

NOTES.md
```diff
@@ -1,3 +1,143 @@
+# Release 2.2.0 (Modern Times)
+
+Released on 2015-11-XX
+
+Release highlights:
+
+* Atomic changefeeds
+* Performance & Scalability improvements
+
+Read the [blog post][2.2-release] for more details.
```
mglukhovsky: I think that this section needs to include one or two sentences describing what is exciting about this release, for example:

> RethinkDB 2.2 introduces atomic changefeeds, which can include existing values from the database into changefeed streams. These improved changefeeds makes building realtime apps dramatically easier: you can use a single code path to populate your application with initial data, and continue receiving realtime data updates.
>
> This release also includes a whole slew of performance and scalability improvements designed to help RethinkDB clusters scale to larger sizes while using fewer resources.

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+* Extended `r.uuid` to accept a string and work as a hash function (#4636)
+
+## Improvements ##
+
+* Server
+ * Improved the scalability of many range queries on sharded tables (#4343)
+ * Improved the performance of secondary-index-based `between` queries (#4862)
+ * Reduced the memory overhead for large data sets (#1951)
+ * Redesigned the internal representation of queries to improve efficiency (#4601)
+ * Removed the protocol buffer driver protocol (#4601)
```
mglukhovsky: I'd say "the deprecated protocol buffer driver protocol"

NOTES.md
```diff
@@ -1,3 +1,143 @@
...
+* Server
+ * Fixed a segmentation fault that could happen when disconnecting a server while
+   having open changefeeds (#4972)
+ * Updated the description of the `--server-name` parameter in `rethinkdb --help` (#4739)
+ * Fixed a crash with the message "Guarantee failed: [ts->tv_nsec >= 0 &&
+   ts->tv_nsec < (1000LL * (1000LL * 1000LL))] " (#4931)
+ * Fixed a problem where backfill jobs didn't get removed from the `'jobs'` table (#4923)
+ * Fixed a memory corruption that could trigger a segmentation fault during
+   `getIntersecting` queries (#4937)
+ * Fixed an issue that could stop data files from RethinkDB 1.13 to be migrated properly
```
mglukhovsky: > Fixed an issue that could stop data files from RethinkDB 1.13 from migrating properly

mglukhovsky:
@danielmewes: OK, left my notes.

NOTES.md
danielmewes: Thanks @AtnNn and @mglukhovsky . That formulation sounds great. I'll use that.

danielmewes:
Thanks a lot for the feedback @mglukhovsky and @AtnNn . I uploaded an updated draft.

NOTES.md
danielmewes: Sounds good, thanks!
I slightly modified the wording, so you might want to double-check that it still sounds good to you @mglukhovsky .

danielmewes:
I'm merging this for now so that I can tag and start the build before leaving to the meetup. We can still polish a few details for the GitHub release listing later.

NOTES.md
```diff
@@ -1,3 +1,165 @@
+# Release 2.2.0 (Modern Times)
+
+Released on 2015-11-12
+
+RethinkDB 2.2 introduces atomic changefeeds. Atomic changefeeds include existing values
+from the database into the changefeed result, and then atomically transition to streaming
+updates.
```
mglukhovsky: If this is a new paragraph, there should be a newline here.

mglukhovsky:
Looks great, barring one missing newline. :+1:

Issue #5043, opened 2015-11-06
ConnectionInstance._reader() Task was destroyed but it is pending


danielnelson:
The code below produces an warning from asyncio.
```
$ cat test.py
#!/usr/bin/env python
import asyncio
import rethinkdb as r

async def stuff():
    conn = await r.connect()
    await conn.close()

r.set_loop_type("asyncio")
loop = asyncio.get_event_loop()
task = loop.create_task(stuff())
loop.run_until_complete(task)
$ python test.py
Task was destroyed but it is pending!
task: <Task pending coro=<ConnectionInstance._reader() running at lib/python3.5/site-packages/rethinkdb/net_asyncio.py:218> wait_for=<Future finished result=None>>
```

If I run another cycle of the loop by adding a second call to `run_until_complete`, then there is no warning.  Perhaps `ConnectionInstance.close()` should yield from its `self._reader_task` before returning?

danielmewes:
@Tryneus do you have any idea about this?

sontek:
I'm also seeing this with my test of using asyncio:

```
@asyncio.coroutine
def asyncio_query(future, survey_id):
    data = []
    tconn = yield from r.connect("mt1-rethinkd1c1")
    tconn._get_json_decoder = create_decoder
    print("survey", survey_id)
    q = r.table(TABLE).filter({'survey_id': survey_id})
    cursor = yield from q.pluck(
        "survey_id", "respondent_id", "row_id", "column_id", "value_id"
    ).map(r.row.values()).run(tconn)

    while(yield from cursor.fetch_next()):
        row = yield from cursor.next()
        data.append(row)

    cursor.close()
    tconn.close()
    print("survey: %s, data: %s" % (survey_id, len(data)))
    future.set_result(data)

def get_async_io():
    r.set_loop_type("asyncio")
    loop = asyncio.get_event_loop()

    futures = []
    for i in range(1, 16):
        future = asyncio.Future()
        asyncio.async(asyncio_query(future, i))
        futures.append(future)

    loop.run_until_complete(asyncio.gather(*futures))
    data = []

    for future in futures:
        data += future.result()
    print('got all data')
    loop.close()
    return data

```

danielnelson:
The workaround I've been using is to shutdown like this:
```
loop.stop()
loop.run_forever()
loop.close()
```

danielmewes:
@Tryneus could you take a look at this please?

sontek:
[...]

PR #5082, opened 2015-11-13
Add an additional yield so _reader can close down

sontek:


sontek:
@danielmewes @Tryneus This should fix #5043

danielnelson:
Here is an alternative fix that doesn't rely on sleep:
```
diff --git a/drivers/python/rethinkdb/net_asyncio.py b/drivers/python/rethinkdb/net_asyncio.py
index f1d54f2..bb43f09 100644
--- a/drivers/python/rethinkdb/net_asyncio.py
+++ b/drivers/python/rethinkdb/net_asyncio.py
@@ -194,6 +194,7 @@ class ConnectionInstance(object):
             yield from self.run_query(noreply, False)
 
         self._streamwriter.close()
+        yield from self._reader_task
         return None
 
     @asyncio.coroutine
```

sontek:
@danielnelson I like it :)

sontek:
I rebased the changes to do @danielnelson's idea instead

danielmewes:
Thanks for the PR! @Tryneus could you review this please?

Tryneus:
Looks good!  Thanks for the help, @sontek and @danielnelson, sorry I didn't have time to look into this myself.

Tryneus:
Cherry-picked the fix into `v2.2.x` in commit adbd608.

Issue #4963, opened 2015-10-16
Handle openssl dependency in MacOS X 10.11


larkost:
In the latest version of MacOS X (10.11 or El Capitan) the OpenSSL library is present, but the headers are not, even with Xcode command-line tools installed. This is intentional, and is the next ratchet in Apple's slow depreciation of OpenSSL.

The main work-around is easy, just add `--fetch openssl` to the `./configure` invocation, but we need to strengthen our checks in that script to ensure that we pickup the lack of headers and complain about it during `./configure` rather than at compile time.

larkost:
I accidentally `push`ed this to `next` rather than `publish`ing it to a new branch... whoops. However @VeXocide has now reviewed it, so we are not going revert the change.

The addition makes sure that we can compile with OpenSSL as given, and seems both to fail (with OpenSSL and no headers) and succeed (with `--fetch openssl` and with a valid OpenSSL install) as expected both on my machine and `newton`.

danielmewes:
@larkost Can you also backport it to `v2.2.x` please?

larkost:
@danielmewes: sure, done with dc0a1298

PR #5088, opened 2015-11-14
Use liressl

dparnell:
Hi,

  I was having trouble getting rethinkdb to compile under OSX El Capitan with XCode 7.1. It could not find the openssl headers so I thought I'd have a go at making things work with libressl instead of openssl. It seems to compile now, however I can't be sure that the SSL stuff is working as I've not yet been able to find how to get things to use it.

Daniel

### DIFF
--- a/configure
+++ b/configure
@@ -24,7 +24,7 @@ init () {
     must_fetch_list='bluebird v8'
     please_fetch_list="handlebars gtest re2 $must_fetch_list"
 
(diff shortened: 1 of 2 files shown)


danielmewes:
@dparnell It's mentioned in the `./configure --help` output.
The issue to track for making this work out of the box again is https://github.com/rethinkdb/rethinkdb/issues/4963 .

PR #5096, opened 2015-11-15
Use Integer.BYTES and Long.BYTES instead of hardcoding size

danielcompton:
Using Integer.BYTES and Long.BYTES makes the intent of the byte handling code clearer, and less error prone if there are future changes to these sections.

cc @deontologician


deontologician:
Hey, sorry I missed this PR. I noticed it had been auto-closed when I deleted the java-driver branch (which has been merged to next now). Would you be willing to re-create this PR targeting next? It looks good to me

danielcompton:
Yep, in #5151.

PR #5101, opened 2015-11-16
add ability to export as newline separated JSON (nsj)

tayloramurphy:
Added ability to export as newline separated JSON (nsj) as well as appropriate language.

Also added language to import.py that indicates nsj is already accepted as an import type.


drivers/python/rethinkdb/_export.py
```diff
@@ -298,14 +306,15 @@ def csv_writer(filename, fields, delimiter, task_queue, error_queue):
             pass
 
 def launch_writer(format, directory, db, table, fields, delimiter, task_queue, error_queue):
-    if format == "json":
+    if format == "json" or format == "nsj":
         filename = directory + "/%s/%s.json" % (db, table)
```
Tryneus: I think the file extension should reflect the data type - since the file can't be parsed as valid JSON, it shouldn't be named with a `.json`.

Tryneus:
@tayloramurphy, this looks pretty good to me, thanks for the pull request!

I think `nsj` is a little strange for the name of the format.  The closest thing I could find to a standard for this format was http://jsonlines.org/ (which suggests `jsonl`) or http://dataprotocols.org/ndjson/ (which suggests `ndjson`).  I think renaming it to `ndjson` would be a little cleaner and more apparent to users.

tayloramurphy:
I like `ndjson`. I'll update the PR.

Thanks for the input!

tayloramurphy:
@Tryneus Updated appropriately. It looks better as `ndjson` for sure.

Tryneus:
@tayloramurphy, awesome!  I almost forgot - one last thing before we can merge it, could you sign our CLA at  http://rethinkdb.com/community/cla/?

tayloramurphy:
Done. I tried to do it earlier but it wouldn't load properly b/c of mixed content types. Seems to have submitted properly this time.

Tryneus:
@tayloramurphy, cool, merging.  Thanks again for the pull request!

Issue #4608, opened 2015-07-31
Consider creating tables with replication by default


danielmewes:
The idea is to have a replication factor of 2 or even 3 enabled by default when creating a new table through `table_create`, assuming there are enough servers in the cluster.

I think most people who run on a cluster will want to use replication, and changing the default makes it a little harder to forget setting it up and losing data in the end when a server dies.

Putting into 2.2 for discussion.

RubenKelevra:
I'm very new to rethinkDB but I would like to set a minimum replication for my cluster. So all new tables have the minimum-level on creation.

Else it would be great to create defaults per database / per cluster for sharding and tables.

j-c-m:
I would like to see this with a default of min(3,servers) so it could handle automatic failover by default.

Maybe a separate issue, but also worth considering is if shards should default to min(32,servers).

The possibility to also store more complex clusterwide defaults to tablecreate options in a system table I think would also be good.  I would be much happier storing a default for instance of

```javascript
{ shards: 12, 
replicas: { us1: 3, us2: 3, eu1: 3 }, 
primary_replica_tag: 'us1' }
```

Compared to having to solely rely on operational process around table and database creation.

WesleyDavid:
Just a note to add a voice to this conversation:

In my scenario I keep track of many RethinkDB clusters for customers, and each cluster is spread across at least three physical hosts. Customers are responsible for their management of the actual data, so I don't cross the threshold into the running Rethink process.

In this case, customers that are unfamiliar with RethinkDB defaults had created tables, but not explicitly configured sharding and replication. The default 1 shard and 1 replica option was used for most, if not all tables. One physical host had an outage that resulted in all data being lost on that host. All customers that had tables residing only on that host were lost. In discussing with customers, it appears that the expectation of most people that aren't deeply familiar with RethinkDB is: "Well, it's a cluster so it must replicate data, right? Right!"

A default of 2 replicas, if there's enough node, would be a positive thing IMO.

danielmewes:
Thanks for adding your experience @WesleyDavid .

Unfortunately the solution for this issue got delayed a bit because we have been working on a bunch of other important features/improvements. I still think that we should change the default though.

danielmewes:
I'd probably prefer a default of 3 rather than 2 to make automatic failover work, like @j-c-m mentioned.

WesleyDavid:
[...]

PR #5142, opened 2015-11-24
Add shard/replica parameters to `rethinkdb restore|import`

### LINKED ISSUE #5140: Add parameters for `rethinkdb restore` to initialize sharding
Currently `rethinkdb restore` and `rethinkdb import` always create tables with the default settings, which means that the tables will be neither sharded nor replicated.

I propose adding `--shards` and `--replicas` parameters to `rethinkdb import` and `rethinkdb restore`, which will be passed into any `table_create` call.

Note that you can only specify one set of shard/replica parameters per import process. If different settings are needed for different databases, users could run `rethinkdb import` multiple times for only a single database or table.

marshall007:
Fixes #5140. Haven't had a chance to test this yet, though.

### DIFF
--- a/drivers/python/rethinkdb/_import.py
+++ b/drivers/python/rethinkdb/_import.py
@@ -117,8 +117,8 @@ def parse_options():
     parser.add_option("--max-document-size", dest="max_document_size",  default=0,type="int")
 
     # Replication settings
-    parser.add_option("--shards", dest="shards", metavar="NUM_SHARDS", default=None, type="int")
-    parser.add_option("--replicas", dest="replicas", metavar="NUM_REPLICAS", default=None, type="int")
+    parser.add_option("--shards", dest="shards", metavar="NUM_SHARDS", default=0, type="int")
+    parser.add_option("--replicas", dest="replicas", metavar="NUM_REPLICAS", default=0, type="int")
 
     # Directory import options
     parser.add_option("-d", "--directory", dest="directory", metavar="DIRECTORY", default=None, type="string")
@@ -154,12 +154,12 @@ def parse_options():
 
     res["auth_key"] = options.auth_key
     res["clients"] = options.clients
-    res["shards"] = options.shards
-    res["replicas"] = options.replicas
     res["durability"] = "hard" if options.hard else "soft"
     res["force"] = options.force
     res["debug"] = options.debug
     res["create_sindexes"] = options.create_sindexes
+
+    res["create_args"] = {k: options[k] for k in ['shards', 'replicas'] if getattr(options, k) != 0}
 
     # Default behavior for csv files - may be changed by options
     res["delimiter"] = ","
@@ -270,7 +270,8 @@ def parse_options():
             if options.custom_header is not None:
                 raise RuntimeError("Error: --custom-header option is only valid for CSV file formats")
 
-        res["primary_key"] = options.primary_key
+        if options.primary_key is not None:
+          res["create_args"]["primary_key"] = options.primary_key
     else:
         raise RuntimeError("Error: Must specify one of --directory or --file to import")
 
@@ -523,9 +524,9 @@ def csv_reader(task_queue, filename, db, table, options, progress_info, exit_eve
 
 # This function is called through rdb_call_wrapper, which will reattempt if a connection
 # error occurs.  Progress will resume where it left off.
-def create_table(progress, conn, db, table, pkey, sindexes, shards, replicas):
+def create_table(progress, conn, db, table, create_args, sindexes):
     if table not in r.db(db).table_list().run(conn):
-        r.db(db).table_create(table, primary_key=pkey, shards=shards, replicas=replicas).run(conn)
+        r.db(db).table_create(table, **create_args).run(conn)
 
     if progress[0] is None:
         progress[0] = 0
@@ -547,12 +548,12 @@ def table_reader(options, file_info, task_queue, error_queue, progress_info, exi
     try:
         db = file_info["db"]
         table = file_info["table"]
-        primary_key = file_info["info"]["primary_key"]
+        create_args = dict(options["create_args"])
+        create_args["primary_key"] = file_info["info"]["primary_key"]
 
         conn_fn = lambda: r.connect(options["host"], options["port"], auth_key=options["auth_key"])
-        rdb_call_wrapper(conn_fn, "create table", create_table, db, table, primary_key,
-                         file_info["info"]["indexes"] if options["create_sindexes"] else [],
-                         options["shards"], options["replicas"])
+        rdb_call_wrapper(conn_fn, "create table", create_table, db, table, create_args,
+                         file_info["info"]["indexes"] if options["create_sindexes"] else [])
 
         if file_info["format"] == "json":
             json_reader(task_queue,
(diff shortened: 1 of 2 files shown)


marshall007:
It's unclear to me why (in `_import.py`) we have both a `table_check` method which implicitly creates the table if it doesn't exist, as well as a `create_table` method, which does that as well as builds the secondary indices. If I'm not mistaken we can refactor this to just use `create_table` everywhere, but I haven't dug in too much.

danielmewes:
Thanks a bunch for the PR @marshall007 !

@Tryneus might be able to explain why there are two different places for creating a table.

marshall007:
@danielmewes no problem! It looks like `create_table` gets used on the "import directory" path, while `table_check` is used when importing a single file. So I think it does actually make sense to consolidate all that into `create_table`. Let me know what your thoughts are @Tryneus!

Tryneus:
@marshall007 - you're right, these functions have quite a bit of overlap, and it could probably be simplified.

Notably: `table_check` and `tables_check` are nearly identical in their purpose, with the following differences:
* `table_check` is called for importing a file, and `tables_check` is called for importing a database.
* `table_check` will create the table, while `tables_check` will not - this is probably because when importing a single file, we don't have sindex information.
* `table_check` may use the pkey specified in command-line arguments or the auto-generated pkey
* `tables_check` must use the pkey from an existing info file

That said, it should be possible to combine these into a single call, although some edge cases may be tricky.  As for combining these with `create_table`, there are a few more considerations:
* The call would need to be made for all tables before importing data
 * This is so users don't spend hours importing data before running into an error that could have been spotted instantly.
 * We probably don't need to handle cases where the user changes the configuration during the import - that is their own fault (i.e. running similar imports from two places or dropping a table during the import).
* If we're rewriting `create_table`, it may be a good opportunity to leverage sindex status so we don't always recreate an index.
 * If the index is outdated, we create it
 * Otherwise, we check if the index function binary matches the one we have
 * If it doesn't match, we drop it and add our index function.

Keep in mind that `create_table` is currently being used in the "import file" path, although it is more of a no-op.  If you'd like to take a shot at this, I'd be happy to review the pull request (you can do it here or separately, whichever works).  Let me know if you have any questions or concerns.

marshall007:
@Tryneus thanks for the write-up, that helps a lot! I was thinking it would be nice to consolidate things in this PR so that we don't have to pass around the shard/replica params to all these methods, but it sounds like that's a lot to think about.

@danielmewes are we going to need to wait on `2.3` for this or could we sneak it in a `2.2.x` release? If we're gonna have to wait for `2.3` anyway, then I'll just do it all at once. Otherwise I'd prefer to just get this working as-is (and refactor in a separate PR) because it would be really useful for us right now.

danielmewes:
@marshall007 We can ship this in a 2.2.x point release if it's helpful for you.

marshall007:
@danielmewes I finally got back around to this and I think it's ready to go. The admin stuff is really hard to test, but the changes were trivial so I'm fairly confident. I have the default values for `shards/replicas` set to 1 so I didn't have to add a bunch of conditional logic. That should result in the same behavior as calling `r.table_create` without those arguments, correct?

I'll hold off on the refactor until #5147 makes it easier to test/debug locally.

danielmewes:
@marshall007 Thanks, sounds good.

Is there a way we could make the default being undefined rather than 1 though? That will play more nicely with https://github.com/rethinkdb/rethinkdb/issues/4608 once we implement it.

marshall007:
@danielmewes yep done, good call.

danielmewes:
Thanks @marshall007 .
@Tryneus could you take another look please?

drivers/python/rethinkdb/_import.py
```diff
@@ -515,9 +523,9 @@ def csv_reader(task_queue, filename, db, table, options, progress_info, exit_eve
 
 # This function is called through rdb_call_wrapper, which will reattempt if a connection
 # error occurs.  Progress will resume where it left off.
-def create_table(progress, conn, db, table, pkey, sindexes):
+def create_table(progress, conn, db, table, pkey, sindexes, shards, replicas):
     if table not in r.db(db).table_list().run(conn):
-        r.db(db).table_create(table, primary_key=pkey).run(conn)
+        r.db(db).table_create(table, primary_key=pkey, shards=shards, replicas=replicas).run(conn)
```
Tryneus: I don't think this does what you intend when `shards` or `replicas` is `None`:

```py
rethinkdb.errors.RqlRuntimeError: Expected type NUMBER but found NULL. in:
r.db('test').table_create('foo', shards=None, replicas=None)
                                        ^^^^
```

It might be easier to pass down some `kwargs` `dict` that you could expand with `table_create(table, **kwargs)`, but that gets complicated a bit by `primary_key`.

drivers/python/rethinkdb/_restore.py
```diff
@@ -135,6 +140,8 @@ def do_import(temp_dir, options):
     import_args.extend(["--directory", temp_dir])
     import_args.extend(["--auth", options["auth_key"]])
     import_args.extend(["--clients", str(options["clients"])])
+    import_args.extend(["--shards", str(options["shards"])])
+    import_args.extend(["--replicas", str(options["replicas"])])
```
Tryneus: It looks like this will be passing down `--shards None` and `--replicas None` if those parameters use their default value, which will error in the import script options parsing.

drivers/python/rethinkdb/_import.py
marshall007: Hmm, yea I was afraid that might happen, which is why I originally had the default set to `1`. I was hoping that passing `None` would just work the same as doing nothing so that it would work with #4608 as @danielmewes. Is there a better way without having to make different calls depending on the value of `shards/replicas`?

Tryneus: I would suggest making a `create_args` dict directly in `options` rather than having both `shards` and `replicas`, something like:

```py
def parse_options():
  ...
  res['create_args'] = dict([(name, options[name]) for name in ['shards', 'replicas'] if getattr(options, name) != None])
  ...

def create_table(..., create_args):
  ...
    r.db(db).table_create(table, **dict(create_args.items() + [(primary_key, pkey)])).run(conn)
  ...
```

This could be made more readable if you're willing to break it into multiple lines.

marshall007:
@Tryneus I think this is ready for another look when you have a second. I did pretty much exactly what you suggested, but also made `primary_key` part of the `create_args` since we have to call `r.table_create` in multiple places.

drivers/python/rethinkdb/_import.py
```diff
@@ -793,7 +803,9 @@ def import_directory(options):
 
     spawn_import_clients(options, files_info)
 
-def table_check(progress, conn, db, table, pkey, force):
+def table_check(progress, conn, db, table, create_args, force):
+    pkey = create_args["primary_key"]
```
Tryneus: This will crash if `primary_key` isn't present in `create_args`, which happens if it is not specified on the command-line.  Line 820 should probably just set `pkey` directly.  The two checks comparing `pkey` to `None`, should probably use `'primary_key' in create_args`.

Tryneus:
@marshall007 - I had one more comment in `_import.py`, and an unaddressed comment in `_restore.py`.  Please take a look and let me know your thoughts.

marshall007:
@Tryneus it was probably not obvious, but I changed the defaults in `_restore.py` to be `0` instead of `None` to address the issue you pointed out.

drivers/python/rethinkdb/_import.py
marshall007: Not sure I understood exactly what you meant concerning L820, but I just pushed another commit to address this.

Tryneus:
@marshall007, yep, I missed that change.  This looks like it'll do it, thanks for all the help!  Merging.

Tryneus:
Merged to `next`.  Also cherry-picked this into `v2.2.x` in commit b2fa7a7.

PR #5151, opened 2015-11-26
Use Integer.BYTES and Long.BYTES instead of hardcoding size

danielcompton:
Using Integer.BYTES and Long.BYTES makes the intent of the byte handling
code clearer, and less error prone if there are future changes to these
sections.


deontologician:
Looks good, thanks!

PR #5167, opened 2015-12-02
fix a java compiler error in eclipse.

bckfnn:
When loading the java driver in Eclipse IDE 4.5.1 gives an error
```
The type HashMap does not define putAll(Object, Object) that is applicable here	Converter.java	/drivers-java/src/main/java/com/rethinkdb/net	line 67
```

Specifying an explicit  type argument fixes the problem.


danielmewes:
Thanks for the PR @bckfnn !

Would you mind signing our CLA if you haven't done so yet? http://rethinkdb.com/community/cla/

Pinging @deontologician for review.

deontologician:
This looks good to me

bckfnn:
I have signed the CLA.

deontologician:
Merged, thanks again @bckfnn

PR #5221, opened 2015-12-17
Add information about Twisted requirement

katafrakt:
Apparently you also need Python's Twisted to run RQL tests (at least it shouted at me running Ruby tests when I did not have it installed). This might be a useful information to add to RQL tests Readme.

### DIFF
--- a/test/rql_test/README.md
+++ b/test/rql_test/README.md
@@ -25,11 +25,11 @@ make -C ../..
 sudo npm install -g mocha
 ```
 
-You also need `twisted` library for Python, with OpenSSL support. You might get it from your system's repositories or install it using `pip` (make sure you install it for correct Python version):
+You also need `twisted` library for Python, with OpenSSL support. Many systems have these installed, but if not they can be installed using `pip` (make sure you install it for correct Python version):
 
 ```
-pip install twisted
-pip install pyopenssl
+sudo pip install twisted
+sudo pip install pyopenssl
 ```
 
 ### Basic Usage
--- a/test/rql_test/README.md
+++ b/test/rql_test/README.md
@@ -25,6 +25,13 @@ make -C ../..
 sudo npm install -g mocha
 ```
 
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks for the PR @katafrakt, and sorry it took us a while to get back to you.

Have you signed our CLA yet? If not, you can do this online at http://rethinkdb.com/community/cla/ . We require the CLA for legal reasons before we can merge a pull request.

I believe the Twisted dependency is only for a certain set of tests. @larkost Is that right?
There are also some tests that require Tornado and gevent on Python, and EventMachine on Ruby. You might want to mention those too.

Assigning @larkost for further review.

larkost:
@katafrakt Thanks for pointing this out. All of my base test OS's have these modules automatically installed that I forget that they are not part of the base install of `python`. I am going to make a pair of line comments on the change, but they are minor. With those two, and a signed CLA, we are all ready to go!

test/rql_test/README.md
```diff
@@ -25,6 +25,13 @@ make -C ../..
 sudo npm install -g mocha
 ```
 
+You also need `twisted` library for Python, with OpenSSL support. You might get it from your system's repositories or install it using `pip` (make sure you install it for correct Python version):
+
+```
+pip install twisted
```
larkost: These should probably have `sudo` prepended since most people will not be running as root.

test/rql_test/README.md
```diff
@@ -25,6 +25,13 @@ make -C ../..
 sudo npm install -g mocha
 ```
 
+You also need `twisted` library for Python, with OpenSSL support. You might get it from your system's repositories or install it using `pip` (make sure you install it for correct Python version):
```
larkost: It sounds a little better to say:
```
many systems have these installed, but if not they can be installed using `pip` (make sure you install it for correct Python version):

katafrakt:
@larkost Thanks for your suggestions. PR is now updated. Should I also squash commits into one?

I already signed CLA some time ago.

larkost:
Looks great! I am merging it now. Thanks again for finding this, and contributing the fix!

PR #5261, opened 2016-01-06
Add Ruby variants to polyglot tests from #4846

### LINKED ISSUE #5251: test/rql_test/polyglot/aggregation.rb1.9 failing in next
Several instances of `Sorry, r.row is not available in the ruby driver.  Use blocks instead.`.

The other (dynamic) languages (Javascript, Python) pass just fine.

Seems to be introduced by #4846; pinging @deontologician and @marshall007

marshall007:
Fixes #5251.


larkost:
This all looks good, and solves the problem. Thanks!

PR #5284, opened 2016-01-13
Add Java driver SSL support

### LINKED ISSUE #5278: Java driver SSL support
I've looked into the code and the driver doesn't seem to support SSL. Can someone clarify on the current state of this?

I'm happy to contribute with a PR that adds SSL support but I'll for sure need to debate with someone. I'm on Slack #driver-dev.

pires:
Fixes #5278

Details on how to test will follow in the next comment.


pires:
In order to test this, I'm using [this code of mine](https://github.com/pires/rethinkdb-ssl-test) and a Compose.io cluster. It should be pretty straightforward to test with your own installment, both secure and unsecure.

It should be the responsibility of the user to provide `SSLContext` to the `Connection.Builder` exposed by `Rethinkdb.r.connection()`. This way, the driver is agnostic to the different SSL scenarios each user may provide.

drivers/java/src/test/java/com/rethinkdb/gen/TransformArray.java
```diff
@@ -2,43 +2,24 @@
...
+import com.rethinkdb.TestingFramework;
+import com.rethinkdb.gen.ast.MakeArray;
+import com.rethinkdb.gen.ast.Union;
 import com.rethinkdb.model.OptArgs;
 import com.rethinkdb.net.Connection;
-import com.rethinkdb.net.Cursor;
-import junit.framework.TestCase;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertArrayEquals;
+import org.junit.After;
```
deontologician: If these additions are made only in this file, it's going to get clobbered when the tests are regenerated from the template

drivers/java/templates/Test.java
```diff
@@ -33,8 +33,8 @@
 import java.util.Collections;
 import java.nio.charset.StandardCharsets;
 
-import static gen.TestingCommon.*;
-import gen.TestingFramework;
+import static com.rethinkdb.gen.TestingCommon.*;
+import com.rethinkdb.gen.TestingFramework;
```
deontologician: This is the template where the generated tests get their imports from. If there's a reduced set of imports that can be used, it should be modified here

drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
```diff
@@ -2,47 +2,86 @@
...
+            final InetSocketAddress addr = new InetSocketAddress(hostname, port);
+            socket = socketFactory.createSocket();
+            socket.connect(addr, timeout.orElse(0L).intValue());
+            socket.setTcpNoDelay(true);
+            socket.setKeepAlive(true);
+
+            // should we secure the connection?
+            if (sslContext.isPresent()){
+                socketFactory = sslContext.get().getSocketFactory();
+                SSLSocketFactory sslSf = (SSLSocketFactory) socketFactory;
```
deontologician: Why are these casts needed here?

drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
```diff
@@ -2,47 +2,86 @@
...
 
 public class SocketWrapper {
-    private SocketChannel socketChannel;
-    private Optional<Long> timeout = Optional.empty();
+    final ReentrantLock mu = new ReentrantLock();
+
+    private Socket socket = null;
+
+    private SocketFactory socketFactory = SocketFactory.getDefault();
+    private SSLSocket sslSocket = null;
```
deontologician: I'd rather use an optional than nullable fields

drivers/java/src/test/java/com/rethinkdb/gen/TransformArray.java
pires: I didn't know the tests were generated. My bad! Will fix.

drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
pires: `socketFactory` is declared as type `SocketFactory` and used in the code for non-SSL sockets.

drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
pires: I can do that but I tried to mimic the previous behavior where `SocketChannel` was nullable.

pires: Actually, we should leave it as is since `sslSocket` wraps `socket`. Besides `connect(...)` we only use `socket` on method `close()`. Hence, no need for null checking.

pires:
@deontologician thank you so much for your valuable feedback. My bad with the tests and will try and fix it ASAP.

Now, regarding the split of the `SSLConnection`, the thing is in Java an `SSLSocket` is actually a wrapper of `Socket`. Whatever operations you do in the underlying `Socket` affect the wrapping `SSLSocket`. So, breaking it apart it seems would only duplicate code.

danielmewes:
Small side question @pires: Have you already signed http://rethinkdb.com/community/cla/ ?

Also many thanks for working on this.

pires:
@danielmewes done.

deontologician:
Looks good from my end. @larkost @danielmewes do we have a compose account (or some other SSL proxied cluster) we can use to do a smoke test?

deontologician:
Alright, I ran the tests successfully on a compose cluster, and verified the test didn't work if I messed up any of the config (that's a nice way of saying I messed up the config a lot).

So, I was talking with @danielmewes about this, and we'd like to reduce the trouble needed to use a cert in the simple case (you have one cert file). Sometimes you'll want the flexibility to provide an SSLContext, but other times it would be nice to just pass an InputStream containing your cert file.

I found this SO answer that indicates you can create an in-memory KeyStore: http://stackoverflow.com/questions/18513792/using-sslcontext-with-just-a-ca-certificate-and-no-keystore . With that, we should be able to provide an option to the builder that does the boilerplate of creating a mini-keystore with the InputStream from the user, and generate the SSLContext from that.

So the concrete request is adding a `certFile` method to `Connection.Builder` that accepts an `InputStream` with the contents of the cacert file, and which builds an SSLContext that it passes to the Connection. The builder should error if the user passes both an SSLContext and a direct InputStream

pires:
Thank you @deontologician and @danielmewes.

I will be adding the requested feature.

pires:
@deontologician @danielmewes done. Instructions updated as well at https://github.com/pires/rethinkdb-ssl-test.

drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
```diff
@@ -2,47 +2,125 @@
...
+            socket.connect(addr, timeout.orElse(0L).intValue());
+            socket.setTcpNoDelay(true);
+            socket.setKeepAlive(true);
+
+            // should we secure the connection?
+            if (certFile.isPresent() || sslContext.isPresent()) {
+                // is certFile provided? if so, it has precedence over SSLContext
+                if (certFile.isPresent()) {
+                    try {
+                        final CertificateFactory cf = CertificateFactory.getInstance("X.509");
```
deontologician: Could you move this part into the builder, so just the SSLContext is passed to the connection constructor?

pires: Makes sense! Will do.

pires: @deontologician done.

deontologician:
Alright, I had another look over this, and I think it's good. Merging, and thanks again @pires, this is really awesome.

pires:
@deontologician my pleasure! So glad it got merged :)

PR #5292, opened 2016-01-15
Connection is now thread-safe.

### LINKED ISSUE #5166: Make Java connections thread-safe
Compare https://github.com/bchavez/RethinkDb.Driver/issues/15 for the C# driver.

I imagine we just need to add a few `synchronized` keywords here and there...

pires:
Fixes #5166


bchavez:
***Disclaimer***: I haven't written Java in years but I'll use my experience from porting the Java driver to .NET driver and some issues I ran into that *may* be worth checking out. Also, I have practically zero knowledge of how Java VM schedules threads. So, on this bit of review, I'm assuming a conceptual threading model similar to .NET threading and how Microsoft Window schedules threads. I apologize in advance if none of this applies to Java.

Assuming we have two `Cursors` each "owned" by a Java thread, name these: **Thread A / Cursor 1** and **Thread B / Cursor 2**. Each thread consumes their initial batch of results. Next, each thread fires off a `Cursor` `CONTINUE`. The server takes a some time processing the continues and both threads enter the `while(ture){...}` loop awaiting their response below:

#### `ConnectionInstance.java`
screen560


**Line 69**: The server hasn't responded yet, so both threads don't see any `headerInProgress`. Both enter `sock.recvall(12,deadline)` with the expectation of reading **12 bytes**.

#### `SocketWrapper.java`
screen564

Both threads `mu.Lock()` but only one lucky thread a ticket to pass though the turn-style. For the sake of simplicity, let's keep things in order, say **Thread A** is the lucky one. **Thread A** acquires the lock, reads **12 bytes** from the socket successfully. Which response is it? Not sure, but I don't think it matters yet.

**Thread A** is done reading the **12 byte** header, `finally{ mu.unlock() }` on **Line 136** and returns.

**Thread A** sets `headerInProgress` in `ConnectionInstance.java` when it returns on **Line 70**. **Thread A** reads `resToken`, but wait, **Thread A**'s quantum runs out. **Thread B** is now scheduled to run. **Thread B** is free to enter the `mu.Lock()` since it was released by **Thread A**. **Thread B** now reads **12 bytes** *into* the response *json body* of **Thread A**'s response. Hmmm. Did **Thread B** just consume **12 bytes** the **json** response body thinking it was a header of `resToken+resLen`? Is part of **Thread A**'s response *json body* partly in the buffer of **Thread B**?

In the .NET world this is a real scenario that could happen. This issue was just one of the issues I ran into in the .NET driver with the original architecture. In the end, I decided to deviate slightly in this area from the original Java architecture because threading got way to complex and out of hand. A small architecture change worked out a lot better for the .NET driver.

Also, I could be completely wrong here maybe this doesn't happen in the Java VM.

danielmewes:
Thanks for another great PR @pires .

@bchavez I believe your analysis is correct. I think the lock needs to be acquired and held in `readResponse` and span both `recvall` calls, rather than being acquired inside of `recvall`.
There might also be other ways for addressing this of course. In fact I think there's no reason for why we would need to have two calls of `readResponse` at the same time. So we could maybe just make `readResponse` `synchronized`?

danielmewes:
Something else we might need to be careful about: The way I read `readResponse`, it only returns once it has received a response the the given token.
Say *cursor A*'s `CONTINUE` request calls `readResponse` first, followed by *cursor B*.
Assume that the server first replies to *cursor B*'s request though. The `readResponse` call of *cursor A* is going to get that response and call `extend()` on *cursor B*. Later it will receive the response for *cursor A* that it was waiting for and return.
However *cursor B* is still stuck in the `readResponse` call. It will never return, since *cursor A*'s call to `readResponse` had already consumed the response for *cursor B* from the connection.

I haven't checked where exactly `readResponse` is called, and this might not actually be an issue.

bchavez:
I used a response push-based model instead of a response pull-based one. In .NET, threads aren't polling for their responses (which put me into a lot of concurrency issues). I don't know if Java has anything push-based for threads that could lend itself as an alternative solution. A lot of these problems went away when I switched to a response push-based model in the .NET driver.

pires:
>Basically, I have a dedicated thread responsible for picking up messages off the wire and delivering them to awaiting thread consumers.

@bchavez how do you correlate messages to consumers?

Other than that, I agree with your thoughts. I was just lazy and tapped myself in the back when I saw the concurrent tests working. But who watches the watchdog, right? Probably the tests are broken too.

bchavez:
The reference implementation for .NET on the topic is here:

https://github.com/bchavez/RethinkDb.Driver/blob/master/Source/RethinkDb.Driver/Net/SocketWrapper.cs

I have an `awatiers` dictionary/map in `SocketWrapper`:
```
private ConcurrentDictionary<long, Awaiter> awaiters = new ConcurrentDictionary<long, Awaiter>();
```
This correlates the incoming responses with `awaiters` (consumer threads) based on the query token that are expecting a response. It looks like Java has something similar [`ConcurrentMap`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/package-summary.html) that looks like it does the same thing. Key: `long` is the QueryToken, and Value: `Awaiter` is a consumer thread awaiting a response. 

Next, what is an `Awaiter`?
````
public class Awaiter : TaskCompletionSource<Response>
```
in .NET, it represents a future value of an asynchronous operation. In Java, it kinda looks like [`FutureTask<V>`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/FutureTask.html). Honestly, I don't know enough about Java concurrency to give you any real guidance on how to use `FutureTask` or best practice guidance on using these concurrent constructs. I'm sure others will have more XP in this area about how to use them.

When incident threads write to the network stream (in .NET `SendQuery`), an awaiter is returned to the caller if one is requested and the thread is expecting a response. The thread immediately returns and awaits the future value (something like this in Java: [`FutureValue.get()`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/FutureTask.html#get()) )

> Description copied from interface: Future: Waits if necessary for the computation to complete, and then retrieves its result.

When a message is received on the socket, the message pump thread reads a response message. Checks the `awaiters` dictionary map, correlates the response token with the `FutureValue`, then sets the value.

*I think* then, in Java, the thread awaiting the response will finally return from `FutureTask.get()`. At least, in theory anyway. :smiley: 

Once you have this mechanism working. The calling semantics change and inversion takes place. Don't call us, we'll call  you. This lets you remove `readResponse` in `ConnectionInstance.java` and avoids the issue @danielmewes mentioned.

Also, you'll want to make sure `cursorCache` is a `ConcurrentMap`. Otherwise, multiple threads adding themselves to the cache can cause concurrency issues in `HashMap` (or at least they do in .NET) when the dictionary is re-sized internally.

Hope that helps. -Brian

pires:
@bchavez thank you for the detailed explanation. The design makes perfect sense.

Now, given this design, it should be really easy to provide a connection pool don't you think?

Also, how does this work for changefeeds?

bchavez:
@pires , no problem m8. I hope the design info helps. BTW, kudos for improving the Java driver. It's a significant contribution for the community.

##### Regarding Changefeeds in The New Design
Changefeeds use an underlying `Cursor` type. Not much changes in `Cursor` code. At a high level, the client driver is controlling the flow of downstream updates it receives from a changefeed/cursor by sending `CONTINUE` over the wire. When a consumer thread is iterating over a `Cursor` internally the code is giving back values from a buffered list of items. When the buffer of items runs out, the consumer thread inside `getNext()` sends `CONTINUE` over the wire to the server. A new batch arrives and the `Cursor` is extended with the new list of buffered items. In this design, when using `FutureValue<Response>`, it is important to keep track of the currently pending `FutureValue<Response>` `CONTINUE` so it can be awaited on in `getNext()`. When the `CONTINUE` request is fulfilled, the consumer thread can "extend" its own buffer list of items.

`MaybeSendContinue()` has a check to ensure only one `Cursor` `CONTINUE` (per `Cursor` instance) is pending over the wire at any given time. So, I just set the reference to the `awatingContinue`.

```
protected internal virtual void MaybeSendContinue()
{
   if( error == null && items.Count < threshold && outstandingRequests == 0 )
   {
      outstandingRequests += 1;
      awaitingContinue = connection.Continue(this);
   }
}
```

The relevant .NET reference for the `Cursor` implementation is here:
https://github.com/bchavez/RethinkDb.Driver/blob/master/Source/RethinkDb.Driver/Net/Cursor.cs

I didn't do much to make a `Cursor` by itself intrinsically thread-safe. My assumption in the `Cursor` implementation still assumes one consumer thread iterating over a `Cursor` instance. Conceptually, I didn't think it made much sense to have multiple threads iterating over a single `Cursor` instance. Maybe it could change in the future if someone comes up with a *really* good use case. But even then, I would be reluctant to make `Cursor` thread-safe. Currently, if a developer needed multiple threads on a single `Cursor` instance I'd recommend the developer use a Mutex over the `Cursor`. I would hope it would encourage the developer to write code in a way that was more explicitly thread-safe on the eventual consumption of the items being iterated over.

Also, there was a proposal somewhere that I encountered that discussed the possibility of pushing changes to client drivers without an explicit `CONTINUE` but I don't think there are any immediate plans for this. If the RethinkDB team wanted to do this in the future then the `ResponsePump` would simply do the cursor extending itself by reaching into the `cursorCache` instead of pushing future values to awaiters. But this was just *proposal* and the design (I think) would still work out.


##### Regarding Connection Pooling
You'd probably need to talk with the RethinkDB team on this one. I think the API would probably need to go though some proposal process before going public. It would be the first official driver to support connection pooling?

Once you have multi-threading working perfectly connection pooling is pretty straight-forward after that.

For the .NET driver, one constraint I had for the .NET driver was keeping the code base similarly aligned to the Java driver. I did not want the .NET driver implementation to change too much other than using .NET semantics. So, this constraint somewhat impacted my design of the Connection Pooling code in the .NET driver.

The design notes for .NET Connection Pooling are here:
https://github.com/bchavez/RethinkDb.Driver/issues/17

And associated documentation for .NET Connection Pooling:
https://github.com/bchavez/RethinkDb.Driver/wiki/Connections-&-Pooling

The implementation is here:
https://github.com/bchavez/RethinkDb.Driver/tree/master/Source/RethinkDb.Driver/Net/Clustering

In the design note, I noted it would have been nice to eliminate `ConnectionInstance` or move `SocketWrapper` into `ConnectionInstance`, so that we could keep the # of layers and concepts of a 'connection' to a minimum of two. `Connection -> ConnecitonInstance` or `Connection -> SocketWrapper`. Currently, we have a conceptual mental model of something like `Connection -> ConnectionInstance -> SocketWrapper`. I kinda feel there is code smell here. Maybe it could be a possibility of merging `ConnectionIntance` and `SocketWrapper` now that there's a possibility of `readResponse` being removed from `ConnectionInstance`. If they are merged, I could make improvements on the Connection Pool API and have a cleaner / more robust implementation. It would simplify a connection "ErrorCallback" mechanism for connection pooling.

I used neumino's rethinkdbdash API and GoRethink API as references. All very good. In particular, GoRethink used [go-hostpool](https://github.com/hailocab/go-hostpool) which I used as a starting point to implement the basis of the connection pooling code. The 1-to-1 C# port of go-hostpool had a lot of locking in Go. As a result, the implementation suffered from performance problems due to all the locking going on inside the host pool. I say "performance problems" because I unit tested 600K concurrent multi-threaded node selections as a baseline and it took 4 seconds to complete. I wanted to squeeze every ounce of performance out of node selection so 98% of CPU time was spent doing actual driver work not node selection. I wanted node selection be virtually free in multi-threaded scenarios.

Having that goal in mind, I refactored the go-hostpool in C# multiple times. Eventually what came out was:

1. Use monotonically increasing array storage for nodes (IE: never remove nodes if they go offline only mark them as dead). This allows me to never get out of bounds in tight for loops from multiple threads. Updates to the node list are only Adds(), no need for locking the host pool and stopping all threads for this "newly added" node. The new node will eventually be selected by a thread.
2. Removed virtually all locks on variables in favor of `Interlocked`/.NET OS/CPU primitives to prevent the CPU from pipelining/reordering instructions when updating memory from multiple threads lock-free.

Now node selection runs in like 100ms instead of 4 full seconds.

A common interface `IConnection` was found between `Connection` and `ConnectionPool`, and the rest is pretty basic. A discover thread is in charge of discovering new nodes added (sets up a change feed on a system table) and adds them to the node pool. A supervisory thread that periodically scans through the node list and tries to bring up failed connections.

Hope that helps,
Brian

pires:
First of all, a big shout at @bchavez for his amazing mentoring on this issue. While the current state of this PR is not as everyone wished for - it doesn't work! -, I wouldn't have gotten this far without his guidance. Thank you!

Now, I have screwed up with `Cursor` and it seems I'm hitting a wall here. I really need your help!

I understand if you guys don't like my latest commit since it breaks with the previous design but I couldn't maintain the old design with all the behavior changes going on. I've tried to my best effort to mimic the .NET driver, though. Unfortunately, that wasn't enough to get this working.

I am more than happy to give contributor access to my branch to anyone trying to help, but I have no problems with someone opening a working PR that satisfies everybody and see this closed. I need this feature.

bchavez:
Hi @pires great work regardless the state of the PR. I think you're very close. Give me a few hours to spike **Java**'s threading model and I'll let you know.

bchavez:
So, after some additional review, I have a strong feeling the `ResponsePump` and `awaiters` stuff should really go into `SocketWrapper` like the .NET driver. I think there's too much state to "remember to clear" in `Connection` class if a connection is restarted... and it will be harder to get right as `Connection` code gets bigger in the future ... if `ResponsePump` and `awatiers` get encapsulated in `SocketWrapper` generally these problems go away like in the .NET driver when a new socket is created on reconnect.... Currently, the Java `Connection` a lil bit more complicated to get right the way it is setup now....  For example: how `awaiters` is being cleared at the `Connection.close()` level ... if we move `ResponsePump` and it's `awaiters` state in `SocketWrapper` these network "state" troubles go away because it's logically separated from the main connection class.

:heavy_check_mark: `Connection` is responsible for closing high-level `Cursors`. Then...
:heavy_check_mark: `SocketWrapper:ResponsePump` is responsible for cleaning up it's own awaiters.

Let me know what you think.

pires:
@bchavez it makes sense and is definitely something we should do. I just wanted to prove the behavior works before working the final PR.

Re: rename `SocketWrapper`, again I agree but I know naming is a hard thing to do, and we need input from @deontologician or @danielmewes on this.

Right now, I need to catch up some late work but will come back to this ASAP.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -125,75 +124,150 @@ public Connection reconnect(boolean noreplyWait, Optional<Long> timeout) throws
         if (!timeout.isPresent()) {
             timeout = connectTimeout;
         }
-        close(noreplyWait);
-        ConnectionInstance inst = instanceMaker.get();
-        instance = Optional.of(inst);
-        inst.connect(hostname, port, handshake, sslContext, timeout);
+        close();
```
deontologician: Shouldn't noreplyWait still be used?

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -125,75 +124,150 @@ public Connection reconnect(boolean noreplyWait, Optional<Long> timeout) throws
...
+                if (!isOpen()) {
+                    log.error("Response Pump: The socket is not open, exiting.");
+                    break;
+                }
+
+                // read response and send it to whoever is waiting, if anyone
+                try {
+                    final Response response = this.socket.orElseThrow(() -> new ReqlDriverError("No socket available.")).read();
+                    final CompletableFuture<Response> awaiter = awaiters.remove(response.token);
+                    if (awaiter != null) { // yes
```
deontologician: Stray `//yes` comment?

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -125,75 +124,150 @@ public Connection reconnect(boolean noreplyWait, Optional<Long> timeout) throws
...
-            try {
-                if (shouldNoreplyWait) {
-                    noreplyWait();
-                }
-            } finally {
-                instance = Optional.empty();
-                nextToken.set(0);
-                inst.close();
+    public void close(boolean mayInterrupt) {
+        // reset token
```
deontologician: it looks like the noreplyWait logic has been removed?

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
pires: Yes.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
pires: Yes.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
pires: Will fix.

pires:
@deontologician addressed your comments, squashed commits, rebased latest `next`, ran tests against `2.2.1` and `2.2.4` and it seems good to merge.

danielmewes:
:+1: :clap: Waiting for @deontologician to make a final check, and then we should be good to go.

Great work on this significant contribution @pires and @bchavez !

bchavez:
:+1: great job! Just don't forget:

* https://github.com/rethinkdb/rethinkdb/issues/5380
* https://github.com/pires/rethinkdb/pull/2

pires:
@bchavez I can't stress enough how important your help was, thanks man. #5380 will be a subsequent task since I can't do it now.

pires:
Rebased to include https://github.com/rethinkdb/rethinkdb/pull/5409. /cc @deontologician @danielmewes

deontologician:
OK, testing this now

deontologician:
Ok, this is now 2.2-beta-5 on sontatype. Thanks again @pires and @bchavez !

danielmewes:
:clap: :-)

pires:
Awesome! Time to celebrate. Beers for everyone.

PR #5354, opened 2016-02-03
Adding __aiter__ and __anext__ to Python AsyncioCursor

dalanmiller:
Looking at [PEP 0492](https://www.python.org/dev/peps/pep-0492/#asynchronous-iterators-and-async-for) I'm adding `__aiter__` and `__anext__` to support the following syntax in Python 3.5. 

Talked with @Tryneus about the timeout and felt that an infinite timeout would be the best default for both changefeeds and normal queries. Users can use `._get_next(...)` if they want to explicitly set a timeout. 

``` python
conn = await r.connect("localhost", 28015)
changes = await r.db("python").table("python").changes()["new_val"].run(conn)

# The changes allow for this async for syntax.
async for change in changes:
    # Work with change
```

Also related to rethinkdb/docs#949.


drivers/python/rethinkdb/asyncio_net/net_asyncio.py
```diff
@@ -71,6 +71,18 @@ def __init__(self, *args, **kwargs):
         Cursor.__init__(self, *args, **kwargs)
         self.new_response = asyncio.Future()
 
+    @asyncio.coroutine
+    def __aiter__(self):
+        return self;
```
adrianliaw: Semi-colon!

dalanmiller: JS has finally truly entrenched itself in my brain. Thanks @adrianliaw.

dalanmiller:
I added a test file with the suffix `_3_5.py3.5` to differentiate from the original `asyncio_connection.py` since these are the tests using syntax found only in Python 3.5. 

All `@asyncio.routine` decorators have been replaced with `async def ....` and removing some deprecated `asyncio` methods in 3.5. 

There's a lot of places where the tests use `cursor.next` where we could now use `async for ... in ...` to iterate automagically over an awaitable iterable but I wasn't super acquainted with the tests so I left it at this.

test/rql_test/connections/asyncio_connection_3_5.py
```diff
@@ -0,0 +1,865 @@
...
+        finally:
+            useSocket.close()
+
+    async def test_empty_run(self):
+        # Test the error message when we pass nothing to run and
+        # didn't call `repl`
+        with self.assertRaisesRegexp(r.ReqlDriverError,
+                                "RqlQuery.run must be given"
+                                " a connection to run on."):
+             r.expr(1).run()
```
Tryneus: Is this missing an `await`?

test/rql_test/connections/asyncio_connection_3_5.py3_5.test
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+
+import os, sys, subprocess
+
+try:
+    import asyncio
+except ImportError:
+    # remove this when implimented https://github.com/rethinkdb/rethinkdb/issues/4139
```
Tryneus: typo: implimented => implemented (probably in the file you copypastad, as well =/)

dalanmiller:
I think before this is merged, it should also include `__aenter__` and `__aexit__` for `r.connect` which allow you to write something like:

```python

# Context manager able to suspend execution 
#  during the opening and closing of the connection
async with r.connect('localhost', 28015) as conn:
    res = await r.db("test").table("test").run(conn)

# conn closed automatically 
```

test/rql_test/connections/asyncio_connection_3_5.py3_5.test
```diff
@@ -0,0 +1,12 @@
...
+    sys.stderr.write('This test requires the asyncio module from Python 3.5')
+    sys.exit(0)
+
+sys.exit(subprocess.call([os.environ.get('INTERPRETER_PATH', 'python'), os.path.join(os.path.dirname(__file__), 'asyncio_connection.py')]))
```
Tryneus: It looks like this is launching the other `asyncio` tests, not the file you've added in this PR.

Tryneus:
I think at least one of the tests should use the `async for ... in ...` syntax, since that is mostly what this PR adds.  It would probably be easiest in the `test_cursor_success` function.

dalanmiller:
@Tryneus I'm (re)assigning this to you. I also went and added an awaitable `_stop` in `net_asyncio.py` that I think is also necessary.

danielmewes:
Unfortunately this won't make it into the initial 2.3 driver. We'll try to get it into a point release update afterwards.

larkost:
The `connections/asyncio_connection_3_5.py3.5` setup was pretty broken:
1. It assumed that `test/rql_test/connections/asyncio_connection_3_5.py3_5.test` would be run with Python 3.5. Rather it is just run as an executable and the shebang line made that the system default Python (for most systems: 2.7).
2. That same file then checked if it was running Python 3.5 or greater, then used `sys.exit(0)`. This signals the test system that it passed.
3. In the case of the wrong version of Python an `AssertionError` would have been thrown, and not caught by the `try` block that was set out for it since that latter was looking for an `ImportError`.
4. The line that ran the actual test file then chose `python3` rather than `python3.5` as the default.

With 6651ae74 I have moved the version check into the actual test,and fixed it to run on Python 3.5. I also did some other generic cleanup on stuff that I noticed (e.g.: unneeded imports). The test is, of course, still failing.

ultrabug:
Awesome :+1: please

danielmewes:
With the recent changes to the test system, we should now just have a single `test/rql_test/connections/asyncio_connection_3_5.py3_5.test` file that contains the actual test code right?

@dalanmiller Apart from getting the tests work properly, is there anything else missing here that you're aware of?

larkost:
After thinking about this a little, I think that the best approach would be to add the changes for this into the normal `asyncio_connection.py3_4+.test` file, but put the testing for `__aiter__` and `__anext__` behind skipping decorators that check for the version. This would probably work (not checked):
```python
def newInPython3_5(obj, attr):
    if sys.version_info >= (3,5):
        return lambda func: func
    else:
        return unittest.skip('%r requires Python3.5 or above')
```

dalanmiller:
> After thinking about this a little, I think that the best approach would be to add the changes for this into the normal asyncio_connection.py3_4+.test file, but put the testing for __aiter__ and __anext__ behind skipping decorators that check for the version.

I'm not sure this will work since 3.5 introduced the `async` keyword. Also there's some difference between `asyncio.async` which turned into `asyncio.ensure_future` in 3.5. So the following examples will fail: 

```py
async for change in changes():
...
async with r.connect() as conn:
...
async def function():
```

dalanmiller:
@danielmewes I recalled incorrectly that maybe `async with` still needed some dunder methods but looks like it just needed `__aenter__` and `__aexit__` so we should be good afaik.

ultrabug:
@dalanmiller any chance you could update this PR's code ?

Your experience on implementing this would be greatly appreciated (I must admit I'm at loss) so we can at least maintain a fork and use it while this PR is merged and released.

dalanmiller:
@ultrabug Do you mean resolve the current conflicts with `next`?

ultrabug:
@dalanmiller that's right so that users in need of these great features can make a patch and apply it to their local lib.

I must say I tried and failed miserably maybe because I couldn't figure out how to fix those relative imports on my development virtualenv in my limited time...

Thanks in advance, I really hope this will be merged soon.

wwoods:
Another issue with this: `__aiter__` should return the iterator itself, not an awaitable, as per PEP-0492 (https://www.python.org/dev/peps/pep-0492/#asynchronous-iterators-and-async-for).  This was changed in Python 3.5.2 and will be relevant going forward.

dalanmiller:
Hey @wwoods from what I understand, is this sufficient to fix it for Python 3.5.2? I've just removed the `@asyncio.coroutine` decorator from `__aiter__` as shown in the [0492 async iterators example](https://www.python.org/dev/peps/pep-0492/#asynchronous-iterators-and-async-for) you linked and added `await` to the `self._get_next(None)` line in `anext`.

> PEP 492 was accepted in CPython 3.5.0 with __aiter__ defined as a method, that was expected to return an awaitable resolving to an asynchronous iterator.

> In 3.5.2 (as PEP 492 was accepted on a provisional basis) the __aiter__ protocol was updated to return asynchronous iterators directly.

Then the requirements listed here - https://www.python.org/dev/peps/pep-0492/#asynchronous-iterators-and-async-for - I think are also met: 

* `AsyncioCursor.__aiter__` returns itself directly, not an awaitable
* `AsyncioCursor` has a `__anext__` async function which returns an awaitable
* `AsyncioCursor.__anext__` returns `asyncio.StopAsyncIteration` upon falsyness of `data`

```py
def __aiter__(self):
    return self
```

```py
@asyncio.coroutine
def __anext__(self):
    data = await self._get_next(None)
    if data:
        return data
    else:
        raise asyncio.StopAsyncIteration
```


```py
@asyncio.coroutine
    def _get_next(self, timeout):
        waiter = reusable_waiter(self.conn._io_loop, timeout)
        while len(self.items) == 0:
            self._maybe_fetch_batch()
            if self.error is not None:
                raise self.error
            with translate_timeout_errors():
                yield from waiter(asyncio.shield(self.new_response))
        return self.items.popleft()
```

p.s. 
@ultrabug this branch is now merge-able again.

drivers/python/rethinkdb/asyncio_net/net_asyncio.py
```diff
@@ -71,6 +71,26 @@ def __init__(self, *args, **kwargs):
         Cursor.__init__(self, *args, **kwargs)
         self.new_response = asyncio.Future()
 
+    @asyncio.coroutine
+    def __aiter__(self):
+        return self
+
+    @asyncio.coroutine
+    def __anext__(self):
+        data = self._get_next(None)
```
ultrabug: I think this should be `data = yield from self._get_next(None)` since self._get_next() returns a coroutine and the `if data` below would always evaluate as True.

drivers/python/rethinkdb/asyncio_net/net_asyncio.py
```diff
@@ -71,6 +71,26 @@ def __init__(self, *args, **kwargs):
...
+        if data:
+            return data
+        else:
+            raise asyncio.StopAsyncIteration
```
ultrabug: There's no `asyncio.StopAsyncIteration`, this should be `StopAsyncIteration`

ultrabug:
@dalanmiller great news, thank you ! I finally got the hold on how to run the lib in develop mode in my virtualenv so I could give it a try and made a few comments as you can see.

Also, I think the `_get_next()` method is wrongly raising an RqlCursorEmpty exception on line 123. My current fix is checking the error instance type like this:

```python
    @asyncio.coroutine
    def _get_next(self, timeout):
        waiter = reusable_waiter(self.conn._io_loop, timeout)
        while len(self.items) == 0:
            self._maybe_fetch_batch()
            if self.error is not None:
                if isinstance(self.error, RqlCursorEmpty):
                    return
                else:
                    raise self.error
            with translate_timeout_errors():
                yield from waiter(asyncio.shield(self.new_response))
        return self.items.popleft()
```

Hope this helps

wwoods:
@dalanmiller Yeah I think that's all that changed.

dalanmiller:
@danielmewes This is ready for review. 

It would be nice to have `async with r.connect() as conn:` but it would involve modifications to `net_asyncio.Connection.reconnect` to not be a coroutine. So in order for it to work at the moment, it needs to be:

```py
async with await r.connect() as conn:
    pass
```

The problem being that we want to manage the context of the connection itself not the connection function. So `__aexit__` and `__aenter__` [are defined there](https://github.com/rethinkdb/rethinkdb/pull/5354/files#diff-5609ca78d50319b2e4e1263ade679e99R304). 

A good example of this properly done is [aiohttp](http://aiohttp.readthedocs.io/en/stable/):

```py
with aiohttp.ClientSession() as session:
    async with session.get('https://api.github.com/events') as resp:
        print(resp.status)
        print(await resp.text())
``` 

Where [here it's defined](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client.py#L94-L129) that there's a standard non-coroutine `def request` which synchronously returns a class implementing `__aexit__` and `__aenter__` which wraps the coroutine request (and sends . 

I say for the moment we ship this and put fixing the context manager properly at a later time and also spend some time to write some more comments.

Alternatively, we can also ship everything except the `async with` functionality in a half-baked state.

Thanks to @sseg for some pointers on this.

danielmewes:
@Tryneus could you review the updated code please?

danielmewes:
@dalanmiller This now has a merge conflict. I suspect due to changes to the asyncio test in `next` that @larkost made. Do you have time to update the branch?

AtnNn:
Merged in #6291

ultrabug:
@AtnNn awesome news, thanks a lot !

PR #5381, opened 2016-02-10
TLS for all connections

### LINKED ISSUE #3151: Simplify setup of encrypted RethinkDB clusters
Right now users have to manually set up SSL tunnels / VPN / SSH tunnels to securely run a RethinkDB cluster on an open network (see e.g. the guide at http://rethinkdb.com/docs/security/).

It would be great if we could make that easier, e.g. by providing some scripts to partially automate the setup process of an encrypted RethinkDB cluster.

I imagine the user experience to work somewhat like this:
- Generate a key (-pair)
- Copy the key to all RethinkDB servers
- Provide the key to the RethinkDB instances through some configuration or command line option
- Have some magic that sets up for example SSH tunnels as needed to get the cluster up

I'm not sure how practical this is, given that it would have to work in a lot of different environments, different versions of OpenSSH etc.
I think it's worth investigating though, as it would significantly improve the user experience for many people.

Maybe we can also do something to make encrypted client access easier to setup? (*Edit:* This is now tracked in #3158)

### LINKED ISSUE #3158: Consider allowing drivers to use encrypted connections

jlhawn:
This set of commits adds the ability to wrap any of the web, driver, or cluster connections with TLS. The `--help` output has been accompanied with the following section:

```
TLS options:
  --web-tls                                   secure the web administration console
                                              with TLS
  --web-tls-key key_filename                  private key to use for web
                                              administration console TLS
  --web-tls-cert cert_filename                certificate to use for web
                                              administration console TLS
  --driver-tls                                secure client driver connections with
                                              TLS
  --driver-tls-key key_filename               private key to use for client driver
                                              connection TLS
  --driver-tls-cert cert_filename             certificate to use for client driver
                                              connection TLS
  --driver-tls-ca ca_filename                 CA certificate bundle used to verify
                                              client certificates; TLS client
                                              authentication disabled if omitted
  --cluster-tls                               secure intra-cluster connections with
                                              TLS
  --cluster-tls-key key_filename              private key to use for intra-cluster
                                              connection TLS
  --cluster-tls-cert cert_filename            certificate to use for intra-cluster
                                              connection TLS
  --cluster-tls-ca ca_filename                CA certificate bundle used to verify
                                              cluster peer certificates
  --tls-ciphers cipher_list                   specify a list of TLS ciphers to use;
                                              default is 'ECDHE+AESGCM'
  --tls-ec-curves curve_list                  specify a list of elliptic curves to
                                              use for ECDHE; default is
                                              'secp521r1:secp384r1:prime256v1'
  --tls-dhparams dhparams_filename            provide parameters for DHE key
                                              agreement; REQUIRED if using DHE
                                              cipher suites; at least 2048-bit
                                              recommended
```

TODO: rebase onto the `next` branch?

fixes #3151
fixes #3158


src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +649,163 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+        return false; 
+    }
+
+    // Mutual authentication.
+    SSL_CTX_set_verify(cluster_tls, SSL_VERIFY_PEER|SSL_VERIFY_FAIL_IF_NO_PEER_CERT, NULL);
+
+    return true;
+}
+
+bool initialize_tls_ctx(SSL_CTX **tls_ctx) {
```
jlhawn: This section needs to be audited.

danielmewes: @VeXocide can you have a look at this section?

src/arch/io/network.cc
```diff
@@ -621,6 +621,424 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+    logNTC("linux_secure_tcp_conn_t destructor\n");
+
+    if (is_open()) shutdown();
+}
+
+/* Creates a new OpenSSL connection using the underlying fd and the given
+OpenSSL context. The given handshake function is used to perform the handshake
+for the client or server using SSL_connect or SSL_accept respectively. */
+void linux_secure_tcp_conn_t::establish_conn(SSL_CTX *tls_ctx, int (*handshake)(SSL *) ) THROWS_ONLY(connect_failed_exc_t) {
+    conn = SSL_new(tls_ctx);
```
VeXocide: This is not exception safe in that an exception will cause us to leak the allocated memory. We'll probably need to create an RAII wrapper for this.

src/clustering/administration/main/command_line.cc
VeXocide: I'll have a closer look tomorrow, this'll require some coffee.

src/arch/io/network.cc
jlhawn: I've moved this allocation into the constructors. Since only one raises an exception, I've added a try/catch block which frees the allocated memory before re-throwing the exception.

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +649,163 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+    *tls_ctx = SSL_CTX_new(SSLv23_method());
+    if (NULL == *tls_ctx) {
+        ERR_print_errors_fp(stderr);
+        return false;
+    }
+
+    // Only allow TLS v1.2 and prefer server ciphers.
+    SSL_CTX_set_options(*tls_ctx, SSL_OP_NO_SSLv2|SSL_OP_NO_SSLv3|SSL_OP_NO_TLSv1|SSL_OP_NO_TLSv1_1|SSL_OP_CIPHER_SERVER_PREFERENCE);
```
VeXocide: This set of options looks great to me, and I agree forcing TLS v1.2 is the way to go.

One small thing to note is that `SSL_OP_NO_SSLv2` and related are deprecated in OpenSSL 1.1.0 and above, there we should use `SSL_CTX_set_min_proto_version(SSL_CTX *ctx, int version)`.

jlhawn: Has OpenSSL 1.1.0 been released? I was developing using the 1.0.2 API and that function doesn't seem to exist. I do see it in the docs for master though.

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +649,163 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+
+    /* This is pretty important. We want to use the most secure TLS cipher
+    suite that we can. This simple list only allows ciphers suites which employ
+    ECDHE (Elliptic Curve Diffie-Hellman with Ephemeral keys) for encryption
+    key agreement for Perfect Forward Secrecy and we want to use AESGCM for
+    efficient authenticated encryption. */
+    if (0 == SSL_CTX_set_cipher_list(*tls_ctx, "ECDHE+AESGCM")) {
```
VeXocide: First off "EECDH+AESGCM" is a great default. I was worried that older versions of OpenSSL such as those on CentOS and a few other distributions we build packages for wouldn't support it. Hence the temptation to suggest "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH" or similar.

Introducing EDH ciphers comes with more trouble though as by default OpenSSL will pick a 1024-bit key for these exchanges, I wish I knew why. We'd have to take care of this by providing a stronger EDH parameter which is a hassle.

In the end it turns out that the OpenSSL version provided with CentOS doesn't even support TLSv1.2 yet, thus on those distributions we can default to building and linking our own version. This also means we can stick to "EECDH+AESGCM", which is great.

It might be nice to allow this to be configurable though, with the "EECDH+AESGCM" default. If a problem were to pop up in the future it could then be circumvented by changing the configuration instead of having to rely on us to provide a new package, or recompiling it yourself.

src/clustering/administration/main/command_line.cc
VeXocide: You're right, it has not, I just noticed this in the documentation for master and jotted it down.

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +649,163 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+    speculate that these curves are compromised in some way by the NSA but most
+    people agree that they are safe to use. Any real risk in using elliptic
+    curves is associated with ECDSA (Elliptic Curve Digital Signature
+    Algorithm) where you must be careful to generate signatures correctly using
+    a secure PRNG. As we are using them only for ECDHE, I stand by this
+    selection. Feel free to add any more that we want - for example 'secp256k1'
+    (widely used and popularized by the Bitcoin cryptocurrency) and Daniel J.
+    Bernstein's 'Curve25519' (though not currently supported by OpenSSL) would
+    be other good choices. */
+    if (0 == SSL_CTX_set1_curves_list(*tls_ctx, "secp521r1:secp384r1:prime256v1")) {
```
VeXocide: I'd love to pick Curve25519 and call it a day, but unfortunately that isn't an option. That said, this set looks like the best we can do for now.

Similarly it might be nice to allow this to be configurable for when Curve25519 becomes available. It might already be if you're using a drop-in replacement like BoringSSL or LibreSSL.

src/clustering/administration/main/command_line.cc
jlhawn: Sure, let's have this be the default and I'll add another config/option for specifying an additional cipher list, maybe `--tls-additional-ciphers`.

Should we also have a `--tls-dh-params` flag too, for specifying diffie-hellman params for the non-EECDH ciphers? If not specified, we could just generate a 2048-bit dhparams file by default the first time that a node starts and save it in some default location. This is still a huge pain though because the server would be stalled for a very long time while it searches for a strong prime, so given that they already have to specify additional ciphers maybe it's not too much to ask for them to generate dhparams too?

jlhawn:
@VeXocide thanks' I'll try to work in your raii wrapper for `SSL_CTX *`.

src/clustering/administration/main/command_line.cc
VeXocide: Instead of having additional ciphers why not allow `--tls-ciphers` instead to specify a string, defaulting to "ECDHE+AESGCM"?

I agree that generating a 2048-bit or stronger dhparams file the first time would be a huge pain, `--tls-dh-params` is the way to go, and it'd be awesome if you're willing to go the distance to add this. Most system administrators worth their salt should already be familiar with such options from their web server and can simply follow those guidelines here as well.

Just a quick note for anyone reading along as this initially confused me, EECDH and ECDHE are the same thing, as are EDH and DHE. Why have one way of doing things when you could have two ..

VeXocide:
Awesome, don't hesitate to ask if you have any other questions regarding C++ or our codebase!

src/clustering/administration/main/command_line.cc
jlhawn: > Instead of having additional ciphers why not allow --tls-ciphers instead to specify a string, defaulting to "ECDHE+AESGCM"?

:+1: 

> Just a quick note for anyone reading along ... EECDH and ECDHE ...

Yeah, I originally learned the latter initialism. And if anyone else reading is interested: `DH` is for [Diffie-Hellman key exchange](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange), `EC` is for [Elliptic Curve cryptography](https://en.wikipedia.org/wiki/Elliptic_curve_cryptography), and the extra `E` is for "Ephemeral" in order have [Forward Secrecy](https://en.wikipedia.org/wiki/Forward_secrecy). Here is a [quick read that explains the difference](http://security.stackexchange.com/questions/46802/what-is-the-difference-between-dhe-and-ecdh).

Unfortunately, I don't know as much about encryption schemes, but [AES](https://en.wikipedia.org/wiki/Galois/Counter_Mode) is a block cipher algorithm which does the actual encryption but only operates on an individual chunk of data (e.g. 128- or 256-bits) while [GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) is a block cipher *mode of operation* which applies the block cipher to a stream of data blocks in such a way that provides encryption and data integrity (i.e., a checksum) for the entire stream of data. This is the method used to actually encrypt the TLS connection. To wind back up the layers of abstraction here, the initial key to use for the block cipher is agreed upon using the Diffie-Hellman exchange, the parameters for which are authenticated to the client using the key pair from the server's certificate.

It's pretty cool how it all gets pieced together, but the major problem it raises is that much of the swappable pieces of this stack are outdated and/or insecure but we still want to maintain compatibility with systems that don't support the latest and greatest.

VeXocide: `ECDHE-RSA-AES256-GCM-SHA384` is the full cipher that OpenSSL would chose on my machine for TLSv1.2 given the `ECDHE+AESGCM` limitation, lets have a look at all parts.

First off is `ECDHE`, which is the algorithm that is used for the key exchange. The key exchange aims to establish a shared secret, that is a secret known to both the server and the client, but not anyone else, not even if you're listening to the communication between the server and the client. The basic underlying algorithm used is called Diffie-Hellman as @jlhawn mentioned, or `DH` for short. If you happen to be familiar with group theory I strongly recommend reading up on it as it's really neat.

There are two problems with plain Diffie-Hellman though, first of which is that there is an attack given smaller primes and the computing resources reserved to nation states. Two solutions exist, and the first is simply using bigger prime numbers. The current recommendation is to use at least 2048-bit primes which is why the OpenSSL default of 1024-bit I was complaining about before is problematic. The other solution is switching to `ECDH` which is a derived algorithm using Elliptic Curve cryptography instead, which for all intents and purposes I consider magic. Mathematically competent people who have wrapped their heads around this have promised me that it's really neat as well, and safe.

The other problem is that it uses the same very hard to solve problem for every connection, which means that if you solve that problem, hard as it may be, you can decrypt all data over those connections. The trick is to use a different very hard problem for each connection, which is what the last `E`, for Ephemeral as @jlhawn said, is for. This is commonly known as providing forward security.

With that out of the way we're on to `RSA` which is a public-key cryptosystem that's used for authentication in this context. A third party, the Certificate Authority, which both the server and client trust and have the public key off signed a document for the server, and the client, that states that they are who they claim they are. This allows the server and client to prove to each other that they are indeed talking to the system they think they are, which is important.

Next is `AES256` or Advanced Encryption Standard which is a block cipher as @jlhawn said. It takes a piece of data and a secret key and mangles it in such a way that you can't read it unless you have the secret key as well. This is one of the reasons that we needed a secret key to begin with. It works on blocks of 256 bits, and pads the message such that it splits up in to 256 bit chunks. This padding is important, and there are ticks to it I'd rather not get into.

There's a problem with straight AES though, if you give it the same data and key twice, it produces the same mangled output. An attacker could learn that you're sending the same message again, which could be important. To solve this they invented cipher block modes and there's a number of them. The simplest is known as CTR or Counter and simply adds an increasing number to each block, thus now it takes the data, key, and counter to produce the mangled output. `GCM` or Galois/Counter Mode is another such mode and the one that is chosen in this cipher, it's state of the art that combines a counter with an authentication code I will go into in a minute.

To recap, we now know that we're talking to the right computer thanks to authentication via `RSA`, we have a shared secret that only we know thanks to `EDCHE`, are able to send them encrypted messages thanks to `AES256` and we won't send the same encrypted message twice if we have the same data due to `GCM`.

This leaves us with `SHA384`, which is used to provide integrity, allowing the client and server to detect when there's someone in the middle changing the message before they receive them. This is usually done using a MAC or Message Authentication Code, or a HMAC or hash version in particular.

Image you have a cryptographically secure hash algorithm, which is exactly what `SHA384` is. In other words, a function that takes a message and generates a short derived message from it, in such a way that it's nearly impossible to find the original message. You can then take the hash of a message, and a secret key and send that along with the original message. As long as the attacker doesn't have the secret key, at which point you'd be in deep trouble, it can't generate the right hash to send along with the message thus allowing you to check whether it's the original message that you've received. Through clever engineering and some wizardry this is now combined into the block cipher mode `GCM` though, in what's called AEAD or Authenticated Encryption with Associated Data.

Hope this helps, feel free to bug me if anything is unclear.

VeXocide: I have to leave in a minute but wanted to add that while it's really cool how it's all pieced together I cut numerous corners in the above explanation. Wikipedia as well as other parts of the internet contain a lot more information on all of the above, and it's definitely worth reading if you want to learn more.

Last but not least @jlhawn is completely right in that a lot of these swappable piece can be replaced with outdated and/or insecure alternatives, that we still keep around to maintain compatibility with systems that don't support the latest and greatest. Fortunately in this case we can give you the latest and greatest, and hopefully keep you secure.

jlhawn:
The last commit adds 3 additional options that we discussed: `--tls-ciphers`, `--tls-ec-curves`, and `--tls-dhparams` to give administrators more control for configuring TLS, though the defaults should be recommended if possible. This should enable backwards (and possible forwards) compatibility with clients and other versions of OpenSSL.

The next few things I was wondering about are:

1. Testing

    How would one add unit/integration tests for these TLS capabilities?

2. Documentation

    Should I add this or is there another team that works on documenting features and 'how-to's?

3. Configuration file

    I haven't added this yet, opting only for command-line options. It should be fairly straightforward to add those options to the config file.

VeXocide:
Awesome, I'm looking forward to this.

1. Testing
    I'm not quite sure how to best fit this into our testing infrastructure to be honest, hopefully @larkost who maintains this can chime in.

2. Documentation
    We have a separate repository for documentation, it's at https://github.com/rethinkdb/docs, and it'd be great if you could open an issue there with the list of parameters you've added. A technical writer can take it from there to add the necessary documentation.

3. Configuration file
    I've just been told that the configuration options are converted to command-line options internally, so this should already work.

Thanks again for all the work you've done here!

jlhawn:
I've got some potentially sad news: I use the following OpenSSL function for ECDHE:

```C
int SSL_CTX_set1_curves_list(SSL_CTX *ctx, char *list);
int SSL_set_ecdh_auto(SSL *s, int onoff);
```

But it looks like these are only available in OpenSSL 1.0.2: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set1_curves.html

I don't see any way of configuring ECDHE in earlier versions of `libssl`.

I've been building this so far on my Mac with OpenSSL 1.0.2 installed and I ran into an error when trying to build it on an Ubuntu 14.04 system. The latest version available on that distro is `OpenSSL 1.0.1f 6 Jan 2014`. It looks like it works on Ubuntu 15.10 though: `OpenSSL 1.0.2d 9 Jul 2015`.

What do you guys think I should do?

danielmewes:
If it's too hard to make this work with versions older than OpenSSL 1.0.2, we can always bump the required OpenSSL version. Though it wouldn't be ideal, since users won't get security updates from their distributors for OpenSSL if we need to link in our own version.

Maybe @VeXocide has an idea?

VeXocide:
OpenSSL 1.0.1. does support `ECDHE` if I understand correctly, and will, by default, send a list of all curves it supports which is clearly undesirable. It boggles my mind that OpenSSL became a defacto standard.

Fortunately we're not the first to run into this and it turns out there's an undocumented `SSL_CTX_set_tmp_ecdh` calling `SSL_CTX_ctrl` with `SSL_CTRL_SET_TMP_ECDH` under water which can be used to set a single curve. This is based on http://openssl.6102.n7.nabble.com/question-about-ecdh-functions-td38239.html which seems to explain how it is to be used. If possible it'd be cool to select the first curve from `--tls-ec-curves` though that might prove tricky.

If this isn't an option I'd be comfortable adding `EDH`, especially now that we have `--tls-dhparams`. I'd like to prevent statically linking OpenSSL for the the reason that @danielmewes mentions, it doesn't have the greatest track record.

VeXocide:
Just after posting I found https://github.com/nodejs/node/blob/master/src/node_crypto.cc#L790 which shows how to select a single curve based on a string. If OpenSSL is at a version below 1.0.2 this is probably the way to go.

larkost:
On the testing front, this is going to be a bit tricky to do in a proper manner since Python's certificate checking is uneven at best. For building (already at `./configure`, which is currently required to run the tests anyways) we already require either `curl` or `wget`, and at least the former has good support for certificate checking. So at least some of the testing could be done though that binary called with Python's `subprocess` module (I would prefer to avoid the PyCurl module since it requires compiling). That would make the web part easy, and it is possible that you could use `curl`'s telnet support to jerry-rig enough to test client and server connections.

Alternately, it would probably not be too difficult to create `c`/`c++` tests around `libcurl` or OpenSSL that could handle all of the cases in a less hacky manner. That would require a little `make` work, but I would be happy to help on that.

larkost:
I don't think that this will be relevant for this work, but on MacOS if the non-system version of OpenSSL is used, then you will not have access to the root certificates. So only self-signed certificates would be viable in that case.

jlhawn:
@danielmewes @VeXocide Yesterday I added a commit which makes this compatible with older versions of OpenSSL. The tradeoff is that we only get to specify one elliptic curve to use for ECDHE. The default is the P-256 NIST/FIPS curve but we still have an option for specifying a different curve if the system admin knows what they are doing. Let me know what you think.

Also, I noticed that there's a developer preview of RethinkDB for Windows which is slated for 2.3. Since this PR is also part of the 2.3 milestone, does that mean I have to make sure this will work on Windows or is there a team of Windows Devs that can port this? Honestly, I haven't investigated how much effort that would be.

Oh, *one more thing!*

I published a Docker image which has a build of this branch: https://hub.docker.com/r/jlhawn/rethinkdb-tls/tags/

I tried to make the image as minimal as possible so it's mostly just the `rethinkdb` binary and some dependent libs. I just recently built it so there's not much documentation for it at the moment, but I'll put together a guide on how to set up a 3 node cluster with mutual TLS authentication and a proxy node with the web UI secured with TLS and driver connections secured with TLS using the existing authkey method for authenticating clients. Expect that some time soon this afternoon.

jlhawn:
Here's that guide I promised: https://gist.github.com/jlhawn/43f84f86579012470dc9

danielmewes:
@jlhawn Ah I see. Sorry I haven't gotten around to taking a deeper look at your code yet. I'll hopefully be able to do that soon.

You shouldn't worry about the Windows compatibility.
We've had some discussion about this internally. For Windows the code will need some changes to get OpenSSL to work with the asynchronous networking code. On Windows you can't just do a non-blocking read or write and wait for a notification once you're supposed to retry, but instead need to register something they call completion ports (this post seems to describe all the details: https://famellee.wordpress.com/2013/02/20/use-openssl-with-io-completion-port-and-certificate-signing/ ).
This is something we can take care of post merge (though if you want to look into it, that would be awesome too of course! The Windows preview is build from the branch atnnn/windows2 I think).

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+    tls_conn_wrapper_t(SSL *_conn) {
+        conn = _conn;
+    }
+
+    ~tls_conn_wrapper_t() {
+        SSL_free(conn);
+    }
+
+    // Transfer ownership of the connection to the caller.
+    void reclaim() {
```
danielmewes: For consistency with other similar structures in our code and in the standard libraries, I suggest changing this function into the following:
```C++
// Transfer ownership of the connection to the caller.
SSL *release() {
    rassert(conn != nullptr);
    SSL *res = conn;
    conn = nullptr;
    return res;
}
```

This also fixes two other minor things:
1. Use `nullptr` instead of `NULL` (this is mostly a stylistic change)
2. Adds an assertion that the connection hasn't been released before. `rasserts` are enabled when compiling with `make DEBUG=1`.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
     event_watcher->stop_watching_for_errors();
 }
 
+/* tls_conn_wrapper_t wraps a TLS connection and safely releases it in
+exceptional cases if it is not reclaimed before execution leaves its scope. */
+class tls_conn_wrapper_t {
+public:
+    tls_conn_wrapper_t(SSL *_conn) {
```
danielmewes: This constructor should have an `explicit` keyword
```C++
explicit tls_conn_wrapper_t(SSL *_conn) {
```

(I'll probably have a bunch of minor stylistic things like this. I hope you don't mind. These things are fairly arbitrary, but you can find our style guide at https://github.com/rethinkdb/rethinkdb/blob/daniel_coding_style/STYLE.md )

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+    }
+
+private:
+    SSL *conn;
+};
+
+/* This is the client version of the constructor. The base class constructor
+will establish a TCP connection to the peer at the given host:port and then we
+wrap the tcp connection in TLS using the configuration in the given tls_ctx. */
+linux_secure_tcp_conn_t::linux_secure_tcp_conn_t(SSL_CTX *tls_ctx, const ip_address_t &host, int port, signal_t *interruptor, int local_port) THROWS_ONLY(connect_failed_exc_t, interrupted_exc_t) :
```
danielmewes: Another small style thing that you couldn't possibly have known because our style guide wasn't available anywhere: This line should be wrapped to an 89 character line length (maximum).

src/arch/io/network.cc
jlhawn: I don't mind at all. I should be able to make fixes like these very quickly. I know coding style consistency is very important for readability and maintainability. Sorry, I didn't even think to look for a style guide, but I figured the project maintainers would point these things out :+1:

danielmewes: Actually even better: Just call `SSL_new` directly in the `tls_conn_wrapper_t` constructor rather than passing it in.

There are two additional things you'll need to do:
1. According to https://www.openssl.org/docs/manmaster/ssl/SSL_new.html `SSL_new` can return `nullptr`. You'll need to check for that and throw a `connect_failed_exc_t`.
2. Make sure that if a connection is thrown, the value of the `conn` field is still initialized to `nullptr`. Then in the `~tls_conn_wrapper_t` destructor, check if `conn` is `nullptr` and only call `SSL_free` if it's not. (throwing the exception from the constructor will immediately run the destructor, so it has to work even in the error case)

danielmewes: Also, since the connection is constructed inside the `tls_conn_wrapper_t` then, you'll need a `TLS *get()` method to get access to the pointer in the `linux_secure_tcp_conn_t`. On the plus side, it makes the code there simpler because you only need to keep the `tls_conn_wrapper_t` around and not both the wrapper and a pointer to the `conn` itself.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+        linux_tcp_conn_t(host, port, interruptor, local_port) {
+
+
+    conn = SSL_new(tls_ctx);
```
danielmewes: See the commend above about getting rid of `conn` and only having `conn_wrapper`

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+    establish_conn(SSL_connect);
+
+    conn_wrapper.reclaim();
+}
+
+/* This is the server version of the constructor */
+linux_secure_tcp_conn_t::linux_secure_tcp_conn_t(SSL_CTX *tls_ctx, fd_t sock) :
+        linux_tcp_conn_t(sock) {
+
+    conn = SSL_new(tls_ctx);
```
danielmewes: This needs to use the `tls_conn_wrapper_t` as well

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+
+    /* If establish_conn() throws an exception, the conn_wrapper deallocates
+    the connection when unwinding the stack. */
+    tls_conn_wrapper_t conn_wrapper(conn);
+
+    establish_conn(SSL_connect);
+
+    conn_wrapper.reclaim();
```
danielmewes: This can be removed once you keep a `tls_conn_wrapper_t` as a member instead of `conn`.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+        logNTC("linux_secure_tcp_conn_t server constructor exception: %s\n", err.info.c_str());
+    }
+}
+
+linux_secure_tcp_conn_t::~linux_secure_tcp_conn_t() THROWS_NOTHING {
+    assert_thread();
+
+    if (is_open()) shutdown();
+
+    SSL_free(conn);
```
danielmewes: (this as well can go away once `conn` is a `tls_conn_wrapper_t`)

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+}
+
+/* Creates a new OpenSSL connection using the underlying fd and the given
+OpenSSL context. The given handshake function is used to perform the handshake
+for the client or server using SSL_connect or SSL_accept respectively. */
+void linux_secure_tcp_conn_t::establish_conn(int (*handshake)(SSL *) ) THROWS_ONLY(connect_failed_exc_t) {
```
danielmewes: While using a function pointer as the handshake isn't wrong, you can make this more flexible by using `std::function<int(SSL *)>` (make sure to add an `#include <functional>` at the top as well).
That allows whoever calls this function to pass in either a traditional function pointer (which gets automatically cast to an std::function), or a C++ 11 lambda function, or the result of `std::bind`.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+    // Add support for partial writes.
+    SSL_set_mode(conn, SSL_MODE_ENABLE_PARTIAL_WRITE);
+
+    if (NULL == conn) {
```
danielmewes: I suggest moving this check into the `tls_conn_wrapper_t` constructor as mentioned above, though you can leave it this way if you prefer.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+        int err_code = SSL_get_error(conn, ret);
+        unsigned long io_err = 0;
+
+        switch (err_code) {
+        case SSL_ERROR_WANT_READ:
+            /* There's no data available right now, so we must wait for a
+            notification from the epoll queue, or for an order to shut down. */
+            {
+                linux_event_watcher_t::watch_t watch(get_event_watcher(), poll_event_in);
+                wait_any_t waiter(&watch);
```
danielmewes: The `wait_any_t` is not necessary here, because it's only waiting on `watch` anyway. You could call `wait_lazily_unordered()` directly on `watch`.

However we should also make sure that the handshake can get interrupted if for example the user shuts down the server while we're in here.
`establish_conn` should take the `signal_t *interruptor` as an argument, and here you should use
```C++
wait_interruptible(&watch, interruptor);
```

`wait_interruptible` throws an `interrupted_exc_t` if it gets interrupted (which also needs to be added to the `THROWS_ONLY` declaration of `establish_conn` above).

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+            }
+            /* Go around the loop and try to complete the handshake */
+            continue;
+        case SSL_ERROR_WANT_WRITE:
+            /* Wait for a notification from the event queue, or for an order to
+            shut down */
+            {
+                linux_event_watcher_t::watch_t watch(get_event_watcher(), poll_event_out);
+                wait_any_t waiter(&watch);
+                waiter.wait_lazily_unordered();
```
danielmewes: Same here about waiting.

src/arch/io/network.cc
danielmewes: Feel free to ignore this. I see that you only pass in `SSL_accept`, so it doesn't make much difference right now.

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+                throw tcp_conn_read_closed_exc_t();
+            }
+            /* Go around the loop and try to write again */
+            continue;
+        case SSL_ERROR_SYSCALL:
+            // An error with the underlying IO.
+            io_err = ERR_get_error();
+            if (io_err != 0) {
+                // The Openssl error queue has data on the error.
+                on_shutdown();
```
danielmewes: I don't understand why it's safe to call `on_shutdown()` here (and in the couple cases below). How do we know that the connection was already shut down?
Should we call `shutdown()` instead?

src/arch/io/network.cc
```diff
@@ -621,6 +621,430 @@ void linux_tcp_conn_t::on_event(int /* events */) {
...
+        if ((ret = SSL_shutdown(conn)) >= 0) {
+            break; // "close notify" was either sent or received.
+        }
+
+        // A negative ret indicates as error.
+        int err_code = SSL_get_error(conn, ret);
+        if (SSL_ERROR_WANT_READ == err_code) {
+            linux_event_watcher_t::watch_t watch(get_event_watcher(), poll_event_in);
+            wait_any_t waiter(&watch);
+            waiter.wait_lazily_unordered();
```
danielmewes: This is a bit risky because if the remote side never reads from the socket or (further down) doesn't send a response, we will be stuck in here forever waiting for `waiter` I think. The server could not even be shut down in that case.
I suggest adding a timeout to `waiter` for all waits in this `shutdown()` function. After the wait you can check if the timeout was pulsed. If it was, we continue streight to line 1024 and simply close the socket.

I suggest using a single timeout for the whole `shutdown()` process, rather than one per wait.

Something like this:
```
// <at the beginning of `shutdown()`:>
// Wait at most 5 seconds for the orderly shutdown. If it doesn't complete by then, we
// simply shutdown the socket.
signal_timer_t shutdown_timeout(5000);

// <further down in the waiter code:>
wait_any_t waiter(&watch, &timeout);
waiter.wait_lazily_unordered();
if (timeout.is_pulsed()) {
  break;
}
```

src/arch/io/network.cc
```diff
@@ -629,13 +1053,13 @@ linux_tcp_conn_descriptor_t::~linux_tcp_conn_descriptor_t() {
     rassert(fd_ == -1);
 }
 
-void linux_tcp_conn_descriptor_t::make_overcomplicated(scoped_ptr_t<linux_tcp_conn_t> *tcp_conn) {
-    tcp_conn->init(new linux_tcp_conn_t(fd_));
+void linux_tcp_conn_descriptor_t::make_connection(SSL_CTX *tls_ctx, scoped_ptr_t<linux_tcp_conn_t> *tcp_conn) {
+    tcp_conn->init((NULL == tls_ctx) ? new linux_tcp_conn_t(fd_) : new linux_secure_tcp_conn_t(tls_ctx, fd_));
```
danielmewes: `nullptr` instead of `NULL`

src/arch/io/network.hpp
```diff
@@ -9,6 +9,9 @@
 #include <arpa/inet.h>
 #include <netinet/in.h>
```
danielmewes: We usually don't have extra newlines in the includes within the block of C include files. This newline should probably go away (the one after the openssl includes and before `<functional>` should stay though, since that's a new block that starts C++ includes)

src/arch/io/network.cc
danielmewes: If it's not included yet, you'll need to `#include "arch/timing.hpp"` to get `signal_timer_t`.

src/arch/io/network.hpp
```diff
@@ -47,6 +50,10 @@ class linux_tcp_conn_t :
             error(en),
             info("Could not make connection: " + errno_string(error)) { }
 
+        explicit connect_failed_exc_t(int en, std::string err_str) :
```
danielmewes: The string should be passed in by const reference:
```C++
connect_failed_exc_t(int en, const std::string &err_str)
```

src/arch/io/network.hpp
```diff
@@ -298,22 +304,68 @@ class linux_tcp_conn_t :
...
 
-    scoped_ptr_t<auto_drainer_t> drainer;
+class linux_secure_tcp_conn_t :
+    public linux_tcp_conn_t {
+public:
+
+    friend class linux_tcp_conn_descriptor_t;
+
+    // Client connection constructor.
+    linux_secure_tcp_conn_t(SSL_CTX *tls_ctx, const ip_address_t &host, int port, signal_t *interruptor, int local_port = ANY_PORT) THROWS_ONLY(connect_failed_exc_t, interrupted_exc_t);
```
danielmewes: Line length

src/arch/io/network.hpp
```diff
@@ -129,8 +136,6 @@ class linux_tcp_conn_t :
 
     ~linux_tcp_conn_t() THROWS_NOTHING;
```
danielmewes: Since this class now has virtual methods, the constructor must also be declared `virtual`.

src/client_protocol/server.hpp
```diff
@@ -148,6 +151,8 @@ class query_server_t : public http_app_t {
     scoped_ptr_t<tcp_listener_t> tcp_listener;
 
     int next_thread;
+
+    SSL_CTX *tls_ctx;
```
danielmewes: Though it's just a pointer and it probably won't matter in practice, this field should be declared above the `drainer` (e.g. below the `handler`). That ensures that during destruction, the `tls_ctx` field remains valid as long as connections can still be active or get established.

src/clustering/administration/main/command_line.cc
```diff
@@ -12,6 +12,11 @@
 #include <sys/stat.h>
 #include <sys/types.h>
```
danielmewes: Similar to above, this newline should probably not be there.

jlhawn:
@danielmewes Thanks for the review! I don't think I'll have much time to make a patch for these comments in the next 24 hours but I expect to be able to have a commit which addresses them all by Friday.

danielmewes:
Awesome. I still need to look at a few more files that you changed, but I
believe @vexocide and I have together covered the most complex parts by now.

This is great work by the way. I'm really happy about this pull request. :)

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+        return false;      
+    }
+
+    if(SSL_CTX_use_certificate_file(tls_ctx, cert_file.c_str(), SSL_FILETYPE_PEM) <= 0) {
+        ERR_print_errors_fp(stderr);
+        return false;      
+    }
+
+    if(1 != SSL_CTX_check_private_key(tls_ctx)) {
+        logNTC("private key and certificate do not match");
```
danielmewes: This should be a `logERR` not `logNTC`.

danielmewes: (also capitalization and punctuation: "The private key and certificate do not match.")

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+
+    return true;
+}
+
+bool configure_web_tls(const std::map<std::string, options::values_t> &opts, SSL_CTX *web_tls) {
+    boost::optional<std::string> key_file = get_optional_option(opts, "--web-tls-key");
+    boost::optional<std::string> cert_file = get_optional_option(opts, "--web-tls-cert");
+
+    if (!(key_file && cert_file)) {
+        logERR("must specify both --web-tls-key and --web-tls-cert");
```
danielmewes: How about "--web-tls-key and --web-tls-cert must be specified together." ?

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+        logERR("must specify both --driver-tls-key and --driver-tls-cert");
+        return false;
+    }
+
+    if (!load_tls_key_and_cert(driver_tls, *key_file, *cert_file)) {
+        return false;
+    }
+
+    if (ca_file) {
+        if (!SSL_CTX_load_verify_locations(driver_tls, ca_file->c_str(), NULL)) {
```
danielmewes: `nullptr`

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+            ERR_print_errors_fp(stderr);
+            return false; 
+        }
+
+        // Mutual authentication.
+        SSL_CTX_set_verify(driver_tls, SSL_VERIFY_PEER|SSL_VERIFY_FAIL_IF_NO_PEER_CERT, NULL);
```
danielmewes: `nullptr`

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+    FILE *get() {
+        return fp;
+    }
+
+private:
+    FILE *fp;
+};
+
+bool initialize_tls_ctx(const std::map<std::string, options::values_t> &opts, SSL_CTX **tls_ctx) {
+    *tls_ctx = SSL_CTX_new(SSLv23_method());
```
danielmewes: We should use a RAII wrapper for `SSL_CTX`, so we don't leak resources in case an exception gets thrown somewhere.

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+    if (NULL == *tls_ctx) {
```
danielmewes: `nullptr`

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+    configure their own ciphers list if their SSL/TLS lib doesn't support these
+    options or if they want to use other cipher suites. */
+    boost::optional<std::string> ciphers_opt = get_optional_option(opts, "--tls-ciphers");
+    /* NOTE: I normally prefer the abreviation 'ECDHE', but older versions of
+    OpenSSL do not seem to like 'ECDHE' even though the matching cipher suites
+    have that prefix. Yeah, I know... it's very frustrating. Both older (1.0.1)
+    and newer (1.0.2) versions seem to be okay with 'EECDH' though. */
+    std::string ciphers = ciphers_opt ? *ciphers_opt : "EECDH+AESGCM";
+    if (0 == SSL_CTX_set_cipher_list(*tls_ctx, ciphers.c_str())) {
+        logNTC("No secure cipher suites available\n");
```
danielmewes: All the `logNTC` in this function should be `logERR` as well.

src/clustering/administration/main/command_line.cc
jlhawn: I thought about that, but then I considered that if an exception were raised here it would probably cause the server to fail to initialize anyway. Or would it continue to chug along?

src/clustering/administration/main/command_line.cc
```diff
@@ -645,6 +650,244 @@ service_address_ports_t get_service_address_ports(const std::map<std::string, op
...
+bool configure_tls(const std::map<std::string, options::values_t> &opts, tls_configs_t *tls_configs) {
+    // Setup OpenSSL context.
+    SSL_library_init();
+    SSL_load_error_strings();
+
+    logNTC("%s\n", SSLeay_version(SSLEAY_VERSION));
+
+    if(!exists_option(opts, "--no-http-admin") && exists_option(opts, "--web-tls")) {
+        if (!(initialize_tls_ctx(opts, &(tls_configs->web)) && configure_web_tls(opts, tls_configs->web))) {
+            SSL_CTX_free(tls_configs->web);
```
danielmewes: Once we have a wrapper around SSL_ctx, this would probably become something like `tls_configs->web.reset()`;

(Alternatively, and somewhat cleaner IMO, `initialize_tls_ctx` should take care of resetting the SSL context itself before returning `false`.)

src/clustering/administration/main/command_line.cc
```diff
@@ -1165,6 +1464,7 @@ void get_rethinkdb_create_options(std::vector<options::help_section_t> *help_out
     help_out->push_back(get_help_options(options_out));
     help_out->push_back(get_log_options(options_out));
     help_out->push_back(get_config_file_options(options_out));
+    help_out->push_back(get_tls_options(options_out));
```
danielmewes: This should not be included in the `create` options.

src/clustering/administration/main/command_line.cc
```diff
@@ -1178,6 +1478,7 @@ void get_rethinkdb_serve_options(std::vector<options::help_section_t> *help_out,
...
```
danielmewes: Suggestion: Move this below the `get_network_options` above. That way the network options will be printed next to the TLS options.

src/clustering/administration/main/command_line.cc
```diff
@@ -1189,6 +1490,7 @@ void get_rethinkdb_proxy_options(std::vector<options::help_section_t> *help_out,
...
```
danielmewes: Same here about moving it below `get_network_options`

src/clustering/administration/main/command_line.cc
```diff
@@ -1203,6 +1505,7 @@ void get_rethinkdb_porcelain_options(std::vector<options::help_section_t> *help_
...
```
danielmewes: ... and same here

src/clustering/administration/main/serve.hpp
```diff
@@ -85,6 +87,13 @@ struct service_address_ports_t {
     int port_offset;
 };
 
+struct tls_configs_t {
+    tls_configs_t() : web(NULL), driver(NULL), cluster(NULL) {}
```
danielmewes: `nullptr` (though this would change anyway with `SSL_CTX` wrappers I assume)

src/clustering/administration/main/command_line.cc
danielmewes: I'll add a suggestion for the wrapper type below in `server.hpp`.

src/clustering/administration/main/serve.hpp
danielmewes: After thinking about it a little, I think the easiest way for making `SSL_CTX` RAII is by using an `std::shared_ptr`. This is how you can do it:
```C++
#include <memory>
...

typedef std::shared_ptr<SSL_CTX> shared_ssl_ctx_t;
```

Then in your `initialize_tls_ctx` function, the first line would become:
+    *tls_ctx = SSL_CTX_new(SSLv23_method())`

src/http/http.hpp
```diff
@@ -13,6 +13,8 @@
 #include <boost/shared_array.hpp>
 #include <boost/optional.hpp>
 
+#include <openssl/ssl.h>
```
danielmewes: This should be at the top of all `#include`s in this file (before the boost includes and before the C++ STL includes)

src/rdb_protocol/query_server.hpp
```diff
@@ -4,6 +4,8 @@
 
 #include <set>
 
+#include <openssl/ssl.h>
```
danielmewes: Same here about include ordering (see point 5 of the style guide https://github.com/rethinkdb/rethinkdb/blob/daniel_coding_style/STYLE.md . This counts as a C system header.)

src/rpc/connectivity/cluster.hpp
```diff
@@ -8,6 +8,8 @@
 #include <utility>
 #include <vector>
 
+#include <openssl/ssl.h>
```
danielmewes: ... and here too

danielmewes:
@jlhawn I finished my pass over the code. Please let me know if you have any questions or something is unclear.

src/clustering/administration/main/command_line.cc
danielmewes: You're right. It's more a general precautionary measure. We don't know how this code will be shuffled around at some point in the future. Putting it into a RAII wrapper makes it impossible to forget freeing it. I added a suggestion to `serve.hpp` that uses `std::shared_ptr` and is super simple. I think it would do the job.

danielmewes:
Thanks again for updating the PR.
I added a few last comments. Once you address those, we can merge this. :-)

danielmewes:
Awesome, this looks good. :tada: :clap: 

I have to port this to `next`, but it should be straight forward.

danielmewes:
This is in `next` as of d14a456 and will ship with RethinkDB 2.3.0.

Many thanks again for your work on this @jlhawn !

Issue #5390, opened 2016-02-11
java client->Response.isFeed(),notes.stream () is Empty,allMatch return true.


xingmu:
use Stream.anyMatch()?

deontologician:
Hi @xingmu, do you think this is this a logic error in the driver? Or are you suggesting a refactoring?

deontologician:
OK, nevermind, we found the bug. Thanks for reporting it!

PR #5400, opened 2016-02-14
Java: ResponseNotes was never used.

bchavez:
Discovered in `C#/.NET` driver by inspection analyzers. #5390 related. `Cursor.isFeed` would still always be false.


deontologician:
Thanks, score one for static analysis

PR #5416, opened 2016-02-18
Remove error log message from normal exit

deontologician:
Due to the way the .close() method works, this error log is always hit unless there's a termination from an io exception. Also moving the isOpen() check to the loop to be more explicit.


danielmewes:
Looks good to me. Let's merge the other one (assuming it looks good to you too) :)

Issue #5380, opened 2016-02-09
Java driver connection/cursor handling simplification


pires:
As per @bchavez comments:

> So, after some additional review, I have a strong feeling the ResponsePump and awaiters stuff should really go into SocketWrapper like the .NET driver. I think there's too much state to "remember to clear" in Connection class if a connection is restarted... and it will be harder to get right as Connection code gets bigger in the future ... it also encourages usage of the awatiers dictionary at the Connection level, which shouldn't be accessed by anything else other than the response pump like this: pires#2
> 
> if ResponsePump and awatiers get encapsulated in SocketWrapper generally these problems go away like in the .NET driver when a new socket is created on reconnect.... Currently, the Java Connection a lil bit more complicated to get right the way it is setup now.... For example: how awaiters is being cleared at the Connection.close() level ... if we move ResponsePump and it's awaiters state in SocketWrapper these network "state" troubles go away because it's logically separated from the main connection class.
> 
> :heavy_check_mark: Connection is responsible for closing high-level Cursors. Then...
> :heavy_check_mark: SocketWrapper:ResponsePump is responsible for cleaning up it's own awaiters.
> 
> SocketWrapper and it's state wont ever be used again, and can safely be GC'ed. And avoids "forgot to clear" state issues.
> 
> Let me know what you think. If you think the name SocketWrapper is too low level, maybe we could rename it. Dunno.

I agree and this would be good to have.

adriantodt:
I implemented this without even knowing that this issue existed. This is now implemented to one of the branches (https://github.com/rethinkdb/rethinkdb-java/pull/12) of the Java driver, which now lives in a separated repository (https://github.com/rethinkdb/rethinkdb-java). This issue can be closed.

PR #5417, opened 2016-02-18
Don't print an error if the socket is closed.

danielmewes:
This can happen as part of a normal conn.close() call.

@deontologician @pires

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -198,7 +198,7 @@ public class Connection {
 
             // close the socket
             socket.ifPresent(SocketWrapper::close);
-            if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
+            if (exec != null && !exec.awaitTermination(60, TimeUnit.SECONDS)) {
                 log.error("Executor did not terminate");
             }
         }
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -194,6 +194,9 @@ public class Connection {
             // terminate response pump
             if (exec != null && !exec.isShutdown()) {
                 exec.shutdown();
+                if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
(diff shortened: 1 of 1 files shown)


danielmewes:
Hmm maybe we should still add an `exec.awaitTermination` to `close()` after it has closed the socket?

deontologician:
Actually, yeah, that seems sensible. If it doesn't finish we know there's something we're missing

pires:
Ah, good call! LGTM.

danielmewes:
It's missing an exception handler:
```
/home/ssd3/daniel/rethinkdb/drivers/java/src/main/java/com/rethinkdb/net/Connection.java:201: error: unreported exception InterruptedException; must be caught or declared to be thrown
            if (exec != null && !exec.awaitTermination(60, TimeUnit.SECONDS)) {
```
We should probably print a warning if we get interrupted.

danielmewes:
... or actually the right thing to do might be passing the interruptedException on and adding it to the `close()` signature?

danielmewes:
Do we have other API functions that handle thread interruptions? How are we handling interrupted exceptions there?

pires:
@danielmewes I thinks only the `responsePump` should deal with execution interruption. It's just one thread btw since `exec = Executors.newSingleThreadExecutor()` so it shouldn't be dramatic to pass the exception to `close()`.

bchavez:
A bit of caution of the bug here https://github.com/rethinkdb/rethinkdb/issues/5448

Also, I'm not sure how safe it is in Java for the executor thread to `.close` on itself since the pump can call `.close` (on exception) then `exec.awaitTermination(60, TimeUnit.SECONDS)`? Would it await on itself for termination? I dunno about this, seems a bit twisty for me.

Lo, I've a subtle fondness for https://github.com/rethinkdb/rethinkdb/issues/5380 and pushing all the awaiter mechanics down a lower level into `SocketWrapper` (or renamed to something else) like the C#/.NET driver. Just my feeling when *connection mechanics* (awaiter stuffs) are at a lower level, there is a **clearer** *separation of concerns* between a logical connection object and the underlying connection mechanics.

danielmewes:
Superseded by https://github.com/rethinkdb/rethinkdb/pull/5449

PR #5449, opened 2016-02-25
Fixes resource leak in Java driver

### LINKED ISSUE #5448: Java disconnection leaks awaiting query requestors
As per @bchavez comments on #5439:

>@pires, there's a subtle bug in the code when processing exceptions this way.
>
>https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L139
>
>After the pump is finished processing a response cleanly, suppose the connection drops outside of the try{}catch{} statement. The pump isn't around to catch it. !isOpen is false. The while(true) loop breaks in Line 145. .close is never called. Hence, leaving all awaiting queries hanging.

pires:
Fixes #5448

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -141,7 +141,8 @@ public class Connection {
             while (true) {
                 // validate socket is open
                 if (!isOpen()) {
-                    log.error("Response Pump: The socket is not open, exiting.");
+                    awaiterException = new IOException("The socket is closed, exiting response pump.");
+                    this.close();
                     break;
                 }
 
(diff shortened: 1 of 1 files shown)


bchavez:
Possibly in conflict with this PR: https://github.com/rethinkdb/rethinkdb/pull/5417

pires:
It does conflict but I'm not yet sure that PR will be merged. Anyway, rebasing should be simple. Waiting on @danielmewes for a call on how to proceed. Thanks for raising awareness.

deontologician:
I agree with your comments from the other PR, but I think we should also remove the error log since this will happen on a normal termination as well.

pires:
@deontologician logs removed.

deontologician:
ok I think this is good to merge, and we can close #5417 since it solves that problem. Any objecttion @danielmewes ?

danielmewes:
Sounds good to me :+1:

deontologician:
Merged, thanks again @pires and @bchavez

PR #5471, opened 2016-03-01
Java Connection implements Closeable

### LINKED ISSUE #5468: Java Cursor as `Closeable`
Thank you for the recently released Java driver! A feature that may make a nice addition to the driver would be if the `com.rethinkdb.net.Cursor` instance could implement the `Closeable` interface. This would make it easier to use through Java's [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) idiom.

deontologician:
Fixes #5468 

ping @danielmewes for review


danielmewes:
Looks good :+1:

PR #5478, opened 2016-03-02
JS Driver: expose `each` and `eachAsync` on all iterable types

### LINKED ISSUE #5469: support eachAsync for `r.expr([1,2,3])` in JavaScript
`eachAsync` is not currently supported for the return of  `r.expr([1,2,3]).run(conn)`. This should probably be done in order to make all of our forms similar.

marshall007:
Fixes #5469

### DIFF
--- a/drivers/javascript/cursor.coffee
+++ b/drivers/javascript/cursor.coffee
@@ -38,8 +38,6 @@ class IterableResult
         @_closeCbPromise = null
 
         @next = @_next
-        @each = @_each
-        @eachAsync = @_eachAsync
 
     _addResponse: (response) ->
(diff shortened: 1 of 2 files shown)


danielmewes:
Oh cool, thanks!

@nighelles can you review this please?

drivers/javascript/cursor.coffee
```diff
@@ -255,6 +246,9 @@ class IterableResult
         return nextCb().nodeify(errCb)
     )
 
+    _each: @::each
+    _eachAsync: @::eachAsync
```
marshall007: These lines are really just for backwards-compatibility. We could remove them if we update the few references of `@_each` to `@each` in this class.

nighelles:
Closing because #5518 has the same changes.

PR #5067, opened 2015-11-10
Only set __proto__ in TermBase if it exists (attempt 2)

mikemintz:
This is a rewrite of PR https://github.com/rethinkdb/rethinkdb/pull/4534 to resolve issue https://github.com/rethinkdb/rethinkdb/issues/162 (IE10 support in the JS driver).

The only difference is replacing `self` with `return self`. I'm not familiar with CoffeeScript, but it's now closer to the original code. The tests that were failing before are now passing for me.

PR #5499, opened 2016-03-09
Add missing return keyword in TermBase constructor

mikemintz:
See https://github.com/rethinkdb/rethinkdb/issues/162 and https://github.com/rethinkdb/rethinkdb/pull/5067 for context.

### DIFF
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -57,7 +57,7 @@ class TermBase
             self.__proto__ = @.__proto__
             return self
         else
-            @
+            return @
 
     run: (connection, options, callback) ->
         # Valid syntaxes are


danielmewes:
Thanks. So the one we merged had the `return` in the `if` case, but didn't have it in the `else` case?

Is this causing an actual problem that you're running into? Just trying to understand how to test this...

mikemintz:
Sorry for the confusion, I wasn't even thinking about the `if` case just now because it's skipped for IE10.

But yes, that's right. This change only affects IE10, not the tests (which have `__proto__`). I tested it manually by adding that `return` keyword in the `ast.js` under node_modules for a test app. As long as the output JS code contains `return this;` in the `else` clause, it'll work as expected.

danielmewes:
Gotcha. Merging this.

danielmewes:
It will ship with RethinkDB 2.2.6.

Issue #5503, opened 2016-03-11
ruby rethink driver does not seem to honour timeout flag


icesmurf:
``
#!/usr/bin/env ruby
require 'timeout'
require 'rethinkdb'
include RethinkDB::Shortcuts

host = '192.168.40.10'

begin
  Timeout::timeout(10) {
    begin
      puts "attempting to connect to #{host}"
      r.connect(:host=> host, :port=>28015, :timeout => 5).repl
    rescue Exception => e
      pp e
      puts "Connection to #{host} not established, error at RQLDriver level"
      exit
    end
  }
rescue Timeout::Error
  puts "Ruby Level Timeout"
  exit
end
``

Output: 
attempting to connect to 192.168.40.10
Ruby Level Timeout

I firewalled 192.168.40.10 with iptables and --DROP

mlucy:
Thanks for the bug report @icesmurf !

I tried this locally and it successfully times out for me.  What's your environment?  (Specifically, what OS are you on, what version of the RethinkDB driver, and what Ruby version.)

icesmurf:
ruby-2.2.1, rethinkdb (2.2.0.4)

```#0:/var/lib/gems/2.1.0/gems/rethinkdb-2.2.0.4/lib/net.rb:688:RethinkDB::Connection:-:       socket = TCPSocket.open(@host, @port)
```

That's not wrapped in a timeout, and it's the section where AFAIK ruby is attempting to setup the socket connection, (which it can't since it's firewalled). Unless i'm missing something (quite possibly, i'm only intermediate at ruby at best), i can't see anything that'd wrap that connection in a timeout in the code at all. If you mean the code i wrote successfully times out, if you get the "Ruby Level Timeout", then that's NOT what we're expecting, we're expecting the RQL error to be triggered in the rescue block. My expectation as a API consumer here is that if i specify a :timeout, i shouldn't have to manually wrap it in a timeout block ;-)

[execution.txt](https://github.com/rethinkdb/rethinkdb/files/170642/execution.txt)

icesmurf:
Actually, looking at that trace a bit more, Timeout.timeout is a bad idea, it spawns a new thread for a start. Probably a better option is to use the tcp_timeout https://github.com/lann/tcp-timeout-ruby , either that, or write our own non-blocking socket code that we can customize the timeout. 

References: 

http://stackoverflow.com/questions/21014272/how-ruby-tcpsocket-timeout-is-defined
[...]

PR #5512, opened 2016-03-13
fix for #5503

icesmurf:
First cut at a patch for ruby connection timeouts when establishing a tcp connection, seems to work as expected here.. I've tested for most of the stuff i was having issues with, and it doesn't seem to have broken anything.. not sure about writing spec tests or anything?


danielmewes:
Nice, thanks for the PR @icesmurf .

If you haven't done this yet, would you mind going to http://rethinkdb.com/community/cla/ and signing the CLA? We require this before we can merge any PR.

Pinging @mlucy for reviewing your patch.

icesmurf:
Yeah, i've already signed it. :-)

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
         verify_cert!(@socket, context)
       else
         @socket = base_socket
+        socket_connect
       end
     end
 
     def base_socket
-      socket = TCPSocket.open(@host, @port)
+      @socket.close if @socket
```
mlucy: I don't think we should be calling `@socket.close` here.  This was previously a pure function, so it's a little weird to have it modifying a member variable.  If we want to be closing the socket, I think we should be doing it in `init_socket`.  In this case, though, `init_socket` is only called by `connect`, which checks that `@socket` isn't set immediately before calling it on line 663, so we should probably just omit this conditional close entirely.

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
         context = create_context(@ssl_opts)
         @socket = OpenSSL::SSL::SSLSocket.new(@tcp_socket, context)
         @socket.sync_close = true
-        @socket.connect
+        socket_connect
```
mlucy: I'm a little confused by this logic.  Below we call `socket_connect` after `base_socket`, but here we call `base_socket`, then put that into an `SSLSocket`, then call `socket_connect` on that.  Before we followed the same code path here for `@tcp_socket` as we did for `@socket` below, why did that change?

(Also, just to double-check, did you test these changes with both SSL and non-SSL connections?)

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
     def base_socket
-      socket = TCPSocket.open(@host, @port)
+      @socket.close if @socket
+      socket = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 0)
       socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
       socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE, 1)
       socket
     end
 
+    def socket_connect
```
mlucy: Could you add a comment explaining what this code is doing and why it solves the timeout issue?

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
+      address = Socket.getaddrinfo(@host, nil, Socket::AF_INET).first[3]
+      sockaddr = Socket.pack_sockaddr_in(@port, address)
+      begin
+        @socket.connect_nonblock(sockaddr)
+      rescue Errno::EINPROGRESS
+        IO.select([@socket], [@socket], [@socket], @timeout)
+        begin
+          @socket.connect_nonblock(sockaddr)
+        rescue Errno::EISCONN
+          return
```
mlucy: This is the case where we've failed to connect, right?  Shouldn't we be throwing an exception?

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
+        rescue Errno::EALREADY
+          @socket.close
+          raise ReqlRuntimeError, "Connection timed out"
```
mlucy: Previously we were raising `Timeout::Error` for connection timeouts.  We should probably continue to do that.

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
       socket
     end
 
+    def socket_connect 
+      address = Socket.getaddrinfo(@host, nil, Socket::AF_INET).first[3]
+      sockaddr = Socket.pack_sockaddr_in(@port, address)
+      begin
+        @socket.connect_nonblock(sockaddr)
+      rescue Errno::EINPROGRESS
+        IO.select([@socket], [@socket], [@socket], @timeout)
```
mlucy: The documentation for `select` says that it will return `nil` if the timeout value is given and no object is ready in that number of seconds.  It seems clearer to check that and raise the timeout error here.

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
+        begin
+          @socket.connect_nonblock(sockaddr)
+        rescue Errno::EISCONN
+          return
+        rescue Errno::EALREADY
```
mlucy: Why does `Errno::EALREADY` indicate a timeout?  The documentation seems to say that `EALREADY` is the same as `EINVAL`, which means "the `socket` is a listening socket".

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
       socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
       socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE, 1)
       socket
     end
 
+    def socket_connect 
+      address = Socket.getaddrinfo(@host, nil, Socket::AF_INET).first[3]
+      sockaddr = Socket.pack_sockaddr_in(@port, address)
+      begin
+        @socket.connect_nonblock(sockaddr)
```
mlucy: Why do we call `connect_nonblock`, and if we get `EINPROGRESS` call `select` and then `connect_nonblock` again?  Couldn't we just call `select` first, followed by `connect_nonblock` once we know it's safe?

drivers/ruby/lib/net.rb
```diff
@@ -677,20 +677,40 @@ def init_socket
...
+      @socket.close if @socket
+      socket = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 0)
       socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
       socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE, 1)
       socket
     end
 
+    def socket_connect 
+      address = Socket.getaddrinfo(@host, nil, Socket::AF_INET).first[3]
+      sockaddr = Socket.pack_sockaddr_in(@port, address)
```
mlucy: Why is this necessary?  Does `Socket.sockaddr_in(@port, @host)` not work for some reason?

drivers/ruby/lib/net.rb
mlucy: What's the reason why this is called by `init_socket` rather than `base_socket`?  Is it not correct to finish connecting the socket before `base_socket` returns?

mlucy:
Thanks for the pull request @icesmurf !  There are a few things I was unclear on in it.  Sorry for all the comments, but I'm a little paranoid about connection logic like this, and I'd like to leave as little magic in the code as possible.

drivers/ruby/lib/net.rb
icesmurf: agree, i'd fixed this already, not sure why it wasn't in the PR, must have messed up my squash

drivers/ruby/lib/net.rb
icesmurf: for some reason i thought it was better to do it individually, but i don't think it matters.

drivers/ruby/lib/net.rb
icesmurf: agree

drivers/ruby/lib/net.rb
icesmurf: heh, been trying to get my head around that as well actually, the examples i was looking at were terrible.

icesmurf:
Agree with mlucy on alot of his comments, i've thrown most of it out ahem :-) 

I think this method is ALOT simpler and more understandable at a glance, referenced from ruby "Socket" documentation, not sure if we need another connect_nonblock and rescue after, it seems redundant to me, but they do that in the Socket examples, thoughts?

danielmewes:
@mlucy Could you take another look at this please?

icesmurf:
Actually, hold off, i've noticed a few issues i'm looking at, esp in
consideration of next, i've had some issues using non-blocking sockets with
mass updates and am trying to figure out why.

srh:
Since this is old and the author says it has issues, closing.

PR #5518, opened 2016-03-14
JS Driver: add `concurrency` optarg to `eachAsync`

### LINKED ISSUE #5509: Add optarg to `eachAsync` for managing concurrency
Now that we have `eachAsync` for automatically awaiting async operations before continuing to read from the cursor, it would be nice if the concurrency could be controlled. This would essentially be equivalent to [`Promise.map(..., { concurrency: <int> })`](http://bluebirdjs.com/docs/api/promise.map.html#map-option-concurrency) from bluebird.

While this can be done in user-code with relative ease, I think it's useful enough (namely in tuning the performance of bulk operations) to consider implementing.

```js
const concurrency = 10;
const pending = [];

cursor.eachAsync((job) => {
  Promise.resolve().then(() => {
    if (pending.length >= concurrency) {
      return Promise.any(pending);
    }
  }).then(() => {
    let p = asyncHandler(job).then(() =>
      pending.splice(pending.indexOf(p), 1);
    );
    pending.push(p);
  });
});
```

Proposal:

```js
cursor.eachAsync(asyncHandler, { concurrency: 10 });
```

marshall007:
Fixes #5509. Note: this is on top of my changes for #5478 since they would conflict.

### DIFF
--- a/drivers/javascript/cursor.coffee
+++ b/drivers/javascript/cursor.coffee
@@ -224,24 +224,43 @@ class IterableResult
         return nextCb().nodeify(onFinished)
     )
 
-    eachAsync: varar(1, 2, (cb, errCb) ->
+    eachAsync: varar(1, 3, (cb, errCb, options = { concurrency: 1 }) ->
         unless typeof cb is 'function'
             throw new error.ReqlDriverError 'First argument to eachAsync must be a function.'
 
-        if errCb? and typeof errCb isnt 'function'
-            throw new error.ReqlDriverError "Optional second argument to eachAsync must be a function"
+        if errCb?
+            if typeof errCb is 'object'
+                options = errCb
+                errCb = undefined
+            else if typeof errCb isnt 'function'
+                throw new error.ReqlDriverError "Optional second argument to eachAsync must be a function or `options` object"
+
+        unless options and typeof options.concurrency is 'number' and options.concurrency > 0
+            throw new error.ReqlDriverError "Optional `options.concurrency` argument to eachAsync must be a positive number"
+
+        pending = []
+
+        userCb = (data) ->
+            return Promise.resolve(cb(data)) if cb.length <= 1 # either synchronous or awaits promise
+            return Promise.fromNode (handler) -> cb(data, handler) # callback-style async
 
         nextCb = =>
             if @_closeCbPromise?
                 return Promise.reject(new error.ReqlDriverError("Cursor is closed."))
             else
-                @_next().then (data) ->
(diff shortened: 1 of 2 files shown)


danielmewes:
Thanks @marshall007 !

Assigning to @nighelles for review.

drivers/javascript/cursor.coffee
```diff
@@ -211,50 +209,65 @@ class IterableResult
...
-                @_next nextCb
-            else
-                onFinished?()
-        @_next nextCb
+        nextCb = =>
+            @_next().then (data) ->
+                return nextCb() if cb(null, data) isnt false
+            .catch (err) ->
+                return if err?.message is 'No more rows in the cursor.'
+
```
nighelles: Do we lose the cb(err) call here?

drivers/javascript/cursor.coffee
```diff
@@ -211,50 +209,65 @@ class IterableResult
...
             if @_closeCbPromise?
                 return Promise.reject(new error.ReqlDriverError("Cursor is closed."))
             else
-                @_next().then (data) ->
-                    return cb(data) if cb.length <= 1 # either synchronous or awaits promise
-                    return Promise.fromNode (handler) -> cb(data, handler) # callback-style async
-                .then (result) ->
-                    return nextCb()
+                return @_next().then (data) ->
+                    return data if pending.length < options.concurrency
```
nighelles: If options is undefined, is this always false?

marshall007: `options.concurrency` will always either be what the user specified or `1` by the time we make it to this line.

nighelles: Ok, missed the default.

drivers/javascript/cursor.coffee
marshall007: Whoops, yea good catch.

larkost:
@marshall007: I have (finally) pushed my branch with an overhaul of the JavaScript connection tests in branch `larkost/5413-javascript-connection-tests`. A lot of this is about better defining our behavior, but there are a lot of failures, both with our current version of the driver, as well as your revision (largely the same issues). I know that @nighelles is looking into these areas, but it seems we are all working on intersecting changes.

nighelles:
Ok, now that these tests are up, I'll see about testing everything with theses new changes.

marshall007:
@larkost nice, this is great! I'll hold off on making any more changes so we don't step on each other's toes trying to get the tests to pass.

nighelles:
This was merged into next along with some other eachAsync fixes.

Issue #5378, opened 2016-02-09
Add option for simplified output of restore utility


gamename:
Slack conversation on the RethinkDB channel:

```
i guys.  Not an emergency, but is there some way to disable the % scale output of the rethink restore util?  I’m calling the utility from a script and the output is huge (see attached)
Or rather, the output makes the wrapper script’s logs huge.

new messages
neilhanlon [5:59 PM] 
>/dev/null
?
er...
>/dev/null
redirect stdio to /dev/null

tennis [6:12 PM] 
@neilhanlon: sure, but that leaves me completely blind.  I’m thinking of some compromise that will be informative, but not ascii art.

neilhanlon [6:19 PM] 
Only thing I can think of (besides actually modifying the code) is to pipe it into tr or something to replace the carriage return with a newline
https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/rethinkdb/_import.py#L601
 
GitHub
rethinkdb/rethinkdb
rethinkdb - The open-source database for the realtime web.

tennis [6:20 PM] 
ah, ok.  That’s a good compromise.  Thanks.

mglukhov [6:28 PM] 
@tennis: could you open an issue describing how you produced that?

[6:28] 
we can explore options there for how to improve the output, or provide a flag for suppressing it

tennis [6:29 PM] 
@mglukhov: Sure. Np. Thanks for working w/me on that.
```

gamename:
![pasted image at 2016_02_08 05_46 pm](https://cloud.githubusercontent.com/assets/42153/12904140/b99bbae8-ce92-11e5-83aa-aa8adfbce9b0.png)

gamename:
![pasted image at 2016_02_08 05_46 pm](https://cloud.githubusercontent.com/assets/42153/12904149/d0ccc9b4-ce92-11e5-8516-7f22b2f06672.png)

gamename:
The script which uses the restore util is a simple bash script:

```
...
if rethinkdb restore "${LOCAL_COPY}" \
   --shards "${CLUSTER_SIZE}"        \
   --replicas "${CLUSTER_SIZE}"      \
   --force; then
  log "${LOCAL_COPY} restoral Succeeded"
  rc=0
else
  log "${LOCAL_COPY} restoral FAILED"
fi
...
```

danielmewes:
So basically an option to disable the progress bar?

That should be pretty easy, and sounds useful for scripting purposes.

danielmewes:
[...]

PR #5528, opened 2016-03-15
rethinkdb-dump: Add --quiet and dumping to stdout

nh2:
This adds two features to `rethinkdb-dump`:

* A `--quiet` flag
* Writing the output archive to stdout

The second feature quite obviously depends on the first one, since we can't print messages and a tar archive to stdout at the same time.

The wording of the help option was inspired by `man tee`.

Example commands you can test:

* `rethinkdb-dump` - should work as before
* `rethinkdb-dump` - should work as before but print no output
* `rethinkdb-dump --file=out.tar.gz` - should work as before
* `rethinkdb-dump --file=-` - should print lots of rubbish into your terminal
* `rethinkdb-dump --file=- | tar tz` should print you the files in the tar file

### DIFF
--- a/drivers/python/rethinkdb/_dump.py
+++ b/drivers/python/rethinkdb/_dump.py
@@ -24,6 +24,7 @@ def print_dump_help():
     print("                                   to 3)")
     print("  --temp-dir DIRECTORY             the directory to use for intermediary results")
     print("  --overwrite-file                 don't abort when file given via --file already exists")
+    print("  -q [ --quiet ]                   suppress non-error messages")
     print("")
     print("EXAMPLES:")
     print("rethinkdb dump -c mnemosyne:39500")
@@ -45,6 +46,7 @@ def parse_options():
     parser.add_option("--temp-dir", dest="temp_dir", metavar="directory", default=None, type="string")
     parser.add_option("--overwrite-file", dest="overwrite_file", default=False, action="store_true")
     parser.add_option("--clients", dest="clients", metavar="NUM", default=3, type="int")
+    parser.add_option("-q", "--quiet", dest="quiet", default=False, action="store_true")
     parser.add_option("--debug", dest="debug", default=False, action="store_true")
     parser.add_option("-h", "--help", dest="help", default=False, action="store_true")
     (options, args) = parser.parse_args()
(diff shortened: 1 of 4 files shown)


danielmewes:
Thanks for another PR @nh2 !

I think we should add the `--quiet` option to `rethinkdb export`, `rethinkdb import` and `rethinkdb restore` as well. That would solve https://github.com/rethinkdb/rethinkdb/issues/5378 .

nh2:
> I think we should add the --quiet option to rethinkdb export, rethinkdb import and rethinkdb restore as well

OK, it's not late yet.

nh2:
I've added a comment `(note that intermediate files will still be written to the --temp-dir directory)` to the help text to make clear that the local disk will still be used even when dumping to stdout (this is important for security, e.g. when you use `--file=-` with the expectation that it never writes any files to disk).

nh2:
> I think we should add the --quiet option to rethinkdb export, rethinkdb import and rethinkdb restore as well.

@danielmewes Done.

nh2:
Another command that works now:

* `rethinkdb-dump --file=- | rethinkdb-restore - --force`

nh2:
I noticed that in the top-level one-line usages like

```python
usage = "rethinkdb restore FILE [-c HOST:PORT] [-a AUTH_KEY] [--clients NUM] [--shards NUM_SHARDS] [--replicas NUM_REPLICAS] [--force] [-i (DB | DB.TABLE)]..."
```

not all options are documented; consequently I've not added `--quiet` to them.

drivers/python/rethinkdb/_dump.py
```diff
@@ -68,13 +73,17 @@ def parse_options():
...
         res["temp_filename"] = "rethinkdb_dump_%s" % datetime.datetime.today().strftime("%Y-%m-%dT%H:%M:%S")
 
-    if options.out_file is None:
-        res["out_file"] = os.path.abspath("./" + res["temp_filename"] + ".tar.gz")
+    if options.out_file == "-":
+        res["out_file"] = sys.stdout
     else:
-        res["out_file"] = os.path.abspath(options.out_file)
+        # The output file is a real file in the file system
+        if options.out_file is None:
```
Tryneus: No reason to next these `if` statements, just use an `elif`

nh2:
@danielmewes @Tryneus I believe I have addressed all comments - ready to take another look?

danielmewes:
@nh2 Thanks for updating the PR and sorry for the delay. Pinging @Tryneus .

Tryneus:
Looks good to me, merging.  Thanks again, @nh2!

PR #5533, opened 2016-03-17
Add .swo to gitignore

aaronorosen:


larkost:
For those (like me) who don't/didn't know; `.swo` (and `.swp, `.swn`, etc..) files are created when you open a file for editing in `vim`. The first concurrent editor makes a `.swp` file, and then they progressively march down the alphabet starting with `.swo`.

Personally I think that stuff like this belongs in your own `git` preferences, but we already have a `*.swp` entry, so there is precedent.

aaronorosen:
I changed it to .sw? to ignore all the vim created files. I think having this here is helpful though I'm okay having it in my local git preferences.

danielmewes:
`.sw?` sounds good to me.
Thanks for the PR @aaronorosen .

Have you already signed the CLA http://rethinkdb.com/community/cla/ ?

aaronorosen:
yup i already have.

danielmewes:
Thanks!

PR #5545, opened 2016-03-19
Add `step` optarg to `r.range`

### LINKED ISSUE #3148: ReQL proposal: add an optarg `step` to `r.range`
Currently `r.range` as proposed in #875 always increments by one, which I propose to be specified through an optarg named `step`.

```
> r.range(step=10)
[0, 10, 20, ...]
> r.range(4, step=2).run
[0, 2]
> r.range(5, 2, step=-1).run
[5, 4, 3]
```

marshall007:
Fixes #3148.


danielmewes:
Thanks for the pull request @marshall007. I think this is your first C++ PR? :-)

I'll take a look at it later.

src/rdb_protocol/datum_stream.cc
```diff
@@ -2014,7 +2016,8 @@ range_datum_stream_t::next_raw_batch(env_t *, const batchspec_t &batchspec) {
                             batchspec.with_at_most(500).to_batcher();
 
     while (!is_exhausted()) {
-        double next = safe_to_double(start++);
+        double next = safe_to_double(start);
+        start+=step;
```
danielmewes: Per our usual style, there should be spaces around the operator: `start += step;`

src/rdb_protocol/datum_stream.cc
```diff
@@ -1991,11 +1991,13 @@ std::vector<changespec_t> union_datum_stream_t::get_changespecs() {
 range_datum_stream_t::range_datum_stream_t(bool _is_infinite_range,
                                            int64_t _start,
                                            int64_t _stop,
+                                           int64_t _step,
                                            backtrace_id_t bt)
     : eager_datum_stream_t(bt),
       is_infinite_range(_is_infinite_range),
```
danielmewes: We should also set `is_infinite_range` if `step` is 0.

src/rdb_protocol/datum_stream.hpp
```diff
@@ -404,6 +405,7 @@ class range_datum_stream_t : public eager_datum_stream_t {
         return false;
     }
     virtual bool is_exhausted() const;
+    virtual bool is_in_range() const;
```
danielmewes: This doesn't need to be `virtual`. The other functions are virtual only because they're overwriting functions declared in `eager_datum_stream_t`.

src/rdb_protocol/datum_stream.cc
danielmewes: Oh I see you're checking for 0 below in the range_term_t. I wonder if we should allow `0`. It could be useful if you want an infinite sequence of one number. Either way should be fine.

marshall007: I'm throwing a logic error [here](https://github.com/rethinkdb/rethinkdb/pull/5545/files#diff-c13480669499466d2358b2c0cd04c327R798) when `step == 0`, would an infinite stream be preferable?

src/rdb_protocol/terms/seq.cc
```diff
@@ -787,11 +787,17 @@ class zip_term_t : public op_term_t {
...
 private:
     virtual scoped_ptr_t<val_t> eval_impl(scope_env_t *env, args_t *args, eval_flags_t) const {
         bool is_infinite = false;
-        int64_t start = 0, stop = 0;
+        int64_t start = 0, stop = 0, step = 1;
+
+        if (scoped_ptr_t<val_t> maybe_step = args->optarg(env, "step")) {
+            step = maybe_step->as_int();
+            rcheck_target(maybe_step, step != 0, base_exc_t::LOGIC,
+                          "Expected non-zero NUMBER.");
```
danielmewes: The error message should also mention that this is for the step argument.
E.g.:
```c++
rcheck_target(maybe_step, step != 0, base_exc_t::LOGIC,
                       "`step` cannot be zero.");
```

drivers/javascript/ast.coffee
```diff
@@ -1355,7 +1355,10 @@ rethinkdb.circle = aropt (cen, rad, opts) -> new Circle opts, cen, rad
 
 rethinkdb.uuid = (args...) -> new UUID {}, args...
 
-rethinkdb.range = (args...) -> new Range {}, args...
+rethinkdb.range = (args...) ->
+    if args.length and typeof args[args.length - 1] is 'object'
```
danielmewes: On other places we're using this check:
```js
((Object::toString.call(args[args.length - 1]) is '[object Object]') and not (args[args.length - 1] instanceof TermBase))
```

This might make a difference if the `step` argument is a ReQL term itself. Though I haven't tested it.

danielmewes:
I added a few comments, but looks good mostly.

Are you planning to add support in the other drivers as well? We can take over that part if you prefer.

danielmewes:
Just saw that there was some discussion in https://github.com/rethinkdb/rethinkdb/issues/3148 on whether this option should be called `stride` or `step`. I personally like `step` better. I'll ask a few people around here for what they think on Monday.

src/rdb_protocol/datum_stream.cc
marshall007: Ah, didn't see your comment before I posted mine. Yea, I thought about allowing `0` too but realized the usage is a little janky since you'd have to specify the meaningless `stop` argument to get the number you want: `r.range(1, 0, { step: 0 })`. Let me know what you think.

marshall007:
> I think this is your first C++ PR? :-)

@danielmewes yep! :D I've made updates based on your comments and refactored all those separate optarg checks in the JS driver into `util.isPlainObject(obj)`. I replied to your comment regarding the behavior when `step == 0`. Just let me know if you think we should support that.

> Are you planning to add support in the other drivers as well? We can take over that part if you prefer.

Would be great if someone else could take over, thanks!

danielmewes:
Thanks for updating the PR and for the `isPlainObject` refactor.

Let's keep disallowing `step == 0`. It avoids potentially unexpected behavior and allowing something later is always easier than the other way around (with respect to backwards compatibility).

The only remaining question is the name of the option. I'll talk to some people to figure this out and let you know later today.

danielmewes:
@marshall007 After talking to some people, we settled on `step`.

The server side of this looks good to me. I haven't reviewed the JavaScript driver changes yet.

marshall007:
@danielmewes sweet! I went ahead and updated the Python driver too.

srh:
What's left is to do the Ruby and Java drivers.  Working on that now...

srh:
Pointing this at `v2.4.x` in preparation for making that branch the main branch of the repo.  I intend to look at this post-2.4.2 release.

srh:
- [ ] the JS/Python driver changes (and Ruby/Java) need to be moved to (or made in) their respective repositories
- [ ] the JS driver for the admin UI will need to get updated (in branch `old_admin`) in a later PR

PR #5596, opened 2016-04-04
feat(auto_reconnect): allow runtime configuration of give_up_ms

mbroadst:
Currently the auto_reconnect instance used to reconnect lost nodes is hardcoded to give up reconnecting after 24 hours. This is not ideal in some scenarios where a user may want to remove a node from a cluster, without having to reset all other participating nodes in the cluster. This patch allows the user to change that value.

### DIFF
--- a/src/clustering/administration/main/command_line.cc
+++ b/src/clustering/administration/main/command_line.cc
@@ -58,6 +58,10 @@
 #define RETHINKDB_RESTORE_SCRIPT "rethinkdb-restore"
 #define RETHINKDB_INDEX_REBUILD_SCRIPT "rethinkdb-index-rebuild"
 
+namespace cluster_defaults {
+    const int reconnect_timeout = (24 * 60 * 60 * 1000);    // 24 hours
+}  // namespace cluster_defaults
+
 MUST_USE bool numwrite(const char *path, int number) {
     // Try to figure out what this function does.
     FILE *fp1 = fopen(path, "w");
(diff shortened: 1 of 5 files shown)


mbroadst:
**NOTE:** I'm not attached to any of the naming here, and eagerly look forward to your review :smile:

danielmewes:
Thanks for the pull request @mbroadst , very nice.

I'll double-check the code shortly. Hope we can include it in 2.3.1 then :)

mbroadst:
@danielmewes great, thanks!

I'd also like to potentially discuss a somewhat more involved future solution for this problem. Would it be out of the question to involve a handshake of sorts on these auto reconnects? Basically, the auto reconnecter would reconnect to a node and verify that the node name is the same in order to complete the reconnect.  I'm just sort of spitballing here, but thought I'd get the conversation started in terms of thinking about why this might/might not be a good idea.

mbroadst:
ping

danielmewes:
Was out of office and didn't get around to this before I left. Going to pick this up again in the next days.

danielmewes:
Ok your patch looks great to me.

The one thing I'd like to see changed is accepting the parameter in seconds, rather than milliseconds. Please make sure you properly check for `int` overflows when converting back to `ms`. Using seconds is more consistent with our other parameters and more useful in practice I think.

Regarding comparing the node name: We should actually compare the server ID rather than the name. I thins this might already happen in the handshake (at least the server IDs are exchanged), but the information isn't propagated up to the `auto_reconnector_t` so it doesn't know that it should stop trying. It sounds completely reasonable though to stop reconnecting to an address from the `auto_reconnector_t` once we find out that the server ID isn't what we were expecting. I can give you some more pointers if you'd like to look into this. Basically this is where we receive the remote server ID https://github.com/rethinkdb/rethinkdb/blob/next/src/rpc/connectivity/cluster.cc#L961 , but we don't actually know in this function which ID we were expecting. The function has an `expected_id`, but that is a `peer_id_t` which changes every time a server is restarted and hence doesn't help for the auto reconnector (since it should try to reconnect even after another server got restarted). I'm not sure if we would still need the expected peer ID check if we added an expected server ID check.

mbroadst:
@danielmewes great, I'll take some time tomorrow to clean up the patch and will look into the handshake

mbroadst:
closed in favor of #5701

PR #5609, opened 2016-04-05
issue-5608: Fix Python driver slice bug in 2.2.x

binarycrayon:
- Fix a bug by using itertools.islice to slice deque used in Cursor


danielmewes:
Closing since we already have a fix for this in `next`. Thanks for the pull request though!

PR #5618, opened 2016-04-06
Cursor now supports timeout in getNext()

iorlas1:
Timeout was previously ignored - now actually used. Hopefully solves: #5603

PR #5640, opened 2016-04-11
Java Client: handle cursor timeout when calling cursor.next(timeout)

QianJin2013:
Hi, i found cursor.next(timeout) is not implemented yet (it is just defined there but ignored timeout parameter)

So i'v modified Cursor.java to implement it mainly by
```
 res = this.awatingContinue.get();
```
to:
```
 res = this.awatingContinue.get(timeout, TimeUnit.MILLISECONDS);
``` 

Teste case testQueryTimeout has also been added. All other test cases works well.

Hope you merge it.

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/annotations/IgnoreNullFields.java
+++ b/drivers/java/src/main/java/com/rethinkdb/annotations/IgnoreNullFields.java
@@ -1,0 +1,15 @@
+package com.rethinkdb.annotations;
+
+import java.lang.annotation.*;
+
+@Target(ElementType.TYPE)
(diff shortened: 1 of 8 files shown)


QianJin2013:
This is necessary when subscribe change notification with timeout.

```
Cursor cursor = r.db("test").table("test").changes().run(con);
//in fact, following statement will not throw timeout exception after 10 seconds.
Map<String, Object> res = (Map<String, Object>)cursor.next(10*1000/*10seconds*/);  
```

deontologician:
This was fixed in #5618, it looks like there are other changes as well (I am guessing they are the changes in #5641)

PR #5648, opened 2016-04-12
Fix for #5647

### LINKED ISSUE #5647: Rethinkdb-restore fails in 2.3.0 with Python 3.4
I get this error while running RethinkDB 2.3.0, python package 2.3.0 and `Python 3.4.2 (default, Oct  8 2014, 10:45:20) [GCC 4.9.1] on linux` inside a debian:jessie docker container.
```
Unzipping archive file...
  File "/usr/local/bin/rethinkdb-restore", line 9, in <module>
    load_entry_point('rethinkdb==2.3.0', 'console_scripts', 'rethinkdb-restore')()
  File "/usr/local/lib/python3.4/dist-packages/rethinkdb/_restore.py", line 266, in main
    run_rethinkdb_import(options)
  File "/usr/local/lib/python3.4/dist-packages/rethinkdb/_restore.py", line 248, in run_rethinkdb_import
    sub_dir = do_unzip(temp_dir, options)
  File "/usr/local/lib/python3.4/dist-packages/rethinkdb/_restore.py", line 150, in do_unzip
    is_fileobj = type(options["in_file"]) is file
NameError: name 'file' is not defined
```

NeilHanlon:
Fixes #5647 

Python 3's sys.stdin is not of type "file" anymore, but rather
_io.TextIOWrapper -- which parses down to just type Class (apparently)

This should fix python3, but also work in python2.

This might be a naive approach, but I don't really see another option. :fearful:

### DIFF
--- a/drivers/python/rethinkdb/_restore.py
+++ b/drivers/python/rethinkdb/_restore.py
@@ -147,7 +147,7 @@ def do_unzip(temp_dir, options):
         path, base = os.path.split(path)
         return (base, db, os.path.splitext(table_file)[0])
 
-    is_fileobj = type(options["in_file"]) is file
+    is_fileobj = type(options["in_file"]) is type(sys.stdin)
 
     # If the in_file is a fileobj (e.g. stdin), stream it to a seekable file
     # first so that the code below can seek in it.
--- a/drivers/python/rethinkdb/_restore.py
+++ b/drivers/python/rethinkdb/_restore.py
@@ -147,12 +147,10 @@ def do_unzip(temp_dir, options):
         path, base = os.path.split(path)
         return (base, db, os.path.splitext(table_file)[0])
 
-    is_fileobj = type(options["in_file"]) is type(sys.stdin)
-
     # If the in_file is a fileobj (e.g. stdin), stream it to a seekable file
     # first so that the code below can seek in it.
     tar_temp_file_path = None
-    if is_fileobj:
+    if options["in_file"] is sys.stdin:
         fileobj = options["in_file"]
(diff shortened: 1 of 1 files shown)


drivers/python/rethinkdb/_restore.py
```diff
@@ -147,7 +147,7 @@ def parse_path(member):
         path, base = os.path.split(path)
         return (base, db, os.path.splitext(table_file)[0])
 
-    is_fileobj = type(options["in_file"]) is file
```
larkost: For this case I think it would be a lot clearer to do this: `options["in_file"] is sys.stdin`. That does not allow for other open files, but in this case we don't allow for that anyways, and would need some more work to avoid copying the file anyways. Additionally I would rather see this line folded into line 155, as that would make it a more linear read.

NeilHanlon: Very good point. I'll make that change!

NeilHanlon:
This look good @larkost ?

larkost:
Perfect! I don't see a previous contribution from you (easy to get wrong though), so have you signed our [CLA](http://rethinkdb.com/community/cla/)? With that I will merge this in. Thanks again!

larkost:
@NeilHanlon never mind, I just confirmed internally that you have signed it, so merging now!

danielmewes:
@larkost Could you also cherry-pick this into `v2.3.x` please?

danielmewes:
Cherry-picked into v2.3.x as e6e6b2b.

snackycracky:
please release this 👍

PR #5649, opened 2016-04-12
Support for connection pooling in Java driver

PeterKnego:
Adds support for an extensible connection pool implementations via a ConnectionPool interface. 

Provides a FixedConnectionPool implementation, which creates connections on-demand to a pre-defined set of cluster nodes. Connections are selected in a round-robin fashion and after use returned to the pool. Pool grows on-demand by creating new connections when existing pool is exhausted. 

Typical use:
```
// define conneciton builders 
Connection.Builder connBuilder1 = Connection.build().hostname("localhost").port(28015);
Connection.Builder connBuilder2 = Connection.build().hostname("localhost").port(28016);

// create a conneciton pool
ConnectionPool pool = new FixedConnectionPool(connBuilder1, connBuilder2);

// use it
RethinkDB.r.db("test").table("pooltest").run(pool);  // provide ConnectionPool instead of Connection

//  optionally register a global connection pool 
RethinkDB.setGlobalConnectionPool(pool);  
RethinkDB.r.db("test").table("pooltest").run();    // no need to pass pool or conn

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/ast/ReqlAst.java
+++ b/drivers/java/src/main/java/com/rethinkdb/ast/ReqlAst.java
@@ -1,10 +1,12 @@
 package com.rethinkdb.ast;
 
+import com.rethinkdb.RethinkDB;
 import com.rethinkdb.gen.exc.ReqlDriverError;
 import com.rethinkdb.gen.proto.TermType;
 import com.rethinkdb.model.Arguments;
 import com.rethinkdb.model.OptArgs;
 import com.rethinkdb.net.Connection;
+import com.rethinkdb.net.ConnectionPool;
 import org.json.simple.JSONArray;
 
 import java.util.HashMap;
@@ -65,6 +67,41 @@ public class ReqlAst {
         return conn.run(this, new OptArgs(), Optional.empty());
     }
 
+	  /**
+	   * Runs this query via connection {@code conn} provided by connection pool, with default options and returns an atom result
+	   * or a sequence result as a cursor. The atom result either has a primitive type (e.g., {@code Integer})
+	   * or represents a JSON object as a {@code Map<String, Object>}. The cursor is a {@code com.rethinkdb.net.Cursor}
+	   * which may be iterated to get a sequence of atom results
+	   * @param pool The connection pool to provide the connection
+	   * @param <T> The type of result
+	   * @return The result of this query
+	   */
+	  public <T> T run(ConnectionPool pool) {
+		    Connection conn = pool.leaseConnection();
+		    T res = this.run(conn);
+		    pool.returnConnection(conn);
+		    return res;
+	  }
+
+	  /**
+	   * Runs this query via connection {@code conn} provided by the internal global connection pool,
+	   * with default options and returns an atom result
+	   * or a sequence result as a cursor. The atom result either has a primitive type (e.g., {@code Integer})
+	   * or represents a JSON object as a {@code Map<String, Object>}. The cursor is a {@code com.rethinkdb.net.Cursor}
+	   * which may be iterated to get a sequence of atom results.
+	   * <p/>Global connection pool must be defined prior to using this metod.
+	   * @param <T> The type of result
+	   * @return The result of this query
+	   */
+	  public <T> T run() {
+	  	  ConnectionPool globalPool = RethinkDB.getGlobalConnectionPool();
+		    if (globalPool == null) {
+			    throw new RuntimeException("Global connection pool must be defined via RethinkDB.setGlobalConnectionPool() " +
+			  		  "prior to calling RethinkDB.run()");
+  		  }
+		    return run(globalPool);
+	  }
+
     /**
      * Runs this query via connection {@code conn} with options {@code runOpts} and returns an atom result
      * or a sequence result as a cursor. The atom result either has a primitive type (e.g., {@code Integer})
@@ -79,7 +116,44 @@ public class ReqlAst {
         return conn.run(this, runOpts, Optional.empty());
     }
 
+	  /**
+	   * Runs this query via connection {@code conn} provided by connection pool,
+	   * with options {@code runOpts} and returns an atom result
+	   * or a sequence result as a cursor. The atom result either has a primitive type (e.g., {@code Integer})
+	   * or represents a JSON object as a {@code Map<String, Object>}. The cursor is a {@code com.rethinkdb.net.Cursor}
+	   * which may be iterated to get a sequence of atom results
+	   * @param pool The connection pool to provide connection to run this query
+	   * @param runOpts The options to run this query with
+	   * @param <T> The type of result
+	   * @return The result of this query
+	   */
+	  public <T> T run(ConnectionPool pool, OptArgs runOpts) {
+		    Connection conn = pool.leaseConnection();
+		    T res = this.run(conn, runOpts);
+		    pool.returnConnection(conn);
+		    return res;
+	  }
+
     /**
+     * Runs this query via connection {@code conn} provided by global connection pool,
+     * with options {@code runOpts} and returns an atom result
+     * or a sequence result as a cursor. The atom result either has a primitive type (e.g., {@code Integer})
+     * or represents a JSON object as a {@code Map<String, Object>}. The cursor is a {@code com.rethinkdb.net.Cursor}
+     * which may be iterated to get a sequence of atom results
+     * @param runOpts The options to run this query with
+     * @param <T> The type of result
+     * @return The result of this query
+     */
+    public <T> T run(OptArgs runOpts) {
+    	  ConnectionPool globalPool = RethinkDB.getGlobalConnectionPool();
+    	  if (globalPool == null) {
+    	  	throw new RuntimeException("Global connection pool must be defined via RethinkDB.setGlobalConnectionPool() " +
+    	  			"prior to calling RethinkDB.run()");
+    	  }
+    	  return run(globalPool, runOpts);
+    }
+
+  	/**
      * Runs this query via connection {@code conn} with default options and returns an atom result
      * or a sequence result as a cursor. The atom result representing a JSON object is converted
      * to an object of type {@code Class<P>} specified with {@code pojoClass}. The cursor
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by deontologician** [comment]
This is beautiful :) Thanks for the PR, have you signed the CLA already?

File: drivers/java/src/main/java/com/rethinkdb/ast/ReqlAst.java [hunk 1]
```diff
@@ -96,6 +170,47 @@ protected Object build() {
...
+     * @param pojoClass The class of POJO to convert to
+     * @param <T> The type of result
+     * @param <P> The type of POJO to convert to
+     * @return The result of this query (either a {@code P or a Cursor<P>}
+     */
+    public <T, P> T run(ConnectionPool pool, Class<P> pojoClass) {
+    	  Connection conn = pool.leaseConnection();
+    	  T res = this.run(conn, pojoClass);
+    	  pool.returnConnection(conn);
+    	  return res;
```
bchavez: Hey there! Great work and contribution! This is pretty awesome! Just had a couple of questions here...

What happens when `this.run(conn, pojoClass)` throws an underlying exception? ReQL syntax `ReqlDriverError`? Seems like a connection could never be returned to the connection pool. Possible memory/connection leak?

Patterns like this (1. take connection, 2. use, 3. put back connection) puts on the onus on the caller of the pool to remember to put back a connection after usage (just my personal feeling of uneasiness here). I think you'll have to ultimately wrap `try{}catch{}finally{}` on each of these `run` statements to make this pattern work.

I have a few more comments but I'm currently on vacation right now but memory/connection leak is the most critical of my questions.

The second without much comment is the unboundedness of a "fixed connection pool" and weather or not all the orchestration (lease / use / return) is necessary when each connection is already thread-safe. Just use with a round-robin counter?

Sorry, I have to go!
Brian

PeterKnego: `ReqlDriverError` is an unchecked exception extending `RuntimeException`, so user code can not really anticipate it or prepare to handle it. The docs and/or examples do not show any required handling in case of this exception (e.g. closing `Connection`).

Normally the connection pool wrapper code handles the lease-use-return pattern, except in cases of exception, which normally break the execution of current thread.

Java hint: unchecked exceptions (e.g. the ones extending `RuntimeException`) are the ones that you can not expect the calling code to recover from. If there is an anticipated or expected way to recover from the exception it should be a checked exception, where calling code is forced to deal with (at compile time).

`ConnectionPool` implementations pass this unchecked exception to the user code, the same as non-connection-pool methods.

Note: in case of exception the `Connection` is not returned back to connection pool and is subject to garbage collection.

bchavez:
Hey there Peter...

IIRC, the driver can throw an exception based on a number of error responses the driver receives from the server [`ErrorBuilder.java`](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/ErrorBuilder.java#L44) [`Connection.java`](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L260). If the driver throws one of these server response exceptions, the underlying TCP connection and `Connection` can be re-used for a subsequent query (in this sense, I guess I see the `Connection` itself in a recoverable state after throwing a server response exception client-side). As you've pointed out, perhaps maybe `RuntimeException` isn't the best base class for the exception hierarchy. I don't know since I'm not a Java expert.

But I think the main concern I see is:

```
// use it
RethinkDB.r.db("test").table("pooltest").run(pool);  // provide ConnectionPool instead of Connection
```
and the implementation:
```
public <T> T run(ConnectionPool pool) {
    Connection conn = pool.leaseConnection();
    T res = this.run(conn);
    pool.returnConnection(conn);
    return res;
}
```
`conn` is leased, a `ReqlError` server response exception is thrown inside `this.run(conn)`, `conn` never gets `pool.returnConnection(conn)` returned. There's no way from the `use it` perspective to "put `conn` back" in the `pool`. `conn` doesn't necessarily need to be closed because a previous query by a different thread could potentially be holding a reference to a `Cursor`. IIRC, a `Cursor` itself holds a reference to the underlying `Connection` in which it was created from.

This is why I have some reservations about the lease-use-return pattern here because a 'lease' in the pool isn't an exclusive connection lease due to a foreign thread possession of a `Cursor` object (and thus underlying `Connection` which could be used at any time in a request for more `Cursor` items).

Additionally, closing `conn` immediately upon an exception, would probably not be the right thing since it would interrupt other threads holding `Cursor` objects associated with the same connection, but "putting it back" (less TCP connectivity issues) would be. Also, I'm thinking simply abandoning a `conn` reference like this and waiting for GC to collect lost references might lead to regressions similar to #5448.

Also `Connection`s are kind of expensive to create and bring online. In the event of an exhausted `pool` a query could take as much time to boot up a new connection when `Connection` reuse could offer better performance.

-------

I was thinking about the `leaseConnection` algorithm and I think there's an edge case that could possibly lead to a lopsided connection pool.

Let's consider,

```
// define conneciton builders 
Connection.Builder connBuilderA = Connection.build().hostname("ServerA").port(28015);
Connection.Builder connBuilderB = Connection.build().hostname("ServerB").port(28016);

// create a conneciton pool
ConnectionPool pool = new FixedConnectionPool(connBuilderA, connBuilderB);
```

Initially, **ServerA** is offline. `leaseConnection()` is called multiple times. Some calls are against an exhausted `pool`, the `pool` accumulates several `Connection`s to **ServerB**. **ServerA** comes online. The next chance for **ServerA** to process a request is the next time the pool is exhausted? In the worst case, could be never? This kinda goes with my the unboundedness thoughts I had earlier. 

-----

I hope I don't discourage your PR. Your PR is a great step forward for Java. I just wanted to share with you my personal experience implementing a connection pool for RethinkDB.

Thanks,
Brian

PeterKnego:
Hi Brian,

thanks for pointing out all the details about the Connection and sharing your experiences. It does not discourage me at all and I'll try to find a way to resolve the issues you pointed out.

1. As I understand there is no way for user code (or ConnectionPool wrapper) to know if an exception thrown by Connection can be recovered from? The logical solution in this case is to discard the connection and not return it to pool. 
(In reality, non-checked exception can not be anticipated and will break the execution of the current thread. Normally in the server-side code a generic exception wrapper will catch such exceptions and just produce a generic error response)
So this is more of a general situation that affects all code using Connection (userland code and/or ConnectionPool) and needs to be cleared by Connection developers: 
- Are there any exceptions thrown by Connection that are recoverable? If yes, than they should be expected by user code and should be a checked exception (forcing user at compile time to deal with this possible situation).
- Can Connection continue to be used after such exception? If yes, then they should probably not be an exception, but an error code. 
- Can en exception happen that affects just one Cursor and others Cursors can happily continue working? If yes, then the exception should be thrown on the Cursor, or the Cursor instead just produces the error code.

At the moment ConnectionPool behaves the same as user code - it stops using the Connection after an exception, as it has no way of knowing 

2. About possible resources leak: should all uses of Connection `try-catch` exceptions and `.close()` the Connection? This is not indicated in docs/examples.

3. Good point about accumulation/exhaustion if one node goes down for a while. The proper way of handling this would be to have a set of connection pools, one for each node. The lease logic would then round-robin through the set of connection pools, skipping pools that can not produce a working Connection (maybe with some time-based back-off algorithm for unavailable nodes). When node is again available, the logic would instantly (depending on the back-off) start producing Connections for it.

bchavez:
Hi Peter,

I'll try my best to answer your questions. Forgive me if I accidentally get something wrong, I don't have my IDE with me at the moment.

> 3. Good point about accumulation/exhaustion if one node goes down for a while. The proper way of handling this would be to have a set of connection pools, one for each node. The lease logic would then round-robin through the set of connection pools, skipping pools that can not produce a working Connection (maybe with some time-based back-off algorithm for unavailable nodes). When node is again available, the logic would instantly (depending on the back-off) start producing Connections for it.

So, my gut feeling here is that using a lease-use-return pattern here is sort of antithetical to the design of the `Connection` object. `Connection` objects themselves are thread-safe, designed, and meant to be used by multiple threads concurrently regardless of the results and responses they produce. For example, five threads holding the same `Connection` object reference can call `.run` at virtually the same time without interfering with each other. Internally, the synchronization primitives inside `Connection` will ensure queries will get serialized individually, in whole, in some order, over the wire. Query responses from the server can be parsed in any order and awaiting threads in any order will be notified individually when their query response is ready: whether or not that response is a `ReqlError` or a legitimate data response.

When a network I/O error occurs (ie: someone unplugs a cable or shuts down a server), all awaiting threads with pending responses on that particular faulted `Connection` are notified with an `IOException` and the `Connection` is internally shutdown and `.closed`. When the `Connection` is in this state, the `Connection` is non-recoverable and must be manually `.reconnected`.

Absent of network I/O error exceptions, the `Connection` can be re-used without problems. `ReqlError` exceptions are channeled to the corresponding thread to be dealt with however they see fit.

So, when we think about lease-use-return patterns like what we've discussed here, what we're really doing is introducing an extra layer of synchronization on top of an already existing internal layer synchronization primitives inside `Connection`, which, in my view, seems unnecessary. Our lease-use-return pattern is really only adding extra overhead and latency to a query with no substantial benefit.

A better "pooling" pattern for `Connection`, I think, is a select-use pattern. Make a `Connection` selection and use it. Don't worry about returning it back to the `pool`. Select-use removes the responsibility of the caller to return the `Connection` to the `pool`. Removes the possibility of regressing in #5448. We don't have to depend on a GC that we can't control. We've minimized the number of locks we take on the way down to the wire. Last but not least we've eliminated the concept of 'exhaustion'. All wins I think. But how? Some pseudo code, in broad strokes:

```
class HostEntry{
  Connection Connection;
  bool IsFaulted;
  DateTime NextRetry;
}

class HostPool{

  HostEntry[] HostArray;
  AtomicInteger next;

  selectConnection(){
    do{
       var index = next.incrementAndGet();
       var entry = HostArray[ index % HostArray.Length ];
    } while( !entry.IsFaulted )
    return entry;
  }
  
  addHost(Connection conn){
    var updatedHostArray = new HostArray[HostArray.Length + 1];
    Array.Copy(HostArray, updatedHostArray);
    
    var newHostEntry = new HostEntry(conn);
    updatedHostArray[LAST] = newHostEntry;
    HostArray = updatedHostArray;
  }
  
}
```

Some precautions like making sure we don't get stuck in the loop when all nodes are down... etc.

Select-use pattern invites the question: Does it make sense to maintain multiple connections to the **same** server from the **same** process? The answer: Maybe, maybe not. Last I can recall from a conversation I had was that multiple client connections to the same server can offer only *slightly* better performance on a multi-core RethinkDB server. Is it worth it? I don't know. It's hard to say without some hard numbers. My gut feeling here is that it may not be worth the cost of extra complexity and maintenance. If I had to hazard a guess, there's probably a point of diminishing returns. That point would converge even faster with multiple connections to the same server especially when multiple application servers are involved (`N app servers x N connections per server` against a RethinkDB server[or cluster]). Some empirical numbers on this subject would be really nice though.

Now, how do we know when a `Connection` `IsFaulted`? Ah... the $10,000 question. A few ways maybe...

Option A: Using exceptions as a control signal (yes, I know, the code smell is real and ugly, but works):
```
public <T> T run(ConnectionPool pool) {
    var hostEntry = pool.selectConnection();
    var conn = hostEntry.Connection;
    try{
      T res = this.run(conn);
      return res;
    }
    catch(IOException ex){
      //network fault
      pool.markFaulted(hostEntry);
      rethrow ex;
    }
    catch(Exception e){
      rethrow e;
    }
}
```
and the debate on weather or not using exceptions as control signals is a good idea in practice rages on... :smiley_cat: 

Option B: Bubble up the `.close` event (or via the `Connection.ResponsePump`). Basically, modify `Connection` to allow the registration of some lambda callbacks that will be called when a fault is detected. The pool would register a callback with the connection when it's setup/connected. When a `Connection` fault is detected we instantly know to exclude the faulted node from selection by registering `pool.markFaulted` in our callback.


> 2. About possible resources leak: should all uses of `Connection` try-catch exceptions and `.close()` the `Connection`? This is not indicated in docs/examples.

I'm not sure. Depends on your application and whether or not you favor fail-early-fail-fast or prefer defensively guarding against query and network errors.

-----

I *hope* I've answered the questions below with my explanation above.

> 1. As I understand there is no way for user code (or ConnectionPool wrapper) to know if an exception thrown by Connection can be recovered from?

> Are there any exceptions thrown by Connection that are recoverable? If yes, than they should be expected by user code and should be a checked exception (forcing user at compile time to deal with this possible situation).

> Can Connection continue to be used after such exception? If yes, then they should probably not be an exception, but an error code.

> Can en exception happen that affects just one Cursor and others Cursors can happily continue working? If yes, then the exception should be thrown on the Cursor, or the Cursor instead just produces the error code.

-----

And finally, some helpful links if you're curious:

Java connection is now thread-safe
https://github.com/rethinkdb/rethinkdb/pull/5292

Specifically, section on **Regarding Connection Pooling**
https://github.com/rethinkdb/rethinkdb/pull/5292#issuecomment-172261625

The design notes for .NET Connection Pooling here:
https://github.com/bchavez/RethinkDb.Driver/issues/17

Associated documentation for .NET Connection Pooling:
https://github.com/bchavez/RethinkDb.Driver/wiki/Connections-&-Pooling

The .NET driver implementation is here:
https://github.com/bchavez/RethinkDb.Driver/tree/master/Source/RethinkDb.Driver/Net/Clustering


My particular implementation is heavily optimized for selection performance so I'm bending OO-SOLID rules. I think I've done both Option A and B. We could probably just get by with option B honestly.

I hope this helps,
Brian

PeterKnego:
Hi Brian,

great writeup. I did not fully realise that Connection is designed to be long-lived and used by multiple threads. 

I took a look at your C# driver and it's ConnectionPool implementation - I think it's the proper way to implement connection pool.

I'm closing this PR and will consider implementing a Connection pool in line with the C# one.

Samistine:
This is still needed

PR #5652, opened 2016-04-12
indent auto-generated java files by modify template

QianJin2013:
Some auto-generated methods are not indented, so i have 
modified drivers/java/templates/AstSubclass.java, 
then run make update-driver to regenerate gen/*.java

Very simple fix.


deontologician:
Looks good, thanks :+1:

PR #5484, opened 2016-03-05
Support almost ALL types in/out. (e.x. Enum, OffsetDateTime, BigDecimal, double, Float... )

QianJin2013:
### [Java-Client-Driver] There are many data type can not be store to or retrieve from db just because the type conversion logic have some problems or not enough.

These problematic types are (23 types):
```
enum
BigDecimal, BigInteger
Boolean, boolean
Double, double, Float, float, Long, long, Integer, int, Short, short, Byte, byte
OffsetDateTime, ZonedDateTime, LocalDate, LocalDateTime, LocalTime, Date
```

In above types, there are some types cause conversion problem when store data to db, both implemented in com.rethinkdb.ast.Util::toReqlAst()
```
enum:
   No error, but stored as null.
```
```
Integer, int:
  Strange unreasonable exception complains should use Long instead of Integer.
com.rethinkdb.gen.exc.ReqlDriverError: Can't convert com.rethinkdb.TestPojo@79b06cab to a ReqlAst: Make getIntegerProperty of com.rethinkdb.TestPojo@79b06cab Long instead of Integer

	at com.rethinkdb.ast.Util.toMap(Util.java:159)
	at com.rethinkdb.ast.Util.toReqlAst(Util.java:120)
	at com.rethinkdb.ast.Util.toReqlAst(Util.java:33)
	at com.rethinkdb.model.Arguments.coerceAndAdd(Arguments.java:43)
	at com.rethinkdb.gen.ast.Table.insert(Table.java:45)
	at com.rethinkdb.RethinkDBTest.testTableSelectOfPojoCursor(RethinkDBTest.java:293)
```
```
LocalDateTime:
  Error when try to format non-existence timezone. Error message is weird, seems non sense.
java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: OffsetSeconds

	at java.time.LocalDate.get0(LocalDate.java:680)
	at java.time.LocalDate.getLong(LocalDate.java:659)
	at java.time.LocalDateTime.getLong(LocalDateTime.java:720)
	at java.time.format.DateTimePrintContext$1.getLong(DateTimePrintContext.java:205)
	at java.time.format.DateTimePrintContext.getValue(DateTimePrintContext.java:298)
	at java.time.format.DateTimeFormatterBuilder$OffsetIdPrinterParser.format(DateTimeFormatterBuilder.java:3338)
	at java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(DateTimeFormatterBuilder.java:2182)
[...]

Issue #5128, opened 2015-11-21
Look into using Jackson for java driver json serialization


deontologician:
We now have several corner cases in the pojo <-> json code that it may make sense to look at a library that's already dealt with those corner cases.

pires:
Jackson or Gson are good choices. I prefer the latter because of the [support for polymorphism](https://futurestud.io/blog/how-to-deserialize-a-list-of-polymorphic-objects-with-gson/).

kclay:
@pires fyi Jackson supports polymorphism ad well by the use of annotations http://stackoverflow.com/a/30386694

pires:
Warning: Don't use `is` prefix for boolean getters, e.g. `isConnected()` for accessing `boolean connected`. **Always** use `get` prefix, e.g. `getConnected()` for accessing `boolean connected`.

dbolotin:
+1

I think Jeckson is a better option here, as it is used as a core JSON module in two most popular reactive web frameworks in Java world: Vert.x and Play. So users of these frameworks will be able to use the same annotated classes to store in DB and send/receive over http/websockets.

Anyway, I also agree with @kclay, that some simple interchangeable abstraction here would be a nice option.

... it looks like in current architecture maps are used as an intermediate objects, and pojos are constructed from maps rather than raw JSON, so incorporation of any JSON framework will not be a trivial task... However, I don't think that simple map to pojo converter (com.rethinkdb.net.Util#toPojo) as implemented now, can satisfy demand for "ORM" in real-world projects (at least in my current project). Flexibility of serialization/deserialization through annotations substantially simplifies development.

Maybe I will come up with some simple solution, after more digging into current code of the library (I understand that code is generated by python scripts :) ) and send you a pull request.

n3integration:
+1

FWIW - I agree that Jackson may be a preferable choice since it is already bundled with a lot of frameworks, as @dbolotin pointed out. For an example, the Play framework's Json serialization [implementation](https://github.com/playframework/playframework/blob/master/framework/src/play-json/src/main/java/play/libs/Json.java) is a fairly straight foreword implementation and handles most use cases.

I think that RethinkDB is a great product and improved serialization support would really help foster further adoption from the Java community.

adriantodt:
The Java driver now lives in https://github.com/rethinkdb/rethinkdb-java. Jackson support is implemented in the main branch. This issue can be closed.

PR #5665, opened 2016-04-14
Support almost all types of data conversion from db to java, include nested List/Array

QianJin2013:
This is almost same as #5484, just resolved conflict.

Hope this easy to review and merge.

### DIFF
--- a/drivers/java/build.gradle
+++ b/drivers/java/build.gradle
@@ -51,6 +51,7 @@ signing {
 dependencies {
     testCompile 'junit:junit:4.12'
     testCompile 'net.jodah:concurrentunit:0.4.2'
+    testCompile 'com.fasterxml.jackson.core:jackson-databind:2.6.5'
     testRuntime 'ch.qos.logback:logback-classic:1.1.3'
     compile 'org.slf4j:slf4j-api:1.7.12'
     compile 'com.googlecode.json-simple:json-simple:1.1.1'
--- a/drivers/java/src/main/java/com/rethinkdb/annotations/IgnoreNullFields.java
+++ b/drivers/java/src/main/java/com/rethinkdb/annotations/IgnoreNullFields.java
@@ -1,0 +1,15 @@
+package com.rethinkdb.annotations;
+
+import java.lang.annotation.*;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+/**
+ * When saving this data to rethink db, skip null properties.
+ * To recursively skip null properties of children properties,
+ * specify this annotation to the children property's classes.
+ */
+public @interface IgnoreNullFields {
+}
--- a/drivers/java/src/main/java/com/rethinkdb/ast/Util.java
+++ b/drivers/java/src/main/java/com/rethinkdb/ast/Util.java
@@ -1,5 +1,6 @@
 package com.rethinkdb.ast;
 
+import com.rethinkdb.annotations.IgnoreNullFields;
 import com.rethinkdb.gen.ast.*;
 import com.rethinkdb.gen.exc.ReqlDriverCompileError;
 import com.rethinkdb.gen.exc.ReqlDriverError;
@@ -85,9 +86,8 @@ public class Util {
         final DateTimeFormatter fmt = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSX");
 
         if (val instanceof LocalDateTime) {
-            ZoneId zid = ZoneId.systemDefault();
-            DateTimeFormatter fmt2 = fmt.withZone(zid);
-            return Iso8601.fromString(((LocalDateTime) val).format(fmt2));
+            ZonedDateTime zdt = ((LocalDateTime)val).atZone(ZoneId.systemDefault());
+            return Iso8601.fromString(zdt.format(fmt));
         }
         if (val instanceof ZonedDateTime) {
             return Iso8601.fromString(((ZonedDateTime) val).format(fmt));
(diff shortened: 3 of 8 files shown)


QianJin2013:
sorry i forgot to change commit author name, it is too short ("q") now.

danielmewes:
Thanks for updating the branch @QianJin2013 .

rgv151:
when will this PR get merged?

deontologician:
Hey @rgv151, I've been busy with the Horizon release recently, but now that things have slowed down I can check this out. Thanks for the poke :)

mcjp78:
Irs been 5 months so bumpy

mcjp78:
@danielmewes

marshall007:
LGTM

@QianJin2013 would you mind rebasing this on top of `next` again so we can cleanup the conflicts? We're going to try to get this merged as soon as we can.

bchavez:
JM2C, I would probably look into going in a different direction by changing the default Java serializer to Jackson/Gson going forward. See: https://github.com/rethinkdb/rethinkdb/issues/5128

It appears there is a lot of special case type handling in this PR. I think, all this special-case code could be better handled by a more convenient JSON serializer for Java. Having a look at [`net/Util.java`](https://github.com/sjitech/rethinkdb/blob/ad253bc42a9799e4f7f821e36902d04ac2015684/drivers/java/src/main/java/com/rethinkdb/net/Util.java) there's a lot of code just for trying to select the right constructor for a POJO and plucking off special casts to/from Java native types.

IMHO, the problem, as I see it, with handling explicit special-case type conversion inside the driver like this is that it leads to a very rigid design. Someone will always have some reason to change the way the driver is handling a specific type, for whatever reason, and then we've gifted ourselves a maintenance headache trying to handle everyone's special case.

I think, the better approach toward maintainability is to push all this special-case type conversion outside the purvue of the driver. But how does one achieve such greatness? Well... if Jackson/Gson work anything like `Newtonsoft` in C# land there are serialization/deserialization extension APIs that can be implemented to output low-level wire JSON. A proper JSON implementation using *something* like [Gson's `TypeAdapter`s](http://static.javadoc.io/com.google.code.gson/gson/2.7/com/google/gson/TypeAdapter.html) could enable native Java type to `$reql_type$` pseudo type conversions without having to go crazy with `if-else` checks everywhere inside the driver. And, developer issues that require special case-conversions, for whatever reason, now become issues of customizing the Jackson/Gson serializer and not issues at odds with the **RethinkDB Java** driver itself.

At least, this was the approach I took with the `bchavez/C#` driver and it's been working out pretty well. Relatively low bug count and pretty much breeze to maintain (*alright, probably jinxed it now*). Type conversions, enums and all that get handled by `Newtonsoft`; so storing enums as `string` or `int` representations wit annotation/attributes, it don't matter, err'body doing their own thing with da JSON. Whatever people want.

:fallen_leaf: :leaves: [**Distance - Falling (ft Alys Be)**](https://www.youtube.com/watch?v=WIfwIv4wNCE)

deontologician:
:+1: for GSON/ Jackson. What initially seemed like a small addition of a
few built in coercions morphed into a maintenance black hole. For obvious
reasons that's a bad idea, so it's probably worth trying to go all the way
with a proper serializer

mcjp78:
Would that just mean replacing all serialization with Jackson or GSON? From what I can see there is a lot besides that which would need to be changed.

QianJin2013:
@marshall007 OK, i am starting rebase code now. 

Be aware that the way i solved the conversion problem was an ad-hoc & just-work way, not perfect. 
Some rare case logic are low efficiency(such as  many type .toString() then get parsed) and
the handling of generic list was a little ugly.

deontologician:
@mcjp78 what else do you think would need to be changed?

mcjp78:
@deontologician by that I meant code that needs to be changed to support Jackson.

mcjp78:
I believe the net/util and ast/util for a proper serializer (may be wrong)

QianJin2013:
hi, everyone. I'v done rebase. Now no conflicts.

mcjp78:
While this currently may be a a maintenance issue, it will work in the interim before the driver is set up to work with Jackson. I vote for a merge, it is git so it can be undone.

PeterKnego:
Please take a look at rethindb-java-orm: https://github.com/DevScore/rethinkdb-java-orm

It's a full-featured lightweight ORM built on top of Java driver.

mcjp78:
@PeterKnego make connection the args so cert files are supported. Also, what is that for.

PeterKnego:
It's a Java ORM (object-relational-mapping) built on top of RethinkDB Java driver. Unfortunately there are no docs, but it's been used in production. See the tests to see what it can do.

It's much more full-featured then built-in properties-to-object mapper.

mcjp78:
@PeterKnego i don't fully understand the use of this... Do you have any sample apps or something?

PeterKnego:
@mcjp78Please check out  the [unit tests](https://github.com/DevScore/rethinkdb-java-orm/tree/master/src/test/java/com/rethinkdb/orm). Here are some [example annotated entity classes](https://github.com/DevScore/rethinkdb-java-orm/tree/master/src/test/java/com/rethinkdb/orm/entities).

mcjp78:
Take a look at #6157 . It converts rethinkdb to using Jackson and therefore supports enums/most datatypes :)

marshall007:
@mcjp78 nice work on #6157! As @deontologician and @bchavez mentioned this is definitely the better long term solution. I'm glad it was relatively straight forward to implement.

mcjp78:
ty

mcjp78:
@QianJin2013 Ty tho for bringing this to light

QianJin2013:
I am regret to here that RethinkDb is shutting down.
So many great works.

OK, it's been a long time since i made this just-work patch, i'v already forgot it.
In fact, i do not like such a bunch of code to just a type conversion.

@mcjp78 I am glad to here you contributed a more simple way to do the conversion.

AtnNn:
Closing in favor of #6157. Thanks for your contributions.

Issue #1320, opened 2013-08-19
ReQL proposal: Passing ReQL api functions as well as lambdas


wmrowan:
Currently we officially support and recommend the following syntax:

``` python
r.expr([1,2]).reduce(lambda x,y: x + y)
```

It just so happens though that our current implementation (at least in the Python driver) is agnostic to the type and origin of the passed function. The following also works (though currently blocked by an unrelated bug in next):

``` python
r.expr([1,2]).reduce(r.add)
```

This proposal is to standardize this form, make sure it is available in all drivers, and document it as a real feature.

thelinuxlich:
@srh I think #1312 already covers this

PR #5673, opened 2016-04-16
[WIP] passing top-level ReQL functions as arguments

marshall007:
This PR lays the groundwork necessary for allowing top-level ReQL terms to be passed as arguments to other terms (see #1320). Thus far I've only updated and tested it on `r.reduce`, but the following now work in the JS driver:

``` js
r.reduce([1,1], r.add) // -> 2
r.expr([1,1]).reduce(r.add) // -> 2
```


marshall007:
Note that in order to implement this, I had to modify the signature of `r.expr`:

```js
r.expr(val, nestingDepth) // before
r.expr(val, { nestingDepth, arity }) // after
```

I felt like this was better than adding another positional argument (ad infinitum), but it would be a breaking change. We could make it backwards-compatible by assuming a numeric argument refers to `nestingDepth`.

`r.expr` will now assume `arity` when wrapping functions where the argument length is unknown. We should maybe consider having it force the arity even for functions of known argument lengths, but I couldn't imagine a use-case for that.

danielmewes:
Thanks @marshall007 .

It looks really cool, but I think we should have an API discussion about this.
We might also want to support this in other drivers.

Let me ping a few people for their opinions: @mlucy @VeXocide @deontologician @coffeemug .

coffeemug:
I'd really like to see @mlucy's opinion on this.

mlucy:
I think it's obviously good to support queries of the form `table.reduce(r.add)`.  It looks like this patch is special-casing things just in the JS driver, though, which comes with a few problems:

* We'd have to re-implement this logic in every driver.
* We have to update this logic every time we add a new term that can take functions.
* The driver has to know a lot more about the semantics of various queries -- that `reduce` takes a two-argument function, for example.  This makes the driver less dumb, and requires more coordination when we change the semantics of things on the server.

---

I think maybe what we should do instead is make the server responsible for this.

We can add a new optarg `variadic` to the `FUNC` term to support variadic functions (with the semantics that if it's set to `true`, the last argument of the function is bound to an array of the remaining arguments rather than to a single argument).  Then the drivers can just have a generic rule that they rewrite `r.add` to `[FUNC, [[gensym], [ADD, [[ARGS, [gensym]]]]], {variadic: true}]` (and the same for any other ReQL command they encounter that isn't called on anything).

marshall007:
@mlucy very much in favor of your proposal. As I was working on this in the JS driver, it certainly occurred to me that doing this on the server would make a lot of sense.

As an aside, I think we should consider generating more driver code from JSON ASTs the way @deontologician recently did for the Java driver. That way tackling and maintaining cross-cutting concerns (that can't or shouldn't be handled on the server) can still be done at a higher level. This might come into play again for something like #1190.

deontologician:
The ruby driver already does something similar to the auto-code generation, only it dynamically generates classes and methods in memory instead of dumping them to a file and compiling them

srh:
Pointing this at `v2.4.x` before we make that the main branch of the repo.

But actually, I'm going to close this and make an issue in the https://github.com/rethinkdb/rethinkdb-javascript repo.

PR #5690, opened 2016-04-19
Java Driver PR: Fixes #5439 and Fixes #5660.

### LINKED ISSUE #5439: No error on changefeed after server restart in Java driver
I set up a changefeed for a table and restarted RethinkDB. After restarting the database the application did not receive any notifications about the changes, and no exceptions were thrown.

How to reproduce the issue:

- Start the latest official RethinkDB docker image (2.2.4, image ID `eeff71def8c6`)
- Create the table `test_table` in database `test`
- Start the code below
- Make any changes in the `test_table` table => the code logs the changes
- Run `docker stop rethinkdb; sleep 1; docker start rethinkdb`
- Make any changes in the table => the code does not log anything, but the app jumps to 100% CPU usage. (I waited a minute to make sure that if the client wants to reconnect, it has enough time for that).

Test code:

```java
package hu.gyim.rethinkdb_test;

import com.rethinkdb.RethinkDB;
import com.rethinkdb.net.Connection;
import com.rethinkdb.net.Cursor;

public class Main {

    public static void main(String[] args) {
        try {
            Connection conn = RethinkDB.r.connection().hostname("192.168.99.100").port(28015).connect();
            conn.use("test");

            Cursor cur = RethinkDB.r.table("test_table").changes().run(conn);
            for (Object change : cur) {
                System.out.println(change);
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

Stack trace after the restart:

```
"main@1" prio=5 tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
      at sun.nio.ch.FileDispatcherImpl.read0(FileDispatcherImpl.java:-1)
      at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:39)
      at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
      at sun.nio.ch.IOUtil.read(IOUtil.java:197)
      at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
      - locked <0x416> (a java.lang.Object)
      at com.rethinkdb.net.SocketWrapper.recvall(SocketWrapper.java:107)
      at com.rethinkdb.net.ConnectionInstance.readResponse(ConnectionInstance.java:70)
      at com.rethinkdb.net.Connection.readResponse(Connection.java:130)
      at com.rethinkdb.net.Cursor$DefaultCursor.getNext(Cursor.java:172)
      at com.rethinkdb.net.Cursor.next(Cursor.java:110)
      at hu.gyim.rethinkdb_test.Main.main(Main.java:15)
```

### LINKED ISSUE #5660: Java Driver: Connection is lost during changes(), never times out.
My code is essentially:

```java
try (Connection conn = dbp.getConnection()) {
  Cursor<HashMap<String, ?>> cursor = r.table("messages").changes().run(conn);
  for (HashMap<String, ?> item : cursor) {
    .. do stuff ...
  }
}
```

I connected to the server and was successfully listening to changes. Then, I killed the database server (only a single instance was running). As far as I can tell, the Java driver does not break the for loop when the connection is lost it just keeps sitting there expecting more data forever. There doesn't appear to be any timeout period. And, obviously, if I restart the database server then changes don't start coming in again on that old cursor.

So my question is: how are we supposed to detect connection losses and recover from them if the driver just sits there in the loop?

bchavez:
Hola,

I think this PR will fix #5439 and fix #5660 for the Java driver.

The core of the bug is described here: https://github.com/rethinkdb/rethinkdb/issues/5660#issuecomment-210734741

The crux of the issue is that only `IOException` is being handled. If another exception type was thrown, the `ResponsePump` thread exceptionally exit without clean up and without notifying cursor changefeeds causing what looks like a stalled changefeed upon a server restart.

I'm not sure how I could write a unit test for this particular issue since it would require restarting the RethinkDB server :electric_plug: during a unit test.

All unit tests pass (except a few string tests). Probably because some M$ Windowz string quirks?

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -118,7 +118,7 @@ public class Connection implements Closeable {
                     if (awaiter != null) {
                         awaiter.complete(response);
                     }
-                } catch (IOException e) {
+                } catch (Exception e) {
                     awaiterException = e;
                     this.close();
                     break;


danielmewes:
Cool, thanks for another PR @bchavez .

@deontologician

pires:
While this works, I'm not sure catching all exceptions is clear about its intents. I mean, I would fix it by throwing an `IOException` [here](https://github.com/bchavez/rethinkdb/blob/connection_fix/drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java#L148). Or at least, I wouldn't catch `Exception`.

Unfortunately, my time to look at this is nil atm so 👍

bchavez:
I don't know what the best practice is in Java. It could certainly be bad practice.

However, I mostly see this from a conceptual point of view though. At least, my thinking is when the driver created a dedicated thread for processing responses, the driver should maintain control over the thread and not let it get out of bounds exceptionally and silently fail/stop.

Also, handling `Exception e` isolates the response thread from the underlying implementation details in `SocketWrapper.java` whereas now the response thread is dependent on the underlying implementation (to catch and only catch `IOException`). Swapping out the underlying `SocketWrapper` with a different implementation (or making a change) that could potentially throw a new kind of exception (maybe an `IndexOutOfBoundsException`?) would alter the correctness of the response loop when the thread is catching and only catching `IOException`. And then we would run into the same issue again with the thread silently failing/stopping without clean up or fanfare on `IndexOutOfBoundsException`.

For what it's worth, the .NET driver also catches `Exception e` in it's response loop.

deontologician:
Let's :shipit: 

Thanks again @bchavez for checking this out and coming up with the solution. I'll backport this to the 2.3.x branch so it can go out in the next driver point release

deontologician:
v2.3.x in 3387ce1

Issue #5230, opened 2015-12-21
Cannot install RethinkDB from latest sources on OS X Yosemite


sreekaranugu:
I'm keep getting this error while trying to install on OS X Yosemite. I can't get rid of this, as there isn't much information related to it.

```
git clone URL
./configure --allow-fetch
make
```

The error is:

```
npm ERR! argv "/usr/local/Cellar/node/5.3.0/bin/node" "/usr/local/bin/npm" "--userconfig" "/Users/sreekar/github/rethinkdb/rethinkdb/mk/support/config/npm.conf" "--cache" "/Users/sreekar/github/rethinkdb/rethinkdb/external/npm_cache-SeFruCOS" "--no-registry" "install" "/Users/sreekar/github/rethinkdb/rethinkdb/external/admin-deps_2.0.3"
npm ERR! node v5.3.0
npm ERR! npm v3.3.12

npm ERR! registry URL is required
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR! https://github.com/npm/npm/issues
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/Cellar/node/5.3.0/bin/node" "/usr/local/bin/npm" "--userconfig" "/Users/sreekar/github/rethinkdb/rethinkdb/mk/support/config/npm.conf" "--cache" "/Users/sreekar/github/rethinkdb/rethinkdb/external/npm_cache-SeFruCOS" "--no-registry" "install" "/Users/sreekar/github/rethinkdb/rethinkdb/external/admin-deps_2.0.3"
npm ERR! node v5.3.0
npm ERR! npm v3.3.12

npm ERR! registry URL is required
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR! https://github.com/npm/npm/issues

npm ERR! Please include the following file with any support request:
npm ERR! /Users/sreekar/github/rethinkdb/rethinkdb/build/external/admin-deps_2.0.3/npm-debug.log

Full error log: ./build/external/admin-deps_2.0.3_install.log
make[1]: *** [build/external/admin-deps_2.0.3/bin/gulp] Error 1
make: *** [make] Error 2
```

Thanks

danielmewes:
@larkost Do you have any ideas about this one?

danielmewes:
@sreekaranugu Could you try `./configure --allow-fetch --fetch npm --fetch node`?

AtnNn:
[...]

PR #5698, opened 2016-04-21
fix(npm-pkg): support node 5 for build process

mbroadst:
The npm shipped with node 5 (at least this is all I've checked) no
longer supports the `--no-registry` flag. Removing this flag keeps
support for older versions, while also supporting a local node 5
installation

/cc @larkost

### DIFF
--- a/mk/support/pkg/npm-pkg.inc
+++ b/mk/support/pkg/npm-pkg.inc
@@ -49,7 +49,7 @@ pkg_install () {
     mkdir -p "$install_dir/node_modules"
     local pkg
     pkg=$(niceabspath "$src_dir")
-    in_dir "$install_dir" npm --no-registry install "$pkg"
+    in_dir "$install_dir" npm install "$pkg"
 
     local bin_dir="$install_dir/node_modules/$full_npm_package/node_modules/.bin/"
     mkdir -p "$install_dir/bin"


danielmewes:
Something else we discussed in https://github.com/rethinkdb/rethinkdb/issues/5230 was supplying a fake registry (such as `--registry http://no-npm-registry.rethinkdb.com`).

I'm not completely sure what the original intention between the `--no-registry` was, but I imagine there was some reason for it to be there.

@AtnNn or @larkost do you have any insights into this?

mbroadst:
ping

danielmewes:
Sorry for the delay. We'll try to look into this end of the week.

larkost:
I looked into this, and thus far I think that `--registry http://no-npm-registry.rethinkdb.com` might be a good option. I think that this is all an attempt to make the non-download portion of our build process work when there is no internet connection. The more I read about this, the more it looks like the `npm` team has flubbed this particular use case, repeatedly in different ways.

I did some preliminary testing on one machine, but need to do more testing with various versions of `node`/`npm`. As a side note, I tried bumping the version of `node` that we use internally to `v4.4.5` (the current LTS version), and everything works, at least on the surface.

mbroadst:
@larkost sounds good. node 6 is their LTS, so I imagine you're going to run into these types of issues. Do you guys have a preference in terms of solution?

larkost:
I have some testing in already for using `--registry http://no-npm-registry.rethinkdb.com`, but just have had other things bump in front of the testing I need to do to close this one off. The biggest annoyance is that I need to do offline testing for a bunch of versions of `npm`, so it is a bit of work.

AtnNn:
The `--no-registry` option has already been removed in `next` (accidentally). It was meant as a safeguard to ensure `npm` installed only packages from the local bundle.

As far as I know, that safeguard was never triggered. I'm happy to leave it out.

PR #5701, opened 2016-04-21
feat(auto_reconnect): allow runtime configuration of give_up_ms (v2)

mbroadst:
Currently the auto_reconnect instance used to reconnect lost nodes
is hardcoded to give up reconnecting after 24 hours. This is not
ideal in some scenarios where a user may want to remove a node from
a cluster, without having to reset all other participating nodes in
the cluster. This patch allows the user to change that value.

### DIFF
--- a/src/clustering/administration/main/command_line.cc
+++ b/src/clustering/administration/main/command_line.cc
@@ -60,6 +60,10 @@
 #define RETHINKDB_RESTORE_SCRIPT "rethinkdb-restore"
 #define RETHINKDB_INDEX_REBUILD_SCRIPT "rethinkdb-index-rebuild"
 
+namespace cluster_defaults {
+    const int reconnect_timeout = (24 * 60 * 60);    // 24 hours (in secs)
+}  // namespace cluster_defaults
+
 MUST_USE bool numwrite(const char *path, int number) {
     // Try to figure out what this function does.
     FILE *fp1 = fopen(path, "w");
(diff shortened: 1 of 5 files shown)


mbroadst:
sorry had to open a new PR because I seem to have lost the branch the original PR was opened on (most likely due to trying to solve issues with gitattributes! 😄)

danielmewes:
I'll take another look today.

src/clustering/administration/main/command_line.cc
```diff
@@ -443,6 +447,28 @@ boost::optional<int> parse_join_delay_secs_option(
...
+        if (!strtou64_strict(timeout_opt, 10, &node_reconnect_timeout_secs)) {
+            throw std::runtime_error(strprintf(
+                    "ERROR: cluster-reconnect-timeout should be a number, got '%s'",
+                    timeout_opt.c_str()));
+        }
+        if (node_reconnect_timeout_secs > std::numeric_limits<int>::max() ||
+            node_reconnect_timeout_secs * 1000 > std::numeric_limits<int>::max()) {
+            throw std::runtime_error(strprintf(
+                    "ERROR: cluster-reconnect-timeout is too large. Must be at most %d",
+                    std::numeric_limits<int>::max()));
```
danielmewes: The maximum in the error message needs to take the * 1000 into account.

src/clustering/administration/main/command_line.cc
```diff
@@ -1519,6 +1545,11 @@ options::help_section_t get_network_options(const bool join_required, std::vecto
                                              strprintf("%d", port_defaults::peer_port)));
     help.add("--cluster-port port", "port for receiving connections from other nodes");
 
+    options_out->push_back(options::option_t(options::names_t("--cluster-reconnect-timeout"),
+                                             options::OPTIONAL,
+                                             strprintf("%d", cluster_defaults::reconnect_timeout)));
+    help.add("--cluster-reconnect-timeout ms", "maximum number of secconds to attempt reconnecting to a node before giving up");
```
danielmewes: This is still saying "ms" in `"--cluster-reconnect-timeout ms"`, should be "secs".
There also is a typo in the help message: `secconds` instead of `seconds`.

danielmewes:
Left just two comments. The rest of it looks great!

src/clustering/administration/main/command_line.cc
mbroadst: I think the maximum in the error message is still consistently the actual maximum (numeric_limits<int>::max === biggest supported integer right?). So the error message is correct, it was the checks that needed to ensure that:

  a) the input number is indeed within that threshold
  b) secondarily, since we intend to use it as ms the number multiplied by 1000 should still be within that range

Perhaps you think the error message should be more descriptive in the event that someone enters a number that is too big only once we are testing for conversion to ms?

lbguilherme: It should be:

```c++
throw std::runtime_error(strprintf(
    "ERROR: cluster-reconnect-timeout is too large. Must be at most %d",
    std::numeric_limits<int>::max() / 1000));
```

mbroadst: oh right right

danielmewes:
Looks good, thanks @mbroadst .

danielmewes:
I'm waiting for some small style improvements to go through review, and will cherry-pick this into `v2.3.x` then. It will ship with RethinkDB 2.3.2.

mbroadst:
@danielmewes fantastic, thanks!

danielmewes:
Cherry-picked into v2.3.x via c64cbe4, with small style improvements for the help output (for this and also other commands) in 057518d.

Issue #5614, opened 2016-04-05
Make the Python and JS drivers cache the results of PBKDF2 to reduce latency.


mlucy:
We aren't including the server nonce in the PBKDF2 calculation, so we can cache the result of it to drastically speed up connection times when re-authenticating as the same user.  This is probably a good idea since some users may be on the old "one connection per query" model.  (This is already in review for Ruby.)

danielmewes:
... and Java driver!

This is not critical for 2.3 (but a nice plus). We can update the drivers later.

(@nighelles, @deontologician, @VeXocide )

nighelles:
This should be fixed for the javascript driver.

danielmewes:
@deontologician implemented this for the Java driver in ad1d38fd16ebcd30509f4882f0c6114f1aaf906. I also cherry-picked the change into v2.3.x via ea074d3.

nighelles:
This should be fixed for the python driver as well.

nighelles:
That should be all the drivers.

PR #5704, opened 2016-04-22
2.3.1 release notes

danielmewes:


danielmewes:
@dalanmiller reviewed this.

I'm keeping this open until tomorrow in case we need to add more items to the list.

danielmewes:
Added one more item describing the resolution of #5614 for the Java driver.

PR #5720, opened 2016-04-26
intra-cluster handshake improvements

mbroadst:
This is a series of patches that aim to improve intra-cluster connections. This is primarily achieved by the following:

- Introduce a `join_result_t` enum in order to return meaningful results from `connectivity_cluster_t::run_t::join_blocking`. This allows for enabling certain behaviors based on that reuslt.

- One of those aforementioned behaviors is to stop retrying intra-cluster reconnects in the `auto_reconector_t` if a join has failed irrecoverably (a `join_result_t::PERMANENT_ERROR`) 

- Finally, a new `PERMANENT_ERROR` was introduced in the `run_t::handle` method: if an expected server id is provided, and the remote server's id does not equal that value upon initial handshake, then that will prevent further reconnects.

/cc @danielmewes

### DIFF
--- a/src/clustering/administration/servers/auto_reconnect.cc
+++ b/src/clustering/administration/servers/auto_reconnect.cc
@@ -97,7 +97,7 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
 
                 if (result == join_result_t::PERMANENT_ERROR &&
                     addresses.find(server) != addresses.end()) {
-                    logNTC("Unrecoverable connection error to remote peer: %s\n", server.print().c_str());
+                    logNTC("Unrecoverable connection error to remote server: %s", server.print().c_str());
                     join_failed.pulse_if_not_already_pulsed();
                     addresses.erase(it);
                 }
--- a/src/clustering/administration/servers/auto_reconnect.cc
+++ b/src/clustering/administration/servers/auto_reconnect.cc
@@ -84,20 +84,24 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
 
     wait_any_t interruptor(&reconnected, &give_up_timer, keepalive.get_drain_signal());
 
+    cond_t join_failed;
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
-        while (!interruptor.is_pulsed()) {
-            join_result_t result =
-                connectivity_cluster_run->join_blocking(
-                    last_known_address, boost::none,
-                    join_delay_secs,
-                    auto_drainer_t::lock_t(&drainer));
+        while (!interruptor.is_pulsed() && !join_failed.is_pulsed()) {
+            coro_t::spawn_now_dangerously([&]() {
+                join_result_t result =
+                    connectivity_cluster_run->join_blocking(
+                        last_known_address, boost::none, server,
+                        join_delay_secs,
+                        auto_drainer_t::lock_t(&connectivity_cluster_run->drainer));
 
-            if (result == join_result_t::PERMANENT_ERROR) {
-                logNTC("Unrecoverable connection error to remote peer: %s\n", server.print().c_str());
-                addresses.erase(it);
-                return;
-            }
+                if (result == join_result_t::PERMANENT_ERROR &&
+                    addresses.find(server) != addresses.end()) {
+                    logNTC("Unrecoverable connection error to remote peer: %s\n", server.print().c_str());
+                    join_failed.pulse_if_not_already_pulsed();
+                    addresses.erase(it);
+                }
+            });
 
             backoff.failure(&interruptor);
         }
--- a/src/clustering/administration/servers/auto_reconnect.cc
+++ b/src/clustering/administration/servers/auto_reconnect.cc
@@ -82,24 +82,48 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
         },
         initial_call_t::YES);
 
-    wait_any_t interruptor(&reconnected, &give_up_timer, keepalive.get_drain_signal());
-
     cond_t join_failed;
+    wait_any_t interruptor(
+        &reconnected, &give_up_timer, keepalive.get_drain_signal(), &join_failed);
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
-        while (!interruptor.is_pulsed() && !join_failed.is_pulsed()) {
-            coro_t::spawn_now_dangerously([&]() {
-                join_result_t result =
+        // These can be safely passed into the coroutine below.
+        // They will be reset to `nullptr` by the assignment_sentry_ts when this function
+        // call ends.
+        std::shared_ptr<cond_t *> join_failed_out(new cond_t *(nullptr));
+        std::shared_ptr<peer_address_t *> last_known_address_out(new peer_address_t *(nullptr));
+        assignment_sentry_t<cond_t *> join_failed_out_assignment(join_failed_out.get(), &join_failed);
+        assignment_sentry_t<peer_address_t *> last_known_address_out_assignment(last_known_address_out.get(), &last_known_address);
+
+        while (!interruptor.is_pulsed()) {
+            // This coroutine can keep running even after this function has returned
+            // or the auto_reconnector_t was destructed. So we must pass everything
+            // we need by value.
+
+            coro_t::spawn_now_dangerously(
+                [this, last_known_address, server, join_failed_out, last_known_address_out]() {
+                auto &connectivity_cluster_run_local = connectivity_cluster_run;
+                auto &join_delay_secs_local = join_delay_secs;
+
+                join_results_t results =
                     connectivity_cluster_run->join_blocking(
                         last_known_address, boost::none, server,
-                        join_delay_secs,
-                        auto_drainer_t::lock_t(&connectivity_cluster_run->drainer));
+                        join_delay_secs_local,
+                        auto_drainer_t::lock_t(&connectivity_cluster_run_local->drainer));
 
-                if (result == join_result_t::PERMANENT_ERROR &&
-                    addresses.find(server) != addresses.end()) {
-                    logNTC("Unrecoverable connection error to remote server: %s", server.print().c_str());
-                    join_failed.pulse_if_not_already_pulsed();
-                    addresses.erase(it);
+                // Check if the `try_reconnect` function that spawned us is still alive
+                if (*join_failed_out != nullptr && *last_known_address_out != nullptr) {
+                    ASSERT_NO_CORO_WAITING;
+
+                    for (auto r = results.begin(); r != results.end(); ++r) {
+                        if (r->second != join_result_t::PERMANENT_ERROR) continue;
+                        (*last_known_address_out)->erase_ip(r->first);
+                        if (last_known_address.ips().empty()) {
+                            logNTC("Unrecoverable connection error to remote server: %s",
+                                server.print().c_str());
+                            (*join_failed_out)->pulse_if_not_already_pulsed();
+                        }
+                    }
                 }
             });
 
--- a/src/clustering/administration/servers/auto_reconnect.cc
+++ b/src/clustering/administration/servers/auto_reconnect.cc
@@ -91,9 +91,12 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
         // They will be reset to `nullptr` by the assignment_sentry_ts when this function
         // call ends.
         std::shared_ptr<cond_t *> join_failed_out(new cond_t *(nullptr));
-        std::shared_ptr<peer_address_t *> last_known_address_out(new peer_address_t *(nullptr));
-        assignment_sentry_t<cond_t *> join_failed_out_assignment(join_failed_out.get(), &join_failed);
-        assignment_sentry_t<peer_address_t *> last_known_address_out_assignment(last_known_address_out.get(), &last_known_address);
+        std::shared_ptr<peer_address_t *> last_known_address_out(
+            new peer_address_t *(nullptr));
+        assignment_sentry_t<cond_t *> join_failed_out_assignment(
+            join_failed_out.get(), &join_failed);
+        assignment_sentry_t<peer_address_t *> last_known_address_out_assignment(
+            last_known_address_out.get(), &last_known_address);
 
         while (!interruptor.is_pulsed()) {
             // This coroutine can keep running even after this function has returned
@@ -102,11 +105,11 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
 
             coro_t::spawn_now_dangerously(
                 [this, last_known_address, server, join_failed_out, last_known_address_out]() {
-                auto &connectivity_cluster_run_local = connectivity_cluster_run;
-                auto &join_delay_secs_local = join_delay_secs;
+                auto connectivity_cluster_run_local = connectivity_cluster_run;
+                auto join_delay_secs_local = join_delay_secs;
 
                 join_results_t results =
-                    connectivity_cluster_run->join_blocking(
+                    connectivity_cluster_run_local->join_blocking(
                         last_known_address, boost::none, server,
                         join_delay_secs_local,
                         auto_drainer_t::lock_t(&connectivity_cluster_run_local->drainer));
--- a/src/clustering/administration/servers/auto_reconnect.cc
+++ b/src/clustering/administration/servers/auto_reconnect.cc
@@ -87,7 +87,18 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
         while (!interruptor.is_pulsed()) {
-            connectivity_cluster_run->join(last_known_address, join_delay_secs);
+            join_result_t result =
+                connectivity_cluster_run->join_blocking(
+                    last_known_address, boost::none,
+                    join_delay_secs,
+                    auto_drainer_t::lock_t(&drainer));
+
+            if (result == join_result_t::PERMANENT_ERROR) {
+                logNTC("Unrecoverable connection error to remote peer: %s\n", server.print().c_str());
+                addresses.erase(it);
+                return;
+            }
+
             backoff.failure(&interruptor);
         }
     } catch (const interrupted_exc_t &) {
--- a/src/clustering/administration/servers/auto_reconnect.hpp
+++ b/src/clustering/administration/servers/auto_reconnect.hpp
@@ -44,10 +44,6 @@ private:
     connected servers we get from the `connectivity_cluster_t`. */
     std::map<peer_id_t, server_id_t> server_ids;
 
-    /* `stop_conds` contains an entry for each running instance of `try_reconnect()`.
-    It's used to interrupt the coroutines if the server reconnects. */
-    std::multimap<server_id_t, cond_t *> stop_conds;
-
     int join_delay_secs;
     int give_up_ms;
 
--- a/src/rpc/connectivity/cluster.cc
+++ b/src/rpc/connectivity/cluster.cc
@@ -412,7 +412,7 @@ join_result_t connectivity_cluster_t::run_t::join_blocking(
 
     // Attempt to connect to all known ip addresses of the peer
     bool successful_join = false; // Variable so that handle() can check that only one connection succeeds
-    join_result_t join_result; // Used to determine the joint result for an individual connection attempt
+    join_result_t join_result; // Used to determine the the joint join result across all individual connection attempts
     static_semaphore_t rate_control(peer.ips().size()); // Mutex to control the rate that connection attempts are made
     rate_control.co_lock(peer.ips().size() - 1); // Start with only one coroutine able to run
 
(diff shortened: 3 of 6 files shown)


danielmewes:
Thanks! I'll try to take a look at this soon.

I think we're going to put this into 2.4.0 rather than a point release since it's a slightly bigger change. Is that ok with you?

mbroadst:
@danielmewes totally fine, we'll be migrating towards building our appliance with yocto relatively soon so I can apply patches as needed manually there rather than e.g. depending on proper ubuntu releases. Looking forward to you (and the team's) thoughts on these changes.

mbroadst:
@danielmewes am I expected to keep this up to date with next, or is it enough to keep this here as reference for when you guys decide to move forward with the code?

danielmewes:
@mbroadst We can update this from our side once we merge it.

mbroadst:
@danielmewes great, any idea on timeline for integration and release btw? just looking for ballpark here nothing fixed - end of summer? Q4?

danielmewes:
I think we'll need about 6-8 more weeks for 2.4. The integration will happen some time before then, hopefully in the next few weeks.

mbroadst:
👍, I appreciate the quick feedback

mbroadst:
updated to resolve the conflict (I thought there were many more for some reason)

src/rpc/connectivity/cluster.cc
```diff
@@ -380,17 +383,18 @@ void connectivity_cluster_t::run_t::connect_to_peer(
...
+        boost::optional<server_id_t> expected_server_id,
         const int join_delay_secs,
         auto_drainer_t::lock_t drainer_lock) THROWS_NOTHING {
     drainer_lock.assert_is_holding(&parent->current_run->drainer);
     parent->assert_thread();
     {
         mutex_assertion_t::acq_t acq(&attempt_table_mutex);
         if (attempt_table.find(peer) != attempt_table.end()) {
-            return;
+            return join_result_t::PERMANENT_ERROR;
```
mbroadst: @danielmewes I think if an attempt is already occurring then maybe this is a `TEMPORARY_ERROR` ?

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -87,7 +87,18 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
         while (!interruptor.is_pulsed()) {
-            connectivity_cluster_run->join(last_known_address, join_delay_secs);
+            join_result_t result =
+                connectivity_cluster_run->join_blocking(
+                    last_known_address, boost::none, server,
+                    join_delay_secs,
+                    auto_drainer_t::lock_t(&drainer));
```
mbroadst: @danielmewes this is causing a guarantee error:
```
2016-06-24T11:16:51.422715984 708.920404s error: Error in src/concurrency/auto_drainer.cc at line 74:
2016-06-24T11:16:51.422848529 708.920537s error: Guarantee failed: [p == parent]
2016-06-24T11:16:51.422891086 708.920579s error: Backtrace:
2016-06-24T11:16:51.863289243 709.360982s error: Fri Jun 24 11:16:51 2016
1 [0xa93e50]: backtrace_t::backtrace_t() at ??:?
2 [0xa941e3]: format_backtrace(bool) at ??:?
3 [0xa705bd]: report_fatal_error(char const*, int, char const*, ...) at ??:?
4 [0xa6ce45]: auto_drainer_t::lock_t::assert_is_holding(auto_drainer_t*) const at ??:?
5 [0xac8511]: connectivity_cluster_t::run_t::join_blocking(peer_address_t, boost::optional<peer_id_t>, boost::optional<server_id_t>, int, auto_drainer_t::lock_t) at ??:?
6 [0xc6a7ae]: auto_reconnector_t::try_reconnect(server_id_t const&, auto_drainer_t::lock_t) at ??:?
7 [0xc6aee7]: callable_action_instance_t<std::_Bind<std::_Mem_fn<void (auto_reconnector_t::*)(server_id_t const&, auto_drainer_t::lock_t)> (auto_reconnector_t*, server_id_t, auto_drainer_t::lock_t)> >::run_action() at ??:?
8 [0x7733f7]: coro_t::run() at ??:?
```

any idea why? not very familiar with your `auto_drainer_t`

mbroadst: oh I see, I think this is because we're reusing `join_blocking` in a different location, and so the assumption of who the lock belongs to has changed..

mbroadst: simply removing the guarantee leads to other crazier errors, so I think we have some logical errors in here as well. I'll wait to hear back from you.

danielmewes:
Thanks for updating the patch @mbroadst . I'll be out of office the coming week, but will try to take a look at this soon after.

mbroadst:
@danielmewes hey thanks for the update. Might there be another team member I could bounce ideas off of in the meantime? It was fine for this to not be merged until 2.4 when I thought it all worked, but it looks like its failing in some pretty clever ways now that I'm trying to test in a real environment and while I'm brushing back up on my rethink internals it would be fantastic to get some pointers from "those who know" 😄

danielmewes:
@VeXocide might have some ideas about what's causing the current issues.

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -84,10 +84,25 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
 
     wait_any_t interruptor(&reconnected, &give_up_timer, keepalive.get_drain_signal());
 
+    cond_t join_failed;
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
-        while (!interruptor.is_pulsed()) {
-            connectivity_cluster_run->join(last_known_address, join_delay_secs);
+        while (!interruptor.is_pulsed() && !join_failed.is_pulsed()) {
+            coro_t::spawn_now_dangerously([&]() {
```
danielmewes: This lambda function has to capture a copy of `keepalive` to make sure that the auto_reconnector_t is not destructed while the coroutine exists:
```
coro_t::spawn_sometime([keepalive, &]() {
```

I think with that keepalive copy, it can be a normal `spawn_sometime`, and doesn't need to be a `spawn_now_dangerously` anymore since the `connectivity_cluster_run` is guaranteed to remain valid for as long as the `auto_reconnector_t` instance is there.

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -84,10 +84,25 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
...
+                join_result_t result =
+                    connectivity_cluster_run->join_blocking(
+                        last_known_address, boost::none, server,
+                        join_delay_secs,
+                        auto_drainer_t::lock_t(&connectivity_cluster_run->drainer));
+
+                if (result == join_result_t::PERMANENT_ERROR &&
+                    addresses.find(server) != addresses.end()) {
```
danielmewes: What is the `addresses.find` check for?

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -84,10 +84,25 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
...
+                    logNTC("Unrecoverable connection error to remote peer: %s\n", server.print().c_str());
```
danielmewes: I suggest using `to remote server %s`.
Since we also have peer IDs, which are different from the server IDs, using the word `peer` here might be confusing.
(also log messages shouldn't end in `\n`, though it is stripped off automatically)

src/clustering/administration/servers/auto_reconnect.hpp
```diff
@@ -44,10 +44,6 @@ class auto_reconnector_t {
     connected servers we get from the `connectivity_cluster_t`. */
     std::map<peer_id_t, server_id_t> server_ids;
 
-    /* `stop_conds` contains an entry for each running instance of `try_reconnect()`.
```
danielmewes: So this was just unused eh? Good catch.

src/clustering/administration/servers/auto_reconnect.cc
danielmewes: Oh I see, that's for the case where a previous join attempt already failed and removed the address?

src/clustering/administration/servers/auto_reconnect.cc
danielmewes: Actually there's another problem that the call to `try_reconnect` can finish while this coroutine is still running. Because `join_failed` is a local variable, and the coroutine accesses it, that would be illegal.

Instead of taking a copy of `keepalive`, I suggest instantiating a new `auto_drainer_t` locally in the `try_reconnect` method. Make sure you declare that drainer after `join_failed`, so that it is drained before `join_failed` gets destructed when `try_reconnect` returns. To make things more explicit, I suggest specifying all variable bindings explicitly in the lambda, rather than using the catch-all `&`:
```
auto_drainer_t::lock_t local_keepalive = local_drainer.lock();
coro_t::spawn_sometime(
    [local_keepalive /* must capture by value */, join_failed &, last_known_address, server, this]() {
```

danielmewes: Note that fixing this might potentially cause dead-locks during shutdown, since `join_blocking` I believe keeps running if the join is successful until the connection is terminated. Depending on how things get shut down, the connections might only get terminated when the `connectivity_cluster_run` is destructed. However `connectivity_cluster_run` can't get destructed as long as the `auto_reconnector_t` is around, and the `auto_reconnector_t` (after the suggested change) cannot be destructed while the coroutine with the `join_blocking` is still running.

I'm not sure what the best solution for this is. If it turns out to be a problem, we can discuss some ideas on Slack maybe.

src/rpc/connectivity/cluster.cc
```diff
@@ -363,9 +366,14 @@ void connectivity_cluster_t::run_t::connect_to_peer(
                 tls_ctx, selected_addr->ip(), selected_addr->port().value(),
                 drainer_lock.get_drain_signal(), cluster_client_port);
             if (!*successful_join_inout) {
-                handle(
+                join_result_t join_result = handle(
                     &conn, expected_id, boost::optional<peer_address_t>(*address),
-                    drainer_lock, successful_join_inout, join_delay_secs);
+                    expected_server_id, drainer_lock, successful_join_inout, join_delay_secs);
+
+                // NOTE: not an ideal way to handle this
```
danielmewes: I think you could return this value from `connect_to_peer`, and then do the "merging" of the different results in a lambda function that you pass to the `pmap` in `join_blocking`.

src/rpc/connectivity/cluster.cc
```diff
@@ -380,17 +388,21 @@ void connectivity_cluster_t::run_t::connect_to_peer(
...
+        boost::optional<server_id_t> expected_server_id,
         const int join_delay_secs,
         auto_drainer_t::lock_t drainer_lock) THROWS_NOTHING {
     drainer_lock.assert_is_holding(&parent->current_run->drainer);
     parent->assert_thread();
     {
         mutex_assertion_t::acq_t acq(&attempt_table_mutex);
         if (attempt_table.find(peer) != attempt_table.end()) {
-            return;
+            // NOTE: this is triggered all the time in the old code and ignored as what
```
danielmewes: Doesn't this just mean that we have an active connection attempt to this peer already?
I think returning a TEMPORARY_ERROR is correct in that case. We don't know yet if the other join attempt is going to actually work out (plus even if it does, it might get interrupted later and we still need to try reconnecting using this address).

src/rpc/connectivity/cluster.cc
```diff
@@ -400,6 +412,7 @@ void connectivity_cluster_t::run_t::join_blocking(
 
     // Attempt to connect to all known ip addresses of the peer
     bool successful_join = false; // Variable so that handle() can check that only one connection succeeds
+    join_result_t join_result; // Used to determine the joint result for an individual connection attempt
```
danielmewes: Is the `joint` a typo or should the comment say "the joint join result across all individual connection attempts"?

src/rpc/connectivity/cluster.cc
```diff
@@ -953,7 +971,7 @@ void connectivity_cluster_t::run_t::handle(
                 }
             }
             fail_handshake(conn, peername, reason, handshake_error_supported);
-            return;
+            return join_result_t::TEMPORARY_ERROR;
```
danielmewes: I think an incompatible protocol version should be a PERMANENT_ERROR. If the administrator restarts the other server later with a different version, that server is going to connect to us. So there is no need to keep trying to connect to it I think.

src/rpc/connectivity/cluster.cc
```diff
@@ -963,16 +981,28 @@ void connectivity_cluster_t::run_t::handle(
...
+        }
+
+        if (expected_server_id && expected_server_id != remote_server_id) {
+            if (peer_addr.ip().is_loopback()) {
+                return join_result_t::TEMPORARY_ERROR;
+            }
+
+            logINF("Rejected a connection from server %s since it does not have expected id %s.",
```
danielmewes: `does not have *the* expected ID` ?

danielmewes: I think this should also be a `logNTC`

src/rpc/connectivity/cluster.cc
```diff
@@ -1028,7 +1058,7 @@ void connectivity_cluster_t::run_t::handle(
                 "server with the `--initial-password auto` option to allow joining "
                 "the password-protected cluster.");
             fail_handshake(conn, peername, reason);
-            return;
+            return join_result_t::PERMANENT_ERROR;
```
danielmewes: To be conservative, we might want to keep this as a TEMPORARY_ERROR.
A user could connect to either this or the other server, and reset or set the admin password respectively, without restarting it.

src/rpc/connectivity/cluster.cc
```diff
@@ -1037,27 +1067,27 @@ void connectivity_cluster_t::run_t::handle(
...
         if (deserialize_universal_and_check(conn, &handshake_result, peername)) {
-            return;
+            return join_result_t::TEMPORARY_ERROR;
         }
         if (handshake_result.get_code() != handshake_result_code_t::SUCCESS) {
             logWRN("Remote node refused to connect with us, peer: %s, reason: \"%s\"",
                    peername,
                    sanitize_for_logger(handshake_result.get_error_reason()).c_str());
-            return;
+            return join_result_t::PERMANENT_ERROR;
```
danielmewes: Assuming there's at least one failed handshake that's a TEMPORARY_ERROR on the sender's side (see above for the admin password case for example), this must also be considered a TEMPORARY_ERROR.

You could distinguish based on the error code in the handshake if you think that it's important to return a PERMANENT_ERROR here for some cases (is it?).

src/rpc/connectivity/cluster.cc
danielmewes: I feel like we should still send a `fail_handshake` with an appropriate `handshake_result_t::error` here.

src/rpc/connectivity/cluster.cc
```diff
@@ -1075,26 +1105,29 @@ void connectivity_cluster_t::run_t::handle(
...
     if (other_id == parent->me) {
         // TODO: report this on command-line in some cases. see issue 546 on github.
-        return;
+        return join_result_t::PERMANENT_ERROR;
     }
+
     if (other_id.is_nil()) {
         logERR("Received nil peer id from %s, closing connection.", peername);
-        return;
+        return join_result_t::PERMANENT_ERROR;
```
danielmewes: I'm not sure in what cases this can happen to be honest (I've never seen it happen either).
However if in doubt, we should return a TEMPORARY_ERROR.

src/rpc/connectivity/cluster.cc
```diff
@@ -409,8 +422,10 @@ void connectivity_cluster_t::run_t::join_blocking(
                    &peer,
```
danielmewes: Actually I think we need to find a way to report permanent failures on a per-IP basis. Otherwise a single wrong IP for a server (e.g. if `127.0.0.1` ends up in the IP address list) can lead to us stopping to reconnect to the server altogether.

In the `auto_reconnector_t`, we must then only remove the individual IPs that failed from the addresses set in the `peer_address_t` structure. Only when that set becomes empty should we stop reconnecting to the server altogether I think.

danielmewes:
@mbroadst I did a pass through your patch. I left a few comments. It generally looks good.

mbroadst:
Great, thanks for the review I'll go through the comments tomorrow morning.

src/clustering/administration/servers/auto_reconnect.cc
mbroadst: iirc this was an issue with the initial implementation where if I didn't specify the drainer of the `connectivity_cluster_run` then I got all sorts of weird threading issues. I'll give this a shot, but this just occurred to me.

src/clustering/administration/servers/auto_reconnect.cc
mbroadst: yes exactly

src/rpc/connectivity/cluster.cc
mbroadst: seems to make sense to me that if the remote host is refusing to connect to us that we would stop trying to connect with it. You know better than I do, but it seems like if you're getting a `INCOMPATIBLE_BUILD` or `INCOMPATIBLE_ARCH`, then there's no saving the situation.

src/rpc/connectivity/cluster.cc
mbroadst: yes this sound like a much better approach

src/rpc/connectivity/cluster.cc
danielmewes: I think we will get here (among other cases) if the remote host thinks that it already has a connection open. So at least some of the cases are temporary errors on the remote end, and we shouldn't permanently stop connecting.
If there's a version or architecture mismatch, we should bail out and return PERMANENT_ERROR above anyway, before we even receive the handshake response from the remote.

mbroadst:
just a little nudging reminder that we completed the review on this one and it probably needs another round of review

danielmewes:
Yep yep, I haven't forgotten about this. Just was busy with other things (TM). Sorry for the delay.

mbroadst:
no worries, just doing my weekly PR maintenance rounds ;)

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -82,24 +82,48 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
...
+
+        while (!interruptor.is_pulsed()) {
+            // This coroutine can keep running even after this function has returned
+            // or the auto_reconnector_t was destructed. So we must pass everything
+            // we need by value.
+
+            coro_t::spawn_now_dangerously(
+                [this, last_known_address, server, join_failed_out, last_known_address_out]() {
+                auto &connectivity_cluster_run_local = connectivity_cluster_run;
+                auto &join_delay_secs_local = join_delay_secs;
```
danielmewes: Both of these must be taken by value rather than reference, i.e. `auto ...` not `auto &...` since `this` might get destructed while we're in `join_blocking`.

I wonder by the way if we can set `this = nullptr;` after these two lines to catch access to it better? `this` might be `const` though, I'm not sure.

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -82,24 +82,48 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
...
+
+                join_results_t results =
                     connectivity_cluster_run->join_blocking(
```
danielmewes: I think this should use `connectivity_cluster_run_local` too.
I'm actually not sure if we need `connectivity_cluster_run_local`. We could maybe just use `connectivity_cluster_run` here, though using the `local` one is probably not a bad idea.

src/clustering/administration/servers/auto_reconnect.cc
```diff
@@ -82,24 +82,48 @@ void auto_reconnector_t::try_reconnect(const server_id_t &server,
...
     exponential_backoff_t backoff(50, 15 * 1000);
     try {
-        while (!interruptor.is_pulsed() && !join_failed.is_pulsed()) {
-            coro_t::spawn_now_dangerously([&]() {
-                join_result_t result =
+        // These can be safely passed into the coroutine below.
+        // They will be reset to `nullptr` by the assignment_sentry_ts when this function
+        // call ends.
+        std::shared_ptr<cond_t *> join_failed_out(new cond_t *(nullptr));
+        std::shared_ptr<peer_address_t *> last_known_address_out(new peer_address_t *(nullptr));
```
danielmewes: These lines are a bit long. Please break to 89 characters at most if possible.

mbroadst:
@danielmewes okay cleaned up those two issues. You were correct that `this` was const, so not assignable.

danielmewes:
Ok made a final pass. This looks good.

Merging, we'll run a few more tests on our end post-merge to make sure everything is working fine.

Thanks again for your work on this @mbroadst !

PR #5737, opened 2016-05-01
Array over size limit to mention array_limit and index. Resolve #3061

### LINKED ISSUE #3061: Array over size limit should provide more guidance

dineyw23:
Improved the error message to mention `array_limit` option and using an index. 
Kindly review. 

Some tests under `regression` fails but will tweak those once this has been reviewed.


danielmewes:
Thanks @dineyw23 . The new error message looks good to me.

Two suggestions:
1. We try to use two spaces after each period in ReQL error messages. This isn't done completely consistently right now, but new messages should follow the convention. So "Array over size limit `%zu`. To [...]" Should be "Array over size limit `%zu`.  To [...]".
2. The error message is currently copied four times. Maybe you could add a helper function `std::string format_array_size_error(size_t limit)` to `rdb_protocol/error.hpp` to contain the error message?

dineyw23:
I did not know about new convention, I'll do that right away.
For suggestion 2, I can add a helper function as per your recommendation but I am not really sure where exactly to that in `error.hpp`. There are some classes and I am not sure where exactly this function will fit in. 

Thanks @danielmewes .

danielmewes:
@dineyw23 I think the helper function should just be a global function in `error.hpp`. Don't worry about adding it to a class.

dineyw23:
Won't this message create a confusion if the user is using data explorer as it recommends to use `.run()` and we cannot use it in data explorer?

Maybe it's just me.

danielmewes:
@dineyw23 Yeah I can see that. I'm not sure if there's a great way for avoiding this.

We could just add a note:
`To raise the number of allowed elements, modify the `array_limit` option to `.run` (not available in the Data Explorer), or use an index.`

dineyw23:
Added helper function. Resolves #3061 
Kindly review.

Thanks,

src/rdb_protocol/error.cc
```diff
@@ -71,7 +71,12 @@ base_exc_t::type_t exc_type(const scoped_ptr_t<val_t> &v) {
     r_sanity_check(v.has());
     return exc_type(v.get());
 }
-
+std::string format_array_size_error(size_t limit) {
+    
+  return "Array over size limit `" + std::to_string(limit) + "`.  To raise the number "
+         "of allowed elements, modify the `array_limit` option to `.run`"
```
danielmewes: This is missing a space between `.run` and `(not available in the`

src/rdb_protocol/error.hpp
```diff
@@ -120,6 +120,9 @@ class bt_rcheckable_t : public rcheckable_t {
     backtrace_id_t bt;
 };
 
+//Helper function for formatting array over size error message.
```
danielmewes: For consistency with other comments, could you add a space between `//` and `Helper`?

src/rdb_protocol/error.hpp
```diff
@@ -150,18 +153,14 @@ class bt_rcheckable_t : public rcheckable_t {
         auto _limit = (limit);                                          \
         rcheck_datum((arr).size() <= _limit.array_size_limit(),         \
                      ql::base_exc_t::RESOURCE,                          \
-                     strprintf("Array over size limit `%zu`. To raise the number " \
-                               "of allowed elements, modify the `array_limit` " \  
-                               "option to `.run`, or use an index.", \
-                               _limit.array_size_limit()).c_str());     \
+                     strprintf("%s",format_array_size_error(_limit     \
```
danielmewes: You can actually replace the call to `strprintf` completely:
```
format_array_size_error(_limit.array_size_limit()).c_str()
```

danielmewes:
Thanks for updating. I added a few final comments. Should be good once you fix those.

src/rdb_protocol/error.hpp
dineyw23: Oops. Exactly!

danielmewes:
@dineyw23 It looks like some changes you did for another issue got rolled into this. Could you either open a new PR with a branch that has only the relevant changes, or revert the subsequent commits?

dineyw23:
Reverted the last irrelevant commit. 

Does this look good?

danielmewes:
Looks good to me. Thank you @dineyw23

Issue #4656, opened 2015-08-10
More descriptive error message for "Indexed order_by can only be performed on a TABLE or TABLE_SLICE"


danielcompton:
I was running 

```
r.db('local_ip').table('todos').filter({list: "home"}).orderBy({index: 'title'})
```

and got the error:

```
RqlRuntimeError: Indexed order_by can only be performed on a TABLE or TABLE_SLICE in:
r.db("local_ip").table("todos").filter({list: "home"}).orderBy({index: "title"})
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This was a confusing error to me, as I thought I was operating on a TABLE. After doing some search I understand the issue with secondary indexing and why this doesn't work. I think it would be helpful for users if the error message explained what type the indexed `order_by` was trying to operate on if it wasn't a TABLE, and maybe link to more docs about this.

mlucy:
Thanks for opening an issue.  I think you're right that we can probably make the error message clearer.

sanchopanca:
I agree, very confusing message. Link to documentation in error message would be helpful.

Issue #4615, opened 2015-08-03
Change error message for non atomic warning to point to function where it came from.


danielcompton:
When running this query:

```js
r.db('local_ip').table('todos').update({"done": r.db('local_ip').table('todos').filter({"done": false}).count()})
```
I get the error:

```
RqlRuntimeError: Could not prove function deterministic.  Maybe you want to use the non_atomic flag? in:
r.db("local_ip").table("todos").update({done: r.db("local_ip").table("todos").filter({done: false}).count()})
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It could be helpful to add to the error message: "Maybe you want to use the non_atomic flag* in the update function*" or something similar to point me to the source of the problem. Seeing this error and not knowing about non atomic I assumed it was a flag I needed to pass to run.

It could also be helpful to link to the rethinkdb docs for this error message for more explanation.

danielcompton:
Alternatively, narrowing down the function that threw the error could also make things easier to debug.

mlucy:
I agree, since it's complaining about the function I think it should underline the argument to `update` rather than the whole query.

PR #5741, opened 2016-05-03
Updated error message for Update query

AmeyDeshpande:
This request is for issues - [#4656](https://github.com/rethinkdb/rethinkdb/issues/4656) and [#4615](https://github.com/rethinkdb/rethinkdb/issues/4615)

### DIFF
--- a/src/rdb_protocol/terms/sort.cc
+++ b/src/rdb_protocol/terms/sort.cc
@@ -83,9 +83,9 @@ private:
         /* Add a sorting to the table if we're doing indexed sorting. */
         } else if (index.has()) {
             rcheck(tbl_slice.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE.");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure ORDER_BY comes before any transformations such as MAP or FILTER.");
             rcheck(!seq.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE.");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure ORDER_BY comes before any transformations such as MAP or FILTER.");
             sorting_t sorting = sorting_t::UNORDERED;
             if (auto _optarg = raw_term.optarg("index")) {
                 sorting = (_optarg->type() == Term::DESC) ?
--- a/src/rdb_protocol/terms/sort.cc
+++ b/src/rdb_protocol/terms/sort.cc
@@ -83,9 +83,9 @@ private:
         /* Add a sorting to the table if we're doing indexed sorting. */
         } else if (index.has()) {
             rcheck(tbl_slice.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure order_by comes before any transformations (such as map) or filters.");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE. Make sure order_by comes before any transformations (such as map) or filters.");
             rcheck(!seq.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure order_by comes before any transformations (such as map) or filters");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE. Make sure order_by comes before any transformations (such as map) or filters");
             sorting_t sorting = sorting_t::UNORDERED;
             if (auto _optarg = raw_term.optarg("index")) {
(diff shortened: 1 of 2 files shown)


danielmewes:
Thanks @AmeyDeshpande .

If you haven't done so yet, would you mind signing our CLA at http://rethinkdb.com/community/cla/ ? We require this for legal reasons before we can merge a pull request.

I'll comment on your actual patch in a bit.

AmeyDeshpande:
I have already signed CLA .

src/rdb_protocol/terms/sort.cc
```diff
@@ -83,9 +83,9 @@ class orderby_term_t : public op_term_t {
         /* Add a sorting to the table if we're doing indexed sorting. */
         } else if (index.has()) {
             rcheck(tbl_slice.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE.");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure ORDER_BY comes before any transformations such as MAP or FILTER.");
```
danielmewes: Is the `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON` a placeholder?

danielmewes: The names of terms should be lower case: "order_by", "map", "filter".

danielmewes: I think it would be a bit more correct to say "transformations (such as `map`) or filters.", since it's debatable whether a `filter` is considered a transformation or not.

src/rdb_protocol/terms/writes.cc
```diff
@@ -278,7 +278,7 @@ class replace_term_t : public op_term_t {
             rcheck(args->arg_is_deterministic(1) == deterministic_t::always,
                    base_exc_t::LOGIC,
                    "Could not prove argument deterministic.  "
-                   "Maybe you want to use the non_atomic flag?");
+                   "Maybe you want to use the non_atomic flag in the Update function");
```
danielmewes: I think "the Update function" is somewhat ambiguous here. Especially since you can get this error from using not only `update`, but also `replace` and I think even `insert`. We can try to come up with a better solution for this one, or leave it out of the PR at first and focus on the `order_by` error message above.

src/rdb_protocol/terms/sort.cc
mlucy: @danielmewes -- we usually use all-caps for type names in error messages.  We don't seem to write command names very often, but I thought we used the same convention for those.  (For example, `r.table('test').order_by(index: 'id').order_by(index: 'id')` produces `RethinkDB::RqlQueryLogicError: Cannot perform multiple indexed ORDER_BYs on the same table.`.)

Also, I'm not sure what the user-facing docs say, but `filter` is called a transformation internally, for the same reason `concat_map` is.

AmeyDeshpande:
@danielmewes I made changes in error message of order_by as you asked me to do.

src/rdb_protocol/terms/sort.cc
```diff
@@ -83,9 +83,9 @@ class orderby_term_t : public op_term_t {
         /* Add a sorting to the table if we're doing indexed sorting. */
         } else if (index.has()) {
             rcheck(tbl_slice.has(), base_exc_t::LOGIC,
-                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE.");
+                   "Indexed order_by can only be performed on a TABLE or TABLE_SLICE, not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`.  Make sure order_by comes before any transformations (such as map) or filters.");
```
danielmewes: These messages still contain the string `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`. What was your plan with that?
I think if you just remove the ", not on a `NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON`" part the change would be good.

src/rdb_protocol/terms/writes.cc
danielmewes: Any suggestions @AmeyDeshpande ? Do you want to remove this change from this PR for now, so we can merge the other one?

src/rdb_protocol/terms/sort.cc
danielmewes: @mlucy Hmm it looks like we are somewhat inconsistent on the capitalization of command names. I thought we used lower-case for commands, and upper-case for type names. But it looks like we're using both variants. (e.g. `ORDER_BY` in https://github.com/rethinkdb/rethinkdb/blob/next/src/rdb_protocol/term_walker.cc#L103 but `order_by` in this message or in https://github.com/rethinkdb/rethinkdb/blob/next/src/rdb_protocol/order_util.cc#L24).

I personally think the lower-case version looks nicer, but I don't particularly care either way. We should probably pick one format though and use it for future changes. Any opinions on which one it should be?

mlucy: I don't feel too strongly either as long as it's consistent.  I remember the convention being uppercase, so my guess would be that a lot of the older code uses uppercase and most of the lowercase examples are newer code.  (The lower-case `order_by` you linked was added by Nighelles earlier this year, which is consistent with that.)

If you think the lower-case is nicer I'd be happy to move to that.

AmeyDeshpande:
@danielmewes  I removed ", not on a NAME_OF_TYPE_ORDER_BY_WAS_CALLED_ON" part

danielmewes:
Thanks @AmeyDeshpande , this looks good to me.

PR #5743, opened 2016-05-03
Issue#5742: Bug: rethinkdb.errors.ReqlDriverError: Could not connect …

saadqc:
…to localhost:28015. Error: can't concat bytes to str

Hi,
I have fixed 5742 issue which was of string concatenation when connecting to redis with python driver.
Kindly review the change and merge PR.

### DIFF
--- a/drivers/python/rethinkdb/net.py
+++ b/drivers/python/rethinkdb/net.py
@@ -332,7 +332,7 @@ class SocketWrapper(object):
         except Exception as ex:
             self.close()
             raise ReqlDriverError("Could not connect to %s:%s. Error: %s" %
-                                  (self.host, self.port, ex))
+                                  (self.host, self.port, str(ex)))
 
     def is_open(self):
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks for the PR @saadqc .

If you haven't done so yet, would you mind signing our CLA at http://rethinkdb.com/community/cla/ and letting me know when you've done so? We require this for legal reasons before we can merge a pull request.

saadqc:
@danielmewes :+1: done!

danielmewes:
Thanks @saadqc . We're currently trying to understand better where the byte string is coming from and whether there might be a different place that we should put the string-conversion in.

We'll update you soon on whether we want to use your patch or fix the issue in a different way.

saadqc:
@danielmewes FYI, just CTRL + F in the code. You will find that the line is used multiple times and everywhere programmer added the str casting just missed at 335.

Issue #5758, opened 2016-05-10
Expose the Connection class from net.coffee (javascript driver)


gergelyke:
Hey,

we are building [Trace](https://trace.risingstack.com/), a Node.js monitoring platform.

For that, we'd like to instrument rethinkdb as well. We would need to access the Connection class from the `net` module. Is it something you would accept a PR for?

The change would be only this:

```javascript
// net.coffee

module.exports.Connection = Connection
```

```javascript
// rethinkdb.coffee

module.exports.net = net
```

Thanks,
Gergely

deontologician:
Do you need the connection class itself? We have a connection factory method that's exposed (`r.connect`)

gergelyke:
We can hook into that as well - that may be more complex to maintain down the road. The plan is to wrap the `_start` method of the `Connection` class.

If  you think there is a better way, I'd like to hear that :)

danielmewes:
Could you monkey-patch the `_start` method after getting the `Connection` object from `r.connect`?

I'm not opposed to exposing the `Connection` class. Just wondering how much easier it actually makes the modification.

gergelyke:
As a reference, you could take a look here:

The current mysql instrumentation looks this:
https://github.com/RisingStack/trace-nodejs/blob/master/lib/instrumentations/mysql.js#L49 - as we don't have access to the prototype, we had to it this way, and overwrite the connection factory. Currently, we would do something similar here as well.

On the other hand: https://github.com/RisingStack/trace-nodejs/blob/master/lib/instrumentations/mongodb.js - as mongodb exposes the prototype as well, it is really just wrapping that method, without have to modify the factory itself

danielmewes:
Ok, unless anyone else has any concerns, I think we should do this.

@gergelyke Would you be willing to open a pull request with the necessary changes? (note that we have a Contributor License Agreement: http://rethinkdb.com/community/cla/ )

gergelyke:
Great, thanks @danielmewes! 

Sure, happy to send one - so if I am right, I just have to fill it out, accept it, and that's it. Right?

danielmewes:
@gergelyke
> Sure, happy to send one - so if I am right, I just have to fill it out, accept it, and that's it. Right?

Yep, that's all that's required as far as the CLA is concerned.

Thanks for contributing!

danielmewes:
This was closed via https://github.com/rethinkdb/rethinkdb/pull/5759

PR #5759, opened 2016-05-11
chore(modules): exposing the net class

gergelyke:
Hey,

PR for #5758 - one question though. I couldn't find any unit/integration tests, did I miss something?

### DIFF
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -1665,3 +1665,8 @@ module.exports.connect = varar 0, 2, (hostOrCallback, callback) ->
                 resolve(result)
         create_connection(host, wrappedCb)
     ).nodeify callback
+
+# Exposing the connection classes
+module.exports.Connection = Connection
+module.exports.HttpConnection = HttpConnection
+module.exports.TcpConnection = TcpConnection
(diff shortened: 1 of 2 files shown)


drivers/javascript/net.coffee
```diff
@@ -1665,3 +1665,8 @@ module.exports.connect = varar 0, 2, (hostOrCallback, callback) ->
                 resolve(result)
         create_connection(host, wrappedCb)
     ).nodeify callback
+
+# Exposing the mail classes
```
tothandras: typo?

danielmewes:
Thanks for opening the PR @gergelyke .

You can run the JavaScript tests by `test/rql_test/test-runner -i js`. You'll need to compile the RethinkDB server before you do this, or alternatively pass in the location of the `rethinkdb` binary to use through the `-b` parameter.

@deontologician Could you take a look at @gergelyke 's patch when you have a few minutes?

deontologician:
This seems reasonable. @larkost can you remember the thing we exported with an underscore previously? We'd talked about exporting something ( I think it was the Connection class) to make testing easier, and landed on having it "privately" exported. I may be remembering something else though

larkost:
@deontologician I think you might be remembering either that we expose `_r` on the `Connection` class in Python, or that we export `_bluebird` on the `rethinkdb` object (the only thing we currently export) in the JavaScript driver.

deontologician:
That was it, thanks.

I am :+1: for merging this

danielmewes:
Back-ported into v2.3.x with 4b5d47a. This will ship with the next driver release.

PR #5778, opened 2016-05-18
Disable spellcheck on CodeMirror input

### LINKED ISSUE #5777: Upgrade CodeMirror
Newer versions of CodeMirror include `field.setAttribute("spellcheck", "false");` on the Data Explorer input which disables OS X Safari autocorrect. If it's too much work to upgrade, I can put in a PR to just patch in that one call.

danielcompton:
Safari on OS X will try to spellcheck if spellcheck is not disabled on the CodeMirror textarea.

Fixes #5777


danielmewes:
Thanks for the PR @danielcompton .

@deontologician does this look good to you?

deontologician:
Looks good to me

PR #5781, opened 2016-05-19
Cache compilation

thedrow:

### Description
Since travis added caching for all types of machines, using ccache might speed up the build a lot.
I'm guessing that clang won't agree with me so hang in there for a moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 # Environment
 language: cpp
+ccache: ccache
 compiler:
   - gcc
   - clang


thedrow:
Have you guys disabled Travis?

danielmewes:
Sorry for the confusion, we're indeed no longer using Travis for testing. We should probably delete the .travis.yml file from our repository.

I don't remember what the reason was for no longer using it. I think the resources might have been too restricted to run our tests.

thedrow:
Please consider using something else in that case. Jenkins has pull requests integration if you guys already use that.
If not, Snap-CI allows you to pick the machine size. It might be good enough for running the tests.

PR #5812, opened 2016-05-31
2.3.3 release notes

danielmewes:
@segphault

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,3 +1,64 @@
+# Release 2.3.3 (Fantasia)
+
+Released on 2016-06-01
+
+Bug fix release
+
+### Compatibility ###
+
+RethinkDB 2.3.3 servers can be mixed with older RethinkDB 2.3.x servers in the same
+cluster. We recommend that you run a mixed-version cluster only temporarily for upgrading
+purposes.
+
+No migration is required when upgrading from RethinkDB 2.3.x. Please read the
+[RethinkDB 2.3.0 release notes][release-notes-2.3.0] if you're upgrading from an
+older version.
+
+[release-notes-2.3.0]: https://github.com/rethinkdb/rethinkdb/releases/tag/v2.3.0
+
+### Windows support ###
+
+RethinkDB has been available for Windows since version 2.3.0. With RethinkDB 2.3.3, we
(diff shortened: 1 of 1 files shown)


NOTES.md
```diff
@@ -1,3 +1,64 @@
...
+
+No migration is required when upgrading from RethinkDB 2.3.x. Please read the
+[RethinkDB 2.3.0 release notes][release-notes-2.3.0] if you're upgrading from an
+older version.
+
+[release-notes-2.3.0]: https://github.com/rethinkdb/rethinkdb/releases/tag/v2.3.0
+
+### Windows support ###
+
+RethinkDB has been available for Windows since version 2.3.0. With RethinkDB 2.3.3, we
```
segphault: The current phrasing is fine, but here's some alternate language to consider:

RethinkDB 2.3.0 was the first version to include native Windows compatibility. In RethinkDB 2.3.3, the Windows port is ready to emerge from "beta" testing. We now officially support RethinkDB on the Windows platform alongside our existing support for Linux and Mac OS X. We're also extending our [commercial support][comm-support] services to include RethinkDB on Windows.

Although RethinkDB is now stable on Windows, there are still a few remaining limitations that we are actively working to address. We also haven't yet carried out as much performance tuning on the Windows port as we have on the Linux and OS X releases.

NOTES.md
```diff
@@ -1,3 +1,64 @@
...
+* JavaScript driver
+ * The `Connection` class is now exported from the RethinkDB JavaScript module (#5758)
+* Java driver
+ * Added the `clientPort` and `clientAddress` methods to the `Connection` class in the
+   Java driver (#5571)
+
+## Contributors ##
+
+Many thanks to external contributors from the RethinkDB community for helping
+us ship RethinkDB 2.3.3. In no particular order:
```
segphault: Maybe remove the "In no particular order" since there is only one. :-)

NOTES.md
danielmewes: Awesome! That's much nicer.

NOTES.md
danielmewes: Haha good catch. I just copied it over from the previous release and didn't think about it.

PR #5875, opened 2016-06-21
Making queries involving r.minval and other constants pickleable.

wwoods:

### Description
As part of a distributed batch-processing framework in Python, it can be convenient to form "root" queries for an application using RethinkDB and to pass those queries on to other servers that can modify or run them.  This allows the machines working with data from RethinkDB to request the data themselves, while a supervising system can distribute chunks of work that should go together.  For a setup like this to work, queries need to be pickle-able.  This pull request converts the constants in rethinkdb/query.py to a format that works with the pickle module.

Needs tests still, but I need guidance as for where those tests should go.

### DIFF
--- a/drivers/python/rethinkdb/query.py
+++ b/drivers/python/rethinkdb/query.py
@@ -266,29 +266,39 @@ class RqlConstant(ast.RqlQuery):
         return 'r.'+self.st
 
 # Time enum values
-monday      = type('monday', (RqlConstant,), {'tt':pTerm.MONDAY, 'st': 'monday'})()
-tuesday     = type('tuesday', (RqlConstant,), {'tt':pTerm.TUESDAY, 'st': 'tuesday'})()
-wednesday   = type('wednesday', (RqlConstant,), {'tt':pTerm.WEDNESDAY, 'st': 'wednesday'})()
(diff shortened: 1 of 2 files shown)


danielmewes:
Cool, thanks for the pull request @wwoods .

@larkost could you give @wwoods some pointers for how to write tests for this?

larkost:
Making queries `pickle`-able is a really good idea. But looking at this I think that rather that the root problem is that we don't expose the class we are using. Since the classes in question provide only a minor addition to `RqlConstant`, and all have just a single instance that we use, that it would be better to just make instances of `RqlConstant` that have the two attributes that we want them to have (`tt` and `st`).

I have a version that does this in branch `larkost/5875-python-pickling`, @wwoods would you mind seeing if that solves your use case as well? If you don't want to check it out through `git` I will be happy to send you a built version.

wwoods:
@larkost I'll admit I didn't run your branch, but looking at the changelog and tests for it, that seems like it should do the trick.  Feels cleaner than the approach I submitted as well.  If that gets merged in, all will be well.  Thanks!

larkost:
After internal review as CR 3669, this has been merged into `next` with 131a174c.

PR #5892, opened 2016-06-28
chore(packaging): speed up debian packaging by parallel builds

mbroadst:
### Description

This speeds up debian packaging by adding j flags to the make steps

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -14,11 +14,11 @@ clean:
 install: build
 	dh_clean
 	dh_installdirs
-	$(MAKE) DEBUG=$(DEBUG) DESTDIR=$(CURDIR)/debian/$(package) PVERSION=$(PVERSION) STRIP_ON_INSTALL=0 install
+	$(MAKE) DEBUG=$(DEBUG) DESTDIR=$(CURDIR)/debian/$(package) PVERSION=$(PVERSION) STRIP_ON_INSTALL=0 -j`nproc` install
 
 build:
 	./configure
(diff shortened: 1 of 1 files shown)


AtnNn:
It would be great if RethinkDB was faster to build and package by default.

However forcing ``-j`nproc``` may break for older versions of make (see #475). Make might also warn that

```
make[1]: warning: -jN forced in submake: disabling jobserver mode.
```

An alternative might be to pass `-jauto` to `dpkg-buildpackage` in [`mk/packaging.mk`](https://github.com/rethinkdb/rethinkdb/blob/next/mk/packaging.mk#L100). It would be easier for users who don't or can't build with `-j` to work around the forced parallelism.

mbroadst:
Ah right! you had mentioned that, I'll check it out

mbroadst:
hmm bummer:
```
mbroadst@retinoid:/zdata/rethinkdb (speed-up-debian-packaging *$)# dpkg-buildpackage -jauto
dpkg-buildpackage: unknown option or argument -jauto

Use --help for program usage information.
```

```
mbroadst@retinoid:/zdata/rethinkdb (speed-up-debian-packaging *$)# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.4 LTS"
```

srh:
Rebased onto next as of c5868f4dca7df0c50f7c7a4ec58a0ab02eee8dab.  No Debians that we support have that old version of make.  Having this would have made my life easier.

Issue #4630, opened 2015-08-05
Ruby EventMachine Handler callback API


nviennot:
Hi :)

I was trying to abort a ongoing query from an em_run() call, but the only way I found was to save the `caller` variable (which is a `QueryHandle`) from any of these callbacks: https://github.com/rethinkdb/rethinkdb/blob/f1f60df493f1c8ba1c510ad6ba72370282a6a010/drivers/ruby/lib/net.rb#L63-L92 and then call `close()` on the `query_handle`.

This might not be ideal, because these callbacks might not have had the time to be called, but `EM.next_tick()` did. Maybe an EM callback would want to close the query_handle already.

Here's what I suggest we do:
1) Remove all the `caller` arguments from the `Handler` callbacks.
2) Add a new callback `on_dispatch(query_handle)` that would be called right around here: https://github.com/rethinkdb/rethinkdb/blob/f1f60df493f1c8ba1c510ad6ba72370282a6a010/drivers/ruby/lib/net.rb#L530

This way we don't need a response from the server to terminate an ongoing query.

---

I've made NoBrainer compatible with EventMachine through the use of Fibers (so the code is still written in an asynchronous manner). If you are curious about how I use the `Handler` callbacks, check it out here: https://github.com/nviennot/nobrainer/blob/master/lib/no_brainer/query_runner/em_driver.rb

Further, I've written a self-contained Goliath server to demonstrate the usefulness of the Fiber abstraction: https://gist.github.com/nviennot/67a0795db7ccad973885
Maybe we could add this as a real-time tutorial to complement http://www.rethinkdb.com/docs/rails/

nviennot:
Nevermind, `em_run()` returns the `query_handle`. I feel dumb.
Well, I maintain my suggestion -- but no strong opinion :)

mlucy:
Thanks for the suggestion.  Tagging as a ReQL proposal since it's an API change and scheduling it for 2.2-polish so we can discuss it.

nviennot:
Also, it would be useful if the ruby driver could support async run() without requiring eventmachine. Or just support plain Fibers (e.g. switch the regular condition variable used in `@waiters` with https://github.com/igrigorik/em-synchrony/blob/master/lib/em-synchrony/thread.rb#L70-L125). Further, some people are fan of Celluloid, we might want to be able to integrate with such framework. So maybe the "wait" functionality should be abstracted away.

mlucy:
[...]

Issue #5620, opened 2016-04-06
decouple the driver building from `./configure`


larkost:
Currently if you want to build any of the drivers you have to also have run `./configure`. For Python and Ruby this is not really necessary, since they need no downloads, and there are no settings for them. There is a longer conversation to be had about the JavaScript driver (it needs a CoffeScript compiler and all its decencies), and some thinking to be done about the Java driver (obviously needs a Java compiler).

In addition to making people jump though possibly unneeded hoops if they want to work on the drivers, this also makes testing on 10.7 difficult, since we no longer support the compiler there, but still need to build the drivers to test.

PR #5973, opened 2016-07-19
Implement async_run as described in #4630

tinco:
### Description

This is an implementation of async_run, as described in #4630. The actual code in `net.rb` is rather non-idiomatic Ruby so I had to make some changes to the specific implementation. 

I also went ahead and implemented `em_run` using the new `async_run` to make sure it passes all tests.

Would be glad to hear your comments.

### DIFF
--- a/drivers/ruby/lib/net.rb
+++ b/drivers/ruby/lib/net.rb
@@ -134,6 +134,7 @@ module RethinkDB
       end
 
       EM_Guard.register(connection)
+      options[:query_handle_class] = EMQueryHandle
       yield
     end
 
@@ -215,14 +216,16 @@ module RethinkDB
       end
       handle_close
     end
-    def safe_next_tick(&b)
-      EM.next_tick {
-        b.call if !@closed
-      }
+
+    # Override this method with an async dispatch, making sure that
+    # when the block is run @closed == false
+    def guarded_async_run(&b)
+      raise "Must override QueryHandle#guarded_async_run"
     end
+
     def callback(res)
       begin
-        if @handler.stopped? || !EM.reactor_running?
+        if @handler.stopped?
           @closed = true
           @conn.stop(@token)
           return
@@ -231,15 +234,17 @@ module RethinkDB
                                   Response::ResponseNote::ATOM_FEED,
                                   Response::ResponseNote::ORDER_BY_LIMIT_FEED,
                                   Response::ResponseNote::UNIONED_FEED]) != []
-          if (res['t'] == Response::ResponseType::SUCCESS_PARTIAL) ||
-              (res['t'] == Response::ResponseType::SUCCESS_SEQUENCE)
-            safe_next_tick {
+
+          case res['t']
+          when Response::ResponseType::SUCCESS_PARTIAL,
+               Response::ResponseType::SUCCESS_SEQUENCE
+            guarded_async_run do
               handle_open
               if res['t'] == Response::ResponseType::SUCCESS_PARTIAL
                 @conn.register_query(@token, @all_opts, self) if !@conn.closed?
                 @conn.dispatch([Query::QueryType::CONTINUE], @token) if !@conn.closed?
               end
-              Shim.response_to_native(res, @msg, @all_opts).each {|row|
+              Shim.response_to_native(res, @msg, @all_opts).each do |row|
                 if is_cfeed
                   if (row.has_key?('new_val') && row.has_key?('old_val') &&
                       @handler.respond_to?(:on_change))
@@ -260,14 +265,14 @@ module RethinkDB
                 else
                   handle(:on_stream_val, row)
                 end
-              }
-              if (res['t'] == Response::ResponseType::SUCCESS_SEQUENCE ||
-                  @conn.closed?)
+              end
+              if res['t'] == Response::ResponseType::SUCCESS_SEQUENCE ||
+                  @conn.closed?
                 handle_close
               end
-            }
-          elsif res['t'] == Response::ResponseType::SUCCESS_ATOM
-            safe_next_tick {
+            end
+          when Response::ResponseType::SUCCESS_ATOM
+            guarded_async_run do
               return if @closed
               handle_open
               val = Shim.response_to_native(res, @msg, @all_opts)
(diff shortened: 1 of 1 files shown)


danielmewes:
This is awesome, thanks for the PR @tinco .

@mlucy could you review this please when you get a chance?

drivers/ruby/lib/net.rb
```diff
@@ -355,13 +398,23 @@ def em_run(*args, &b)
...
       end
 
-      EM_Guard.register(args[:conn])
-      args[:conn].run(@body, args[:opts], args[:block])
+      async_handler.callback = args[:block]
+      async_handler.connection = args[:conn]
+      async_handler.options = args[:opts]
+
+      async_handler.run do
+        async_handler.connection.run(@body, async_handler.options, async_handler.handler)
```
tinco: There's some serious class jealousy here. Do you think it would be prudent to make the AsyncHandler run function just take a callback, connection, options and body so we could just collapse these 6 lines?

tinco:
I just realized the below API for AsyncHandler might be more elegant, would you agree @mlucy? We wouldn't expose a bunch of attributes, reduce class jealousy in the async_run method and make the interface a little more explicit. The trade-off is that implementing AsyncHandlers would have to call `connection.run()` with all the arguments themselves, which requires them to know of that API.

```ruby
  # Extend this class to define a new way of running rethinkdb queries
  # asynchronously
  class AsyncHandler
    # This method is called with a block that runs a rethinkdb connection
    # synchronously
    def run(connection, options, callback, query_body)
      raise "Must override AsyncHandler#run"
    end

    # This method should return a handler that will deal with incoming messages
    def handler
      raise "Must override AsyncHandler#handler"
    end
  end
```

tinco:
I was tearing my hair out trying to get it to work without EM, turns out QueryHandle has a whole bunch of EM specific code that silently fails when EM is not present. This comment is just to have some deliverable for a few hours work, finding out what and why QueryHandle works like this is next..

tinco:
Well, it's official: this PR isn't winning any beauty contests.. I added a weird :query_handle_class option so the QueryHandle class can be overridden for async alternatives.

How is test coverage on the Ruby driver? I used this command: `RUBY_DRIVER=`which ruby` ./test-runner -i rb em` and it ran only 3 tests that didn't fail even though I had made an obvious mistake.

tinco:
Ok the tests are definitely not good enough. The previous test should've definitely broken all EM based code, but the circle-ci just said the full suite passed. Or doesn't circleci run the full test?

larkost:
I don't know who has setup CircleCI, but that is not how we do testing, and it does not seem to be testing anything. To run our Ruby tests [1] for checking the changes you made you should [2] run the following commands:
1. `./configure --allow-fetch` (see #5620 for why)
2. `./test/reql_tests/test-runner -i rb`

Note I am assuming that you are modifying the driver in `drivers/ruby`. You only need to play with `RUBY_DRIVER` if you are trying to modify one in another location (and `which ruby` would not give you the correct value for that).

[1] Yes, you should run all of them.
[2] I am including instructions from the root of our repository. The testing system actually does not care what your `pwd` is for most things.

danielmewes:
@tinco Have you been able to run the tests as @larkost described? Is this in a good state for us to check?

(also apologies for the delay from our end! We'll get to this very soon and in time for the 2.4 release)

tinco:
Hi Daniel, sorry for not checking yet, I just bought a house and things are a bit hectic out here :D I'll see if I can run the tests.

danielmewes:
@tinco No worries, congratulations on the house!

Depending on how much time you think you can spend, we could also take this over from here and do the testing (and any potentially necessary fixes) ourselves. Let me know if you'd prefer that. Your work on this PR is a big help in either case!

danielmewes:
Yep, I think r_http just requires some special modules, and `table` is currently broken in `next` (patch in progress). So that looks good! Thanks for running the tests.

We should be able to review your code shortly.

tinco:
Alright, just for the record I am very much ok with someone taking over. I just reviewed the code to find out why I thought the PR had become so ugly, and it's because of the QueryHandle class I had to extract. This makes it necessary for a new async handler to define an extra class that inherits from QueryHandle class, and subsequently inject that class into the options hash which is an awful interface. You can see one example of this in my implementation of the em_run in this PR, and below I've got the class I used in my Rails 5 spreadsheets tutorial:

```ruby
  class ConcurrentAsyncHandler < RethinkDB::AsyncHandler
    def run(&action)
      options[:query_handle_class] = AsyncQueryHandler
      yield
    end

    def handler
      callback
    end

    class AsyncQueryHandler < RethinkDB::QueryHandle
      def guarded_async_run(&b)
        Concurrent.global_io_executor.post(&b)
      end
    end
  end
```

Fixing this API requires a deeper refactoring, I did not there attempt it for this PR, as it would make it more opaque.

I think if this API is going to be public it has to be nicer, so I think you should not release it like this. If you have someone who could spend a few hours on this that would be great. I could have a go at refactoring the query handle, but it might take a while as work is keeping me busy, it would also make this PR significantly larger so it would be more difficult to review for you guys.

tinco:
So it seems someone from the company taking this over has suddenly become less likely ;) I can make the changes needed, is there someone available to do a review or will I be trusted to make a large PR? Did someone assume a lead role over the Ruby driver?

marshall007:
@tinco we don't have explicit leadership roles defined for the client drivers and you're definitely correct that it's unlikely we'll have the resources to implement this in the short term.

More than ever, we definitely encourage and welcome contributions though!

If you'd like to spend some time on the refactor, we're actively reviewing and merging PRs. I doubt this would make it would make it in the 2.4 release, but we could probably slate it for 2.4.1 or something shortly thereafter.

gabor-boros:
@tinco What do you think about implementing this change in the extracted driver? (https://github.com/rethinkdb/rethinkdb-ruby)

srh:
Closing this before we make v2.4.x the main branch of the repo.

I have made https://github.com/rethinkdb/rethinkdb-ruby/issues/24

Hopefully that issue won't live on zombie status forever...

PR #5974, opened 2016-07-19
Initial work to resurrect FreeBSD port

RodoMa92:
Since i need this database to work to being able to use SageMathCloud on my home server, i started to see if i was able to port this server to FreeBSD. 
I was thinking to not being enough capable of doing this, instead i’ve already achieved to being able to fully compile it and even to run tests for a good half of it. I've worked on this for the last two days, especially because i had a lot of problem with V8.
Logs:

[configure_logs.txt](https://github.com/rethinkdb/rethinkdb/files/372165/configure_logs.txt)
[test_logs.txt](https://github.com/rethinkdb/rethinkdb/files/372168/test_logs.txt)

Now it seems that two problem remains:
Some error about deleting folders, that makes starting the main executable impossible, for now.

```
error: Error in thread 1 in src/utils.cc at line 540:
error: Guarantee failed: [res == 0]  (errno 2 - No such file or directory) Fatal error: failed to delete '/tmp/rdb_unittest.QUYBCw/tmp'.

```

Seems to be some directory filesystem problem, but i've still to track it down.

The other seems to be a lot of failures on src/unittest/rpc stuff, same stuff above, need to understand from where are coming.

And lastly, some crash happens after a while with this error, while executing the tests:

```
error: Error in thread 1 in src/errors.cc at line 175:
error: std::terminate() called without any exception.
```

The backtrace is available on the logs.


src/arch/io/disk.cc
```diff
@@ -482,7 +482,11 @@ file_open_result_t open_file(const char *path, const int mode, io_backender_t *b
 #ifdef __linux__
     flags |= O_LARGEFILE;
 #endif
-
+#if defined(__FreeBSD__)
+    // FreeBSD does not have O_DSYNC / F_FULLSYNC. Instead, it relies on file
+    // system features like "soft updates" to reduce metadata syncing time.
+    flags |= O_SYNC;
```
danielmewes: Do you know if we actually need the O_SYNC here? Maybe just the `fsync` call in  line 632 is enough?
If in doubt, we can leave the O_SYNC in. It might be necessary for safety, and in the worst case will cost performance.

rodomar705: According to the fsync(2) FreeBSD manual, yes, it should be enough. I will try to comment it out and test it without as soon as i can.

src/unittest/extproc_test.cc
```diff
@@ -263,9 +263,9 @@ class segfault_job_t : public base_crash_job_t {
 
 private:
     static bool worker_fn(read_stream_t *, write_stream_t *) {
-        int bad_data = *reinterpret_cast<volatile int*>(NULL);
```
danielmewes: Why are these disabled?

mk/support/pkg/curl.sh
```diff
@@ -96,7 +102,7 @@ pkg_link-flags () {
             -lidn)    out `pkg link-flags libidn idn` ;;
             -lssl)    out_openssl ssl ;;
             -lcrypto) out_openssl crypto ;;
-            -ldl)     dl_libs=-ldl;; # Linking may fail if -ldl isn't last
```
danielmewes: We'll need to make sure that this change doesn't cause issues on other platforms. I assume the `-ldl` was there for a reason.

rodomar705: Whoops, indeed. Forgot to remove it. On FreeBSD was causing a linker error, so i had to comment it out. Tried to cleanup the modifications, but someone always escape to the checks 👍

src/unittest/extproc_test.cc
rodomar705: Was causing a compilation error on clang. I do not have the slightest idea why, but it wasn't liking them for some reason.

danielmewes: Interesting. Which version of Clang was this with?

We compile with Clang frequently on both Linux and OS X, so it should generally work. This might have stopped working with a recent version of Clang, in which case we'll need to fix it for Linux anyway. Do you remember the error message that Clang gave?

rodomar705: Yes, of course. A lot of the problem on the source was caused from using null pointer and calling on them reinterpret_cast to another type. A lot of headaches in V8 sources was for the same exact reason. The only semiworkaround was to use something different from nullptr or NULL. The version of clang is 3.4.1 on my system, that's seems quite old to me. I will try to update it first and then report back.

danielmewes: You can also try compiling with the `ALLOW_WARNINGS=1` option to `gmake`. Our default settings are quite strict when it comes to compiler warnings, and we usually treat them as errors.

rodomar705: It was a compilation error, not a warning. I've disabled the error on warning on the config file, to ease myself with the porting effort, but it wasn't liking it at all. I will post the full error if you need it.

danielmewes: The full error might be interesting, but no hurry. We could probably disable this test on FreeBSD if necessary.

rodomar705:
Fixed the evident problem on my edits, should be ok now.

rodomar705:
After two hours of reading, i still don't get where the problem lies, about filesystem traversal. The function nftw(3) from BSD claim to do exactly the same of the Linux counterpart, but it seems that it doesen't, because it claims always to fail with errno 2 - No such file or directory. The headers are identical, which is to be expected since it's a POSIX call. Tried to fix even the segfault changing the memory allocator to system, but it seems it's not random at all, fails always on ClusteringContractExecutor.SimpleTests, might be derived from all the other failures, especially on RDP. 

I think i've bite off more than i can chew, don't know if i can do this alone. I will try if i can understand where the problem lies, hoping for the best.

rodomar705:
Understood why it's failing, the folder is not there. Should be obvious, since it was clear inside the error, but since the tmp folder was correctly created, i assumed that creation was working. Need to debug it further.

rodomar705:
I'm literally clueless. It always fails deleting a folder named tmp folder inside the data folder that is correctly created. Inside the data folder the log file is present with the crash. So it seems to not being able to create folder recursively on the filesystem? Creating that by hand will throw correctly an exception signaling that the database file is not present, but the folder tmp created by hand remains. Don't know why, though. If anyone can have a look at it, i would be grateful.

rodomar705:
I relied on the easiest solution, and installed a Debian jail over the FreeBSD kernel, allowing to run the software without having to port it manually. Tried to build my change sets over a OS X machine, fails over fetching protobuf, even if i've changed just the make flag. I would love to have it natively on the system, but i'm not skilled enough to be able to do this by myself. Hope that someone will do it. You can close the pull request, if you desire so.

danielmewes:
Thanks a lot for your work on this @rodomar705 . Could you post the full (log) output of the RethinkDB server in the case where it segfaulted?

rodomar705:
Of course, but the segfault has been always thrown by the test system executable, the server itself is not able to startup at all for the filesystem bug I was talking earlier. Anyway, the full log including the segfault is present at the end of the test log posted above, right at the end where it says core dumped. I've took a look with gdb and printout the stack trace of the segfault, but it was literally a bunch of anonymous subfunction call, 30+ functions all with a ?? as name. I think, from what I've seen, some threading/fibers glitch, but as said I'm not skilled enough to say anithing conclusive.

fabianfreyer:
I'm not sure if this is already reflected in this work, but I've found a [very promising thread](https://lists.freebsd.org/pipermail/freebsd-ports/2015-September/thread.html#100404) on this issue over at the freebsd-ports mailing list. It might be worth taking a look through that.

src/arch/io/disk.cc
```diff
@@ -623,6 +629,11 @@ int perform_datasync(fd_t fd) {
     int res = fdatasync(fd);
     return res == -1 ? get_errno() : 0;
 
+#elif defined(__FreeBSD__)
```
srh: We should be calling fdatasync here, like with linux, no?

src/arch/io/disk.cc
srh: We definitely do not want O_SYNC.  Our syncing is done with explicit calls to fdatasync and such.

src/arch/io/disk.cc
MaRod92: The problem is that FreeBSD doesn't have a fdatasync(2) system call :(

jyrihovila: > The problem is that FreeBSD doesn't have a fdatasync(2) system call :(

https://www.freebsd.org/cgi/man.cgi?query=fdatasync&sektion=2

jyrihovila:
I would really, really love to see FreeBSD support finally being implemented.
RethinkDB is so amazing, and FreeBSD shines with it's jails system.
Could we (the people following this thread) try to work together to work out a solution?

src/arch/io/disk.cc
RodoMa92: And now it does :eyes:

RodoMa92:
Unfortunately I've migrated my systems to Linux now, and I don't really need this port up and running anymore, but it should be possible to pick it up from here and get it up and running :)

srh:
I've pointed this PR towards `v2.4.x` in preparation for making that the main branch of this repo.

With v2.4.x having replaced v8 with QuickJS, it might be worth revisiting the possibility of FreeBSD support.  Not "support" per se but just getting it working.

srh:
Closing now that #7088 got merged.  RethinkDB now builds on (some recent version of) FreeBSD.

Issue #6009, opened 2016-07-29
hidden memory leak when close Changefeeds cursor


QianJin2013:
hi, our team encountered a memory leak problem seems concerned with Cursor.close or connection.stop method, for Changefeeds cursor.

I am making a test case to prove it later. 

But now from static source analyze, i think there does have a problem. 
[drivers/java/src/main/java/com/rethinkdb/net/Cursor.java#L55](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Cursor.java#L55)
```
    public void close() {
        connection.removeFromCache(this.token);
        if (!error.isPresent()) {  /////here is the problem entry
            error = Optional.of(new NoSuchElementException());
            if (connection.isOpen()) {
                outstandingRequests += 1;
                connection.stop(this);     ////here is the problem 
            }
        }
    }
```

For a ChangeFeeds cursor, its error will not be set so can be called on next() method again.
Other type of cursor will be set error after got data to indicate it can not be used further more.

So, ChangeFeeds cursor will enter `!error.isPresent()` branch, finally call connection.stop(this) which in turn make a new cursor to send a close statement.
[drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L346](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L346)
```
    void stop(Cursor cursor) {
        // The server does reply to a stop request, though the response doesn't have a value.
        runQuery(Query.stop(cursor.token));
    }
 ```
The above code is strange. Previously it use `runQueryNoreply` which does not cause new cursor creation.

But the runQuery of course create new cursor.
[drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L283](https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L283)
```
Cursor cursor = Cursor.create(this, query, res, pojoClass);
```

So what is the problem? Calling close cause new cursor be added into cursorCache.
Repeatedly calling ChangeFeeds cursor query + close will cause memory leak.

QianJin2013:
I have created a pull request for this issue.  Hope someone check it as soon as possible.

danielmewes:
Thanks @QianJin2013 ! We'll try to get your fix in quickly.

Issue #5938, opened 2016-07-12
runNoReply in java does not work due to toCompletableFuture


maarest:
Hi,

I don't have a fix for the issue. If you try out this code, you'll see that it will hang:

```
   Connection conn = r.connection().hostname("localhost").port(28015).db("test").connect();
    System.out.println("Connected!");

    r.table("authors").insert(r.hashMap("name", "William Adama")
                    .with("tv_show", "Battlestar Galactica")
                    .with("posts", r.array(
                            r.hashMap("title", "Decommissioning speech")
                                    .with("content", "The Cylon War is long over..."),
                            r.hashMap("title", "We are at war")
                                    .with("content", "Moments ago, this ship received..."),
                            r.hashMap("title", "The new Earth")
                                    .with("content", "The discoveries of the past few days...")
                            )
                    )).run(conn);

    System.out.println("Inserted in Run!");
    r.table("authors").insert(
            r.hashMap("name", "William Adama")
                    .with("tv_show", "Battlestar Galactica")
                    .with("posts", r.array(
                            r.hashMap("title", "Decommissioning speech")
                                    .with("content", "The Cylon War is long over..."),
                            r.hashMap("title", "We are at war")
                                    .with("content", "Moments ago, this ship received..."),
                            r.hashMap("title", "The new Earth")
                                    .with("content", "The discoveries of the past few days...")
                            )
                    )).runNoReply(conn);
    System.out.println("Inserted in RunNoReply!");
```

This is apparently due to this line:
https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Connection.java#L211
`toCompletableFuture` that never completes in the future

This is I would say something quite important as we really need to have the massive performance improvement brought by runNoReply

Regards,
Stephane

bchavez:
:+1: 

Comparing to the C# driver, seems like the `awaitResponse` check is missing in the Java driver:

https://github.com/bchavez/RethinkDb.Driver/blob/master/Source/RethinkDb.Driver/Net/SocketWrapper.cs#L288

```
SendQuery:
   if( queryIsExpectingResponse ){
      //create the awaiter and keep track of it
   }
   //write the query to the network
```

danielmewes:
Looking into this. Hope we can release a fix in the next days together with the fix for https://github.com/rethinkdb/rethinkdb/issues/5980 .

bchavez:
[...]

PR #6010, opened 2016-07-29
Changefeeds cursor close cause new cursor cause memory leak

QianJin2013:
fix [hidden memory leak when close Changefeeds cursor · Issue \#6009](https://github.com/rethinkdb/rethinkdb/issues/6009)

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -344,7 +344,9 @@ public class Connection implements Closeable {
 
 
     void stop(Cursor cursor) {
-        // The server does reply to a stop request, though the response doesn't have a value.
+        // While the server does reply to the stop request, we ignore that reply.
+        // This works because the response pump in `connect` ignores replies for which
+        // no waiter exists.
         runQueryNoreply(Query.stop(cursor.token));
     }
 
--- a/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/Connection.java
@@ -345,7 +345,7 @@ public class Connection implements Closeable {
 
     void stop(Cursor cursor) {
         // The server does reply to a stop request, though the response doesn't have a value.
-        runQuery(Query.stop(cursor.token));
+        runQueryNoreply(Query.stop(cursor.token));
     }
 
     /**
--- a/drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
+++ b/drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
@@ -412,7 +412,7 @@ public class RethinkDBTest{
     }
 
     @Test
-    public void test() throws Exception {
+    public void test_Changefeeds_Cursor_Close_cause_new_cursor_cause_memory_leak() throws Exception {
         Field f_cursorCache = Connection.class.getDeclaredField("cursorCache");
         f_cursorCache.setAccessible(true);
 
--- a/drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
+++ b/drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
@@ -10,6 +10,7 @@ import net.jodah.concurrentunit.Waiter;
 import org.junit.*;
 import org.junit.rules.ExpectedException;
 
+import java.lang.reflect.Field;
 import java.time.OffsetDateTime;
 import java.util.Arrays;
 import java.util.List;
@@ -408,6 +409,25 @@ public class RethinkDBTest{
     @Test
     public void testNoreply() throws Exception {
         r.expr(null).runNoReply(conn);
+    }
+
+    @Test
+    public void test() throws Exception {
+        Field f_cursorCache = Connection.class.getDeclaredField("cursorCache");
+        f_cursorCache.setAccessible(true);
+
+        Map<Long, Cursor> cursorCache = (Map<Long, Cursor>) f_cursorCache.get(conn);
+        assertEquals(0, cursorCache.size());
+
+        Cursor c = r.db(dbName).table(tableName).changes().run(conn);
+
+        try {
+            c.next(1000);
+        } catch (TimeoutException ex) {
+        }
+        c.close();
+
+        assertEquals(0, cursorCache.size());
     }
 }
 


drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -345,7 +345,7 @@ public void runNoReply(ReqlAst term, OptArgs globalOpts) {
 
     void stop(Cursor cursor) {
```
QianJin2013: Someone changed this and i just change back. This method is only used by Cursor.java currently.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -345,7 +345,7 @@ public void runNoReply(ReqlAst term, OptArgs globalOpts) {
...
         // The server does reply to a stop request, though the response doesn't have a value.
-        runQuery(Query.stop(cursor.token));
+        runQueryNoreply(Query.stop(cursor.token));
```
QianJin2013: Someone changed this and i just change back. This method is only used by Cursor.java currently.

drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
```diff
@@ -409,5 +410,24 @@ public void testConcurrentCursor() throws TimeoutException, InterruptedException
     public void testNoreply() throws Exception {
         r.expr(null).runNoReply(conn);
     }
+
+    @Test
+    public void test_Changefeeds_Cursor_Close_cause_new_cursor_cause_memory_leak() throws Exception {
```
QianJin2013: test cursorCache is nothing after cursor.close

QianJin2013:
BTW, i found the Cursor.java has a weird `outstandingRequests` field.
https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Cursor.java#L55
```
protected int outstandingRequests = 0;
```

I can not understand the logic about it because 
* it actually set to 1 after run query.   Seems it does not follow common sense.
```
    protected void maybeSendContinue() {
        if(!error.isPresent()
                && items.size() < threshold
                && outstandingRequests == 0 ) {
            outstandingRequests += 1;
            this.awaitingContinue = connection.continue_(this);
        }
    }
```
* it does not get reset to 0 after cursor.close, in ChangeFeeds case, it will be 2 after close! 
```
    public void close() {
        connection.removeFromCache(this.token);
        if (!error.isPresent()) {
            error = Optional.of(new NoSuchElementException());
            if (connection.isOpen()) {
                outstandingRequests += 1;
                connection.stop(this);
            }
        }
    }
```
* it did not check against minus value  (over subtract reference).
```
    void extend(Response response) {
        outstandingRequests -= 1;    ########### i am worried, it takes reader too much time to check it ########
        maybeSendContinue();
        extendInternal(response);
    }
```

After all, the `outstandingRequests` is not a clear paring count.

drivers/java/src/test/java/com/rethinkdb/RethinkDBTest.java
```diff
@@ -409,5 +410,24 @@ public void testConcurrentCursor() throws TimeoutException, InterruptedException
...
+        Cursor c = r.db(dbName).table(tableName).changes().run(conn);
+
+        try {
+            c.next(1000);
+        } catch (TimeoutException ex) {
+        }
+        c.close();
+
+        assertEquals(0, cursorCache.size());
+    }
```
QianJin2013: test cursorCache is nothing after cursor.close

QianJin2013:
BTW, i found the Cursor.java has a weird `outstandingRequests` field, this value maybe cause potential leak problem.

https://github.com/rethinkdb/rethinkdb/blob/next/drivers/java/src/main/java/com/rethinkdb/net/Cursor.java#L55
```
protected int outstandingRequests = 0;
```

I can not understand the logic about it because 
* it actually set to 1 after run query.   Seems it does not follow common sense.
```
    protected void maybeSendContinue() {
        if(!error.isPresent()
                && items.size() < threshold
                && outstandingRequests == 0 ) {
            outstandingRequests += 1;
            this.awaitingContinue = connection.continue_(this);
        }
    }
```
* it does not get reset to 0 after cursor.close, in ChangeFeeds case, it will be 2 after close! 
```
    public void close() {
        connection.removeFromCache(this.token);
        if (!error.isPresent()) {
            error = Optional.of(new NoSuchElementException());
            if (connection.isOpen()) {
                outstandingRequests += 1;
                connection.stop(this);
            }
        }
    }
```
* it did not check against minus value  (over subtract reference).
```
    void extend(Response response) {
        outstandingRequests -= 1;    ########### i am worried, it takes reader too much time to check it ########
        maybeSendContinue();
        extendInternal(response);
    }
```

After all, the `outstandingRequests` is not a clear paring count.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
bchavez: Yuuuuppp. `runQuery` in `stop()` would cause the memory leak to come back. Nice hunting. :+1: 

IIRC, @danielmewes made the change in:
https://github.com/rethinkdb/rethinkdb/commit/91511142dd55328cfae120fc7500ba3e0e1b402f#diff-f233fec96ca1256a7793f216a89cb657R348

In reference to: https://github.com/rethinkdb/rethinkdb/issues/5938#ref-commit-9151114 ... da code be buggin' yo.

-----

:fire: :volcano: [***"We're building it up, to break it back down..."***](https://www.youtube.com/watch?v=dxytyRy-O1k)

bchavez:
Hey @QianJin2013 great work on tracking down that bug. You got mad (aka good) debugging skills tracking this one down. Your PR is :100: on point.

So, I think (and you probably already know) `outstandingRequests` basically acts as a small gate to prevent too many continues from being sent over the wire. I think `outstandingRequests` is mostly useful in `maybeSendContinue()` where it checks for `&& outstandingRequests == 0`.

By the time `close()` actually happens, `outstandingRequests` is forever `+1`. I think it's just an extra check so any crazy operations on the `Cursor` that happen after a `close()` won't accidentally fire off and send a continue. "Closed is closed," so says the shopkeeper...

I don't see a possible leak yet but maybe you could explain more about how you see `outstandingRequests` could cause a potential leak somewhere. Ultimately the continue callstack looks like `maybeSendContinue() -> conn.continue_() -> sendQuery` not necessarily going through the `runQuery` execution path.

I'll agree with you, though, using `outstandingRequests` as a gated counter makes the code harder to reason about. Perhaps it was originally designed to allow firing off multiple continues at a single time, but realistically, that's not what's happening in the current implementation... IIRC, there is only one continue over the wire at any given time. This is one of the reasons why I completely rewrote the `Cursor` implementation in the C# driver. The C# implementation doesn't rely on a gated `outstandingRequests` counter and it makes the code much more easy to reason about (at least I think) (and so far, fewer bugs I'm hoping). Fingers crossed. :smiley_cat: 

:art: :dash: [**Marshmello - CoLoUR**](https://www.youtube.com/watch?v=6UhNJC8OyVA)

deontologician:
outstanding requests comes from the python driver, where most of the net code was adapted from: 

https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/rethinkdb/net.py#L134

It's very possible it's not needed any more.

Thanks @QianJin2013 for the PR, I am checking it out now

danielmewes:
Yeah it seems like using a noreply run here works, assuming that the main read loop will ignore the extra response from the server for which no waiter can be found.

We should also change the comment above that line though and explain that we ignore the response.

drivers/java/src/main/java/com/rethinkdb/net/Connection.java
```diff
@@ -345,7 +345,7 @@ public void runNoReply(ReqlAst term, OptArgs globalOpts) {
 
     void stop(Cursor cursor) {
         // The server does reply to a stop request, though the response doesn't have a value.
```
danielmewes: Thanks first of all for your research on this and the bug fix @QianJin2013 !
My only suggestion is changing this comment to state something like this:
```
// While the server does reply to the stop request, we ignore that reply.
// This works because the response pump in `connect` ignores replies for which
// no waiter exists.
```

QianJin2013:
hi, i am back.  thanks all of you.
@bchavez 😄 thanks for your detail description. Ok, i will reference C# implementation when in case modify Cursor.java. 

@danielmewes I'v added your comment to source.

danielmewes:
Thanks again for the patch @QianJin2013 (and @bchavez for your comments!). Merging.

Hronom:
When this change will be in maven central?

PR #6024, opened 2016-08-02
Fix the build on case-sensitive macOS file systems

ilovezfs:
"Availability.h" not "availability.h" is the correct name of the header
in question. Fixing this spelling is necessary for a successful build
if the file system is case sensitive. With the incorrect spelling the
following build failure occurs on macOS:
  src/clustering/administration/main/cache_size.cc:15:10:
  fatal error: 'availability.h' file not found
  #include <availability.h>"

### DIFF
--- a/src/clustering/administration/main/cache_size.cc
+++ b/src/clustering/administration/main/cache_size.cc
@@ -12,7 +12,7 @@
 #include <unistd.h>
 
 #if defined(__MACH__)
-#include <availability.h>
+#include <Availability.h>
 #include <errno.h>
 #include <mach/mach.h>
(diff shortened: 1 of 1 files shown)


danielmewes:
Thanks @ilovezfs ! The patch looks good.

Have you had a chance to sign our CLA? http://www.rethinkdb.com/community/cla/

danielmewes:
We'll fix this ourselves in the next days...
Thanks for bringing this to our attention @ilovezfs .

ilovezfs:
Excellent

danielmewes:
Fixed in next via 7604f0f and v2.3.x via 3c623fb.

Issue #6043, opened 2016-08-09
asyncio loop type unavailable when installing via setup.py


byronh:
Installing the `rethinkdb` Python driver as a required package of a project's `setup.py` seems to make the `asyncio` loop type unavailable.

setup.py:
```python
import setuptools

setuptools.setup(
    name='example',
    install_requires=['rethinkdb'],
    ....
)
```
```bash
python setup.py install
```
```
Python 3.5.2 (default, Aug  8 2016, 20:46:29) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rethinkdb as r
>>> r.set_loop_type('asyncio')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/rethinkdb-2.3.0-py3.5.egg/rethinkdb/net.py", line 673, in set_loop_type
ValueError: Unknown loop type: 'asyncio'
```

However, when installing via pip, everything is fine:
```bash
pip install rethinkdb
```
```
Python 3.5.2 (default, Aug  8 2016, 20:46:29) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rethinkdb as r
>>> r.set_loop_type('asyncio')
>>>
```

I tried a number of versions including `2.3.0` and `2.3.0post1` through `2.3.0post5` and have replicated it in different clean virtual environments as well as docker images based on `python:3.5.2`.

My suspicion is that this is because of the logic at https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/rethinkdb/net.py#L676 which is looking for a file in a specific place in the file system - does this work from inside a `.egg`?

larkost:
I can confirm that that logic is the problem, and that the same problem occurs if use `easy_install` rather than `pip` (the only normal way of getting true `egg`s). I have a possible solution to this that I am testing now.

And it does not seem that in your setup that you will run into it, but FYI: there is logic in `setup.py` that will only include the `asyncio` file on versions of Python that support it. So you need to do your packaging with a version of Python that supports it (by default 3.4+).

larkost:
This is working in my local testing, so it is up for review in CR 3722 from branch `larkost/6043-loop-type-fix`.

larkost:
This is in `next` with bf4be17f and `v2.3.x` with a1054107. It will ship with the next version of the Python driver (probably 2.3.5).

PR #6077, opened 2016-08-26
RethinkDB 2.3.5 release notes

danielmewes:
@mglukhovsky or @dalanmiller, could you review please?

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,3 +1,87 @@
+# Release 2.3.5 (Fantasia)
+
+Released on 2016-08-26
+
+Bug fix release
+
+### Compatibility ###
+
+On 32-bit platforms and on Windows (64 and 32 bit), RethinkDB 2.3.5 servers should not
+be mixed with servers running RethinkDB 2.3.3 or older in the same cluster. Doing so can lead to
+server crashes when using the web UI or when accessing the `logs` system table.
+
+On 64-bit platforms, RethinkDB 2.3.5 servers can be mixed with older RethinkDB 2.3.x
+servers in the same cluster. We recommend that you run a mixed-version cluster only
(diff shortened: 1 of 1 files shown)


NOTES.md
```diff
@@ -1,3 +1,87 @@
+# Release 2.3.5 (Fantasia)
+
+Released on 2016-08-26
+
+Bug fix release
+
+### Compatibility ###
+
+On 32-bit platforms and on Windows (64 and 32 bit), RethinkDB 2.3.5 servers should not
```
danielmewes: This section is almost the same as for 2.3.4

NOTES.md
```diff
@@ -1,3 +1,87 @@
...
+* JavaScript driver
+ * Fixed the server nonce validation in the connection handshake (#5916)
+ * The `host` argument to `connect` is now optional (#5846)
+* Java driver
+ * Cursors now implement the `Closeable` interface (#5468)
+ * Fixed no-reply queries as run through `runNoReply` (#5938)
+ * Fixed a bug in the `reconnect` method (#5841)
+ * Fixed a memory leak in the `Connection` object that was caused by the driver not
+   properly cleaning up closed cursors (#5980)
+* Python driver
```
dalanmiller: What about `pip`?

danielmewes: It has always been available when installing through `pip`. This was just a problem with our install script when run outside of `pip`.

larkost: @danielmewes: probably the more concise way of saying it would be:
```The `asyncio` loop type is now available when run from a Python .egg file (#6043)```

danielmewes:
Merging this so I can start building. If anything else turns up, we can still change it for the GitHub releases page.

PR #6105, opened 2016-09-13
allow tests to be run on ruby 2.3

tinco:
### Description

This change simply adds Ruby 2.3 to the list of supported Ruby versions. Running the test suite has the same result as running it on Ruby 2.1.

The following tests fail (just as they do on 2.1):

```
== Failed 2 tests (of 159) in 260.47 seconds!
        SETUP   connections/r_http.rb2.3
        FAILED  polyglot/meta/table.rb2.3
```

Without this change the tests will not run at all, the runner exits with a message about RUBY_DRIVER being invalid and being unable to find a valid interpreter.


danielmewes:
Thanks for another pull request @tinco .

Assigning to @larkost for review.

larkost:
Because of a similar (but more frustrating) situation in `node.js`, I am reworking how we detect our interpreters, which winds up paving over this situation altogether. I will leave this open until that is checked in, but anticipate that I will not merge this.

This is what I am working on right now, so anticipate that will be in `next` in the next day or two.

tinco:
Did it get in? :)

srh:
Assuming it didn't make it in because I had no problem rebasing your commit.  This is next as of 19a8f1d1a2747104ba373eac82c48358f434f011, also cherry-picked to v2.4.x and v2.3.x. Thank you!

PR #6135, opened 2016-10-05
Avoid side effects on 'hostOrCallback'

athlite:
It's better to clone 'hostOrCallback' to avoid side effects.

If you run r.connect more than one time with the same config object,
it gets mutated in case 'username' is set instead of 'user'.

Then, the second time you run r.connect, with the same config object,
the driver will complain that you can't use both 'username' and 'user'
at the same time.

Object.assign on 'hostOrCallback' will fix this issue.

### DIFF
--- a/drivers/javascript/net.coffee
+++ b/drivers/javascript/net.coffee
@@ -1634,7 +1634,7 @@ module.exports.connect = varar 0, 2, (hostOrCallback, callback) ->
         # Otherwise, the `callback` variable is already correctly
         # holding the callback, and the host variable is the first
         # argument
-        host = hostOrCallback || {}
+        host = Object.assign {}, hostOrCallback
 
     # `r.connect` returns a Promise which does the following:
     #


srh:
There's a bug here, yes.  Object.assign is only supported by Node.js 4 and up.  Might that be a problem?  Anybody?

marshall007:
@srh Node v0.12.x	 and earlier (aka anything before v4.x) are well past their [LTS lifecycle](https://github.com/nodejs/LTS#lts-schedule1), so I have no objection to relying on newer runtimes.

srh:
A problem with this PR is that r.connect can take a string, and this breaks that behavior.

marshall007:
Yea we'll need to do something like:

```coffee
if typeof hostOrCallback is 'function'
    callback = hostOrCallback
else if typeof hostOrCallback is 'string'
    host = hostOrCallback
else
    host = Object.assign {}, hostOrCallback
```

athlite:
If I was to write vanilla js, it would be something like
```
var connect = function () {

    var args = Array.prototype.slice.call(arguments,0,2);
    var host = typeof args[0] === 'function' ? 
        {} : Object.assign({}, args.shift());
    var callback = args.shift();
    /* Promise connect ... which happily mutates host object  */
};
```

athlite:
How about:
```
connect = () ->
    args = Array.prototype.slice.call arguments,0,2
    { host, callback } = ( (hostOrCallback,callback)->
        if typeof hostOrCallback == 'function'
            host: {}, callback: hostOrCallback
        else if typeof hostOrCallback == 'string'
            host: hostOrCallback, callback: callback
        else
            host: Object.assign({}, hostOrCallback), callback: callback
    ).call(args[0],args[1])
    # connect Promise can happily mutate host
    # ....

```

srh:
Closing for #6575 which is just @marshall007's implementation.

PR #6142, opened 2016-10-13
Add separate arity tests for Ruby 2.3

katafrakt:
### Description

The syntax for ArgumentError has changed in Ruby 2.3 and because of that
tests for arity are not passing. This is a fix adding separate asserions for Ruby 2.3.

Because of the discussion in https://github.com/rethinkdb/rethinkdb/pull/6105 I did not add Ruby 2.3 to the list of testable versions in the PR as - from what I understand - it is going to change anyway.


srh:
I merged this to next as of e393655cdff99defad5e36094573894823a35247, cherry-picked into v2.3.x and v2.4.x.

Issue #5453, opened 2016-02-26
JavaScript `run` command returns a promise if conn is null/undefined even with a function


larkost:
The JavaScript driver is returning a promise if the supplied connection is `null` or `undefined`, even when a callback function is defined. For example:
```JavaScript
r.expr(1).run(null, function(err, value) { console.log('I should get here, but do not'); })
.catch(function(err) { console.log('Instead I get here') });
```
Without the `.catch` in the example above Node.js will send an `Unhandled rejection` notice to stderr, but the problem otherwise goes unhandled.

larkost:
Note that upcoming new tests catch this.

PR #6156, opened 2016-10-25
When connection is non-existential, continue to parse parameters.

Timer:
### Description

When connection is non-existential (`undefined` or `null`), we should continue to parse parameters. This allows options to treated as callback, fixing issue #5453 (the exception is then thrown properly).

### DIFF
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -70,7 +70,7 @@ class TermBase
         # Depreciated syntaxes are
         # optionsWithConnection, callback
 
-        if net.isConnection(connection) is true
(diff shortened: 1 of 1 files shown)


srh:
Closing.  Why now?  I'm cleaning out PR's targeting `next` in preparation for making `v2.4.x` the main branch of the repo.

It looks like this change is desirable and because the JS driver was split out to a separate repo, I've made an issue in its repo:  https://github.com/rethinkdb/rethinkdb-javascript/issues/7

Honestly it's not clear to me why this PR didn't get merged after it got created.  I'll try and revisit it after v2.4.2 is released.

PR #6157, opened 2016-10-26
Java - Jackson

mcjp78:
Now uses Jackson (Json parser)  for the java driver

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
+++ b/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
@@ -15,10 +15,10 @@ public class RethinkDB extends TopLevel {
         return Connection.build();
     }
 
-    public static void setObjectMapper(ObjectMapper mapper1){
+    public static void setObjectMapper(ObjectMapper mapper1) {
         mapper = mapper1;
     }
-    public static ObjectMapper getObjectMapper(){
+    public static ObjectMapper getObjectMapper() {
         return mapper;
     }
 }
--- a/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
+++ b/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
@@ -1,5 +1,6 @@
 package com.rethinkdb;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.rethinkdb.gen.model.TopLevel;
 import com.rethinkdb.net.Connection;
 
@@ -8,9 +9,16 @@ public class RethinkDB extends TopLevel {
     /**
      * The Singleton to use to begin interacting with RethinkDB Driver
      */
-    public static final RethinkDB r = new RethinkDB();
-
+    private static ObjectMapper mapper = null;
     public Connection.Builder connection() {
         return Connection.build();
+    }
+
+    public static void setObjectMapper(ObjectMapper mapper1){
+        mapper = mapper1;
+    }
+    public static ObjectMapper getObjectMapper(){
+        if (mapper != null) return mapper;
+        return new ObjectMapper();
     }
 }
--- a/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
+++ b/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
@@ -9,7 +9,7 @@ public class RethinkDB extends TopLevel {
     /**
      * The Singleton to use to begin interacting with RethinkDB Driver
      */
-    public static RethinkDB r = new RethinkDB();
+    public static final RethinkDB r = new RethinkDB();
     private static ObjectMapper mapper = new ObjectMapper();
     public Connection.Builder connection() {
         return Connection.build();
--- a/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
+++ b/drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
@@ -9,6 +9,7 @@ public class RethinkDB extends TopLevel {
     /**
      * The Singleton to use to begin interacting with RethinkDB Driver
      */
+    public static RethinkDB r = new RethinkDB();
     private static ObjectMapper mapper = new ObjectMapper();
     public Connection.Builder connection() {
         return Connection.build();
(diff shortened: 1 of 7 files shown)


PeterKnego:
Does this break existing API? If so then this should be optional.

drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
```diff
@@ -8,9 +9,16 @@
...
     public Connection.Builder connection() {
         return Connection.build();
     }
+
+    public static void setObjectMapper(ObjectMapper mapper1){
+        mapper = mapper1;
+    }
+    public static ObjectMapper getObjectMapper(){
+        if (mapper != null) return mapper;
+        return new ObjectMapper();
```
deontologician: If we're going to create a new mapper, we should probably assign it to `mapper` before returning it.

deontologician:
@PeterKnego I'm a bit torn on this. So in an ideal world, we'd be backwards compatible for a while. In practice though, this is a big maintenance burden and obviously now there's less manpower available for it.

Given that Jackson is better at this stuff than the hand-rolled system we have now, I think it's reasonable to replace it, even if it results in some small changes in client code.

mcjp78:
Right, I agree on POJO tests, I will run some. I can also just set the object mapper to a new instance and then return that.

mcjp78:
Update: ID annotation fixed and tested. POJO's tested. ObjectMapper get and set methods revised.

@deontologician its ready

deontologician:
Ok, this looks pretty good, but I'm going to run the tests before merging, which might take a little time since I haven't got it set up on my laptop here. Bear with me

drivers/java/src/main/java/com/rethinkdb/RethinkDB.java
```diff
@@ -8,9 +9,15 @@
     /**
      * The Singleton to use to begin interacting with RethinkDB Driver
      */
-    public static final RethinkDB r = new RethinkDB();
```
deontologician: This should be re-added

PeterKnego:
As noted: this will break stuff for a lot of people. You should not break APIs on the minor version, ideally even major versions should be backwards compatible.

drivers/java/src/main/java/com/rethinkdb/net/Util.java
```diff
@@ -77,94 +78,10 @@ public static String fromUTF8(byte[] ba) {
...
-    }
-
-    private static boolean areParametersMatching(Parameter[] parameters, Map<String, Object> values) {
-        return Arrays.stream(parameters).allMatch(parameter ->
-                values.containsKey(parameter.getName()) &&
-                        values.get(parameter.getName()).getClass() == parameter.getType()
-        );
+      T t = RethinkDB.getObjectMapper().convertValue(map,pojoClass);
+        return t;
     }
```
srh: The indentation is F'd up here.

srh:
Can somebody explain what the benefit of this patch is?

PeterKnego:
@srh Current Java driver has a limited ability do map values provided by DB (basically a JSON)  to Java objects (you don't wan to marshall/demarshall data coming/going to DB by hand, right?). 

This patch would provide a much better mapping abilities, as provided by Jackson, IMO the best JSON-to-Java-object mapper there is. So this is a good patch that would provide a much better functionality.

However, this will change current API (the data returned/sent to driver will be different) and will break things for current users. As a basic rule minor updates should not change API contract, even major updates should be backwards compatible.

My suggestion would be to make this optional via a config setting and announce this change, than later change make it default when major number is bumped.

danielcompton:
> However, this will change current API (the data returned/sent to driver will be different) and will break things for current users.

My understanding was that the Java driver was still kind of in beta. It's also relatively new, and missing features like async responses which means that many people aren't able to adopt it. I suspect that the number of people who would be affected *and* wouldn't have the engineering resources to update this would be low.

> As a basic rule minor updates should not change API contract, even major updates should be backwards compatible.

That's a good general principle, but may not work with the RethinkDB drivers, as they are versioned against the RethinkDB version, rather than using semver on their own. The RethinkDB protocol is also very stable and RethinkDB is backwards compatible with older protocols, so there is nothing stopping people from relying on the older driver while they migrate.

Especially considering the reduced manpower available for RethinkDB, anything that can reduce maintenance burden is a big win in my book.

mcjp78:
I am conducting tests and I have the requested changes ready to commit.
@PeterKnego, upon looking through the serialization, the JSON actually won't change.

srh:
Please see https://github.com/srh/rethinkdb/tree/sam/javajackson which has a style fix commit.

mcjp78:
@srh Sorry, I didn't see any official style guid :P. Personally, I use allman but I didn't know what here. anyway, can you explain to me why this matters? I am a bit of anstatte

chrisvariety:
@mcjp78 When you add code to an existing codebase you will want to match the existing conventions so all your code looks and reads the same as all the code around it. Doing so will make the codebase one cohesive unit. You'll find this in every open source project, and the majority of closed source projects as well.

I believe RethinkDB has an official style guide though I'm not sure it is public at the moment. In lieu of an official style guide, it should be easy to match the other code already existing in the repository.

srh:
I thought that Javadoc comments don't mean anything (to the Javadoc tool) inside a function body.  At least a quick google suggested that.

I'm not aware of a Java-specific RethinkDB style guide, but having the statements in a block be at the same level of indentation is a more or less universal standard for the same reason that they're indented in the first place.  Regarding spacing, one of the following happens:  Somebody goes and cleans it up later, polluting git blame history.  Or, before that, somebody reading the code will see the weird style, think to themselves, "Oh what is up with this, I better check this out," and start scrutinizing it for quality problems.  So now they're doing work that doesn't need to be done, because of a false bad code smell.  Good code should smell good, so that bad-smelling code is a more reliable signal.

mcjp78:
Ah I see, I will fix the styling then

mcjp78:
@srh  Would you mind checking again to make sure that I fixed everything?

marshall007:
LGTM - @srh or @deontologician can you guys glance over this once more?

deontologician:
LGTM

srh:
Looks good to me but my opinion on Java code isn't worth much.

deontologician:
Ok merging

PR #6162, opened 2016-10-29
Bugfix rwlock_in_line_t move constructor.

### LINKED ISSUE #6158: rwlock_in_line_t and new_mutex_t move constructor is broken.
This bug doesn't manifest in the existing codebase because every place of use undergoes copy elision.

The bug is that `rwlock_in_line_t` was given a default move constructor, when its fields are

```
    rwlock_t *lock_;
    access_t access_;
    cond_t read_cond_;
    cond_t write_cond_;
```

This means that after moving from a src object to a dest object, the src object will still have a non-null `lock_` field.  As a result, its destructor will behave badly.

The fix would be to make a non-default move constructor that sets lock_ to null.

This bug was introduced in commit 8589d792223979fae7ec74aed2ce830e21a9b3de and then still broken after commit f1b9aed5fb312f1c0237bf2a3f9da7c8849c96b6.

Edit: This would be fixed by the first commit in #6163

srh:

Fixes #6158.

### DIFF
--- a/src/concurrency/rwlock.cc
+++ b/src/concurrency/rwlock.cc
@@ -70,6 +70,16 @@ rwlock_in_line_t::rwlock_in_line_t(rwlock_t *lock, access_t access)
     lock_->add_acq(this);
 }
 
+rwlock_in_line_t::rwlock_in_line_t(rwlock_in_line_t &&other)
+    : lock_(other.lock_),
+      access_(other.access_),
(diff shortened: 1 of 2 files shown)


srh:
This is obviated by #6163, it's one of the commits there.  You could still reopen and merge this separately if you wanted to.

PR #6163, opened 2016-10-30
A bunch of changes.

srh:

This is a _bunch_ of changes.  You might want to pick and choose between them -- each commit is a stand-alone change, though maybe a couple depend on the previous one.  Some of the commits merely shuffle deck chairs (like pulling stuff out of utils.hpp).  I recommend viewing this pull request commit-by-commit.

(I'm not inclined to do the work to separate this into a bunch of commits and separate pull requests which then get merged together in an arbitrary order, sorry.)

I tested that all the commits compile on Windows.

This obviates pull request #6162, because it includes that commit.

Real questionable commits:
- Update OpenSSL version.

I don't know what the deal is there -- is that necessary?  Is it right?  It was necessary to get things compiling on Windows, and seems like the right thing.  (Is it the case that we don't fetch OpenSSL on other platforms?)
- Forked <blah> off utils.hpp.

Harmless changes, but not very useful in terms of build times.
- Added home-grown optional, removed boost::optional.

I'm confident in the quality of the code here, but it didn't affect build times much.  (Not compared to a boost::variant removal branch I made.)  Only about 2%.
- Fragmented datum_stream.hpp

Just touches a lot of lines and kills git blame history.  You might not want this one.  IIRC it made some stuff a lot less necessary to include in everything.


srh:
Added some more stuff:

- Word-wrapped STYLE.md,
- Made RPC mailbox types be variadic, instead of being generated by macros.
- (In the process, implemented std::tuple serialization.)

STYLE.md
```diff
@@ -29,92 +34,140 @@ This document describes guidelines that people new to the codebase tend not to i
...
+6. Don't use non-const lvalue references, except, perhaps, in return values.
 
-  > Reason: Somebody calling `foo(bar)` can currently expect `bar` not to be modified.  Pass a pointer to bar, if you want it to be mutable.
+  > Reason: Somebody calling `foo(bar)` can currently expect `bar` not to be
+    modified.  Pass a pointer to bar, if you want it to be mutable.
 
   > Exception: `std::swap` and STL-like `swap` methods, perhaps, and other peculiar situations.
 
-1. Don't make fields whose type is a reference type (const or non-const), and don't turn references into pointers in ways that might surprise the caller.
+7. Don't make fields whose type is a reference type (const or non-const), and
```
Qix-: > Don't make reference-type fields (regardless of const-ness), and

sounds much clearer to me. I was expecting another clause after this due to `whose`. Had to re-read it a few times to understand it.

STYLE.md
```diff
@@ -29,92 +34,140 @@ This document describes guidelines that people new to the codebase tend not to i
...
-  > The include-what-you-use errors exist to make you suffer -- but keeping awareness of this makes efforts to untangle header file dependences a lot less labor-intensive.
+  > The include-what-you-use errors exist to make you suffer -- but keeping
+    awareness of this makes efforts to untangle header file dependences a lot
+    less labor-intensive.
 
   > Bogus errors can be suppressed with `// NOLINT(specific/category)`.
 
-  > This script has caught several bugs (mostly race conditions caused by not using reentrant "_r" versions of libc functions).
+  > This script has caught several bugs (mostly race conditions caused by not
+    using reentrant "_r" versions of libc functions).
```
Qix-: Would you consider using `"&#95;r"` instead of `"_r"` here since it causes syntax highlighting for Markdown to wig out (since `_` is an alternative form for italic syntax)?

src/arch/runtime/event_queue/iocp.hpp
```diff
@@ -1,12 +1,15 @@
 #ifndef ARCH_RUNTIME_EVENT_QUEUE_IOCP_HPP_
 #define ARCH_RUNTIME_EVENT_QUEUE_IOCP_HPP_
 
+#include <stdint.h>
```
Qix-: Why not `cstdint`?

STYLE.md
srh: No, because that would be less readable.

src/arch/runtime/event_queue/iocp.hpp
srh: Everywhere in the codebase uses stdint.h.

Qix-: That doesn't make it right. [`stdint.h` is deprecated](http://stackoverflow.com/questions/13642827/cstdint-vs-stdint-h).

srh: That doesn't make it wrong.

chrisvariety: Best to match existing codebase convention.

Qix-: So it should be changed in the rest of the codebase, then. It's _deprecated_ - there is a functional impact by using improper includes. It goes beyond style or convention...

chrisvariety: Super, if that's correct then you should open a pull request for that change. It's not relevant to this pull though -- it doesn't make sense to change in just one place.

srh: > there is a functional impact by using improper includes.

@Qix- What is this impact and why is it bad?

Qix-: @srh Because it's not conformant to the C++ standard. Not using the correct headers creates potential incompatibilities with compilers.

srh: stdint.h is supported by POSIX and Windows so I don't see any potential for incompatibilities.  Using <cstdint> exports stuff in the std namespace so that's got a much higher cost of use.  Bad idea!

srh:
I can make a pull request for a hard durability speedup, that depends on some of these changes, once these changes get merged.

AtnNn:
These commits all LGTM except for switching out `boost::optional`.

Is it really worth doing? Is it more efficient? The only difference I see in the new API is making the type conversations explicit. But it introduces pain points such as having to be maintained and making the code less accessible to outsiders.

srh:
We wouldn't have had the bug that was fixed in https://github.com/rethinkdb/rethinkdb/pull/6163/commits/a27f9bdab22ba5be564a42566aa738f9bcb60144 with the existing optional implementation.

srh:
Also, the future of optional is std::optional from C++17, which this API moves in the direction of.

AtnNn:
Sounds good to me.

GCC 4.8 complains about an ambiguous call to `send`. I've fixed that in 8c6fad2558.

Issue #3583, opened 2015-01-16
r.now and changefeeds


AtnNn:
The current behaviour of `r.now` makes it less useful in the context of changefeeds.

If it was not constant, it would allow adding timestamps to the results and filtering based on the current time:

```
.changes().map(lambda row: [r.now(), row])
.changes().filter(lambda row: row.end_time > r.now())
```

Timer:
This would seem like a natural progression resulting from the addition of change feeds -- I would like to see this implemented as I need it for my application.

mlucy:
Changing the behavior of `r.now` in that way would make it non-deterministic, which would break some people's queries since it could no longer be used in atomic updates.  Maybe we could add an optarg to change the behavior?

shadowcodex:
Was wondering if this has been addressed yet. Here is my use case:

Example: Watching user action records on a conditional of (if action was less than 5 minutes ago). Pull a query that matches that, and if as time passes no records match anymore to emit an update. Thus easily showing what users are “Currently Active" without having to do a polling situation.

thelinuxlich:
+1 and I think for this to be really useful it would need to send changefeed events when the document leaves the time window({new_val: null, old_val: { ... }})

tatsujin1:
Yes, filtering using the current time would be very useful in some situations I have!

But I guess there would need to be two `r.now()`; one `r.query_start_now()` (i.e. the current one), and one `r.current_now()`, if you know what I mean?   :)

One simple case I have, or want to have, because doing it without this feature is complicated :), is this:
```
for change in table.filter(lambda doc: r.current_now() - doc['modtime'] > 600).changes().run():
    stuff...
```
That is, get the document once it has been "stable" (unmodified) for 10 minutes.
Completely awesome!  Would this be possible?    Pleeeeease? :D

danielmewes:
@tatsujin1 Good point. We indeed need to figure out a nice way for distinguishing the two variants since both can be useful as you say. Maybe with a separate term, or with the optarg (I think the former might be easier to implement).

thelinuxlich:
A third option(probably crazy), would be a lazy wrapper:

```
.changes().filter(lambda row: row.end_time > r.lazy(r.now()))
```

tatsujin1:
I like this.
In theory such a `lazy` operator might be useful in other situations as well. I have no idea, currently, what these might be, however. :)

[...]

PR #6191, opened 2016-12-07
Restore deterministic r.now

### LINKED ISSUE #6125: r.now() and the non_atomic flag
The query: `r.table("table").get(1).update { { "expires_at" => r.now } }`

fails with: `Could not prove argument deterministic.  Maybe you want to use the non_atomic flag?`

I know the query may not be deterministic, but the update can still be performed in an atomic fashion. Can  the use of `r.now()` not require the use of the `non_atomic` flag?

AtnNn:
Fixes #6125

Replaces the pseudo-boolean `deterministic_t` with a bitmask.

### DIFF
--- a/src/rdb_protocol/op.cc
+++ b/src/rdb_protocol/op.cc
@@ -286,7 +286,7 @@ const std::vector<counted_t<const term_t> > &op_term_t::get_original_args() cons
 deterministic_t op_term_t::is_deterministic() const {
     const std::vector<counted_t<const term_t> > &original_args
         = arg_terms->get_original_args();
-    deterministic_t det = DET_DETERMINISTIC;
+    deterministic_t det = deterministic_t::DETERMINISTIC;
     for (const auto &arg : original_args) {
         det |= arg->is_deterministic();
     }
@@ -352,21 +352,25 @@ bool bounded_op_term_t::open_bool(
     }
 }
 
-std::string deterministic_string(deterministic_t det) {
-    if (det == DET_DETERMINISTIC) {
-        return "DETERMINISTIC";
+void debug_print(printf_buffer_t *buf, deterministic_t det) {
+    if (det == deterministic_t::DETERMINISTIC) {
+        buf->appendf("DETERMINISTIC");
     }
     std::string ret;
-    if (det & DET_NONDET) {
-        ret += "NONDET ";
+    if (det & deterministic_t::NONDET) {
+        buf->appendf("NONDET ");
     }
-    if (det & DET_CONSTANT_NOW) {
-        ret += "CONSTANT_NOW ";
+    if (det & deterministic_t::CONSTANT_NOW) {
+        buf->appendf("CONSTANT_NOW ");
     }
-    if (det & DET_SINGLE_SERVER) {
-        ret += "SINGLE_SERVER ";
+    if (det & deterministic_t::SINGLE_SERVER) {
+        buf->appendf("SINGLE_SERVER ");
     }
-    return ret;
 }
+
+const deterministic_t deterministic_t::NONDET{1};
+const deterministic_t deterministic_t::SINGLE_SERVER{2};
+const deterministic_t deterministic_t::CONSTANT_NOW{4};
+const deterministic_t deterministic_t::DETERMINISTIC{0};
 
 } // namespace ql
--- a/src/rdb_protocol/op.cc
+++ b/src/rdb_protocol/op.cc
@@ -283,21 +283,17 @@ const std::vector<counted_t<const term_t> > &op_term_t::get_original_args() cons
     return arg_terms->get_original_args();
 }
 
-deterministic_t worst_determinism(const deterministic_t a, const deterministic_t b) {
-    return a < b ? a : b;
-}
-
 deterministic_t op_term_t::is_deterministic() const {
     const std::vector<counted_t<const term_t> > &original_args
         = arg_terms->get_original_args();
-    deterministic_t worst_so_far = deterministic_t::always;
+    deterministic_t det = DET_DETERMINISTIC;
     for (const auto &arg : original_args) {
-        worst_so_far = worst_determinism(worst_so_far, arg->is_deterministic());
+        det |= arg->is_deterministic();
     }
     for (const auto &arg : optargs) {
-        worst_so_far = worst_determinism(worst_so_far, arg.second->is_deterministic());
+        det |= arg.second->is_deterministic();
     }
-    return worst_so_far;
+    return det;
 }
 
 void op_term_t::maybe_grouped_data(scope_env_t *env,
@@ -356,5 +352,21 @@ bool bounded_op_term_t::open_bool(
     }
 }
 
+std::string deterministic_string(deterministic_t det) {
+    if (det == DET_DETERMINISTIC) {
+        return "DETERMINISTIC";
+    }
+    std::string ret;
+    if (det & DET_NONDET) {
+        ret += "NONDET ";
+    }
+    if (det & DET_CONSTANT_NOW) {
+        ret += "CONSTANT_NOW ";
+    }
+    if (det & DET_SINGLE_SERVER) {
+        ret += "SINGLE_SERVER ";
+    }
+    return ret;
+}
 
 } // namespace ql
--- a/src/rdb_protocol/op.cc
+++ b/src/rdb_protocol/op.cc
@@ -286,12 +286,12 @@ const std::vector<counted_t<const term_t> > &op_term_t::get_original_args() cons
 deterministic_t op_term_t::is_deterministic() const {
     const std::vector<counted_t<const term_t> > &original_args
         = arg_terms->get_original_args();
-    deterministic_t det = deterministic_t::DETERMINISTIC;
+    deterministic_t det = deterministic_t::DETERMINISTIC();
     for (const auto &arg : original_args) {
-        det |= arg->is_deterministic();
+        det = det.join(arg->is_deterministic());
     }
     for (const auto &arg : optargs) {
-        det |= arg.second->is_deterministic();
+        det = det.join(arg.second->is_deterministic());
     }
     return det;
 }
@@ -351,26 +351,5 @@ bool bounded_op_term_t::open_bool(
               key.c_str(), v->trunc_print().c_str());
     }
 }
-
-void debug_print(printf_buffer_t *buf, deterministic_t det) {
-    if (det == deterministic_t::DETERMINISTIC) {
-        buf->appendf("DETERMINISTIC");
-    }
-    std::string ret;
-    if (det & deterministic_t::NONDET) {
-        buf->appendf("NONDET ");
-    }
-    if (det & deterministic_t::CONSTANT_NOW) {
-        buf->appendf("CONSTANT_NOW ");
-    }
-    if (det & deterministic_t::SINGLE_SERVER) {
-        buf->appendf("SINGLE_SERVER ");
-    }
-}
-
-const deterministic_t deterministic_t::NONDET{1};
-const deterministic_t deterministic_t::SINGLE_SERVER{2};
-const deterministic_t deterministic_t::CONSTANT_NOW{4};
-const deterministic_t deterministic_t::DETERMINISTIC{0};
 
 } // namespace ql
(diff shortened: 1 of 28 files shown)


src/rdb_protocol/order_util.cc
```diff
@@ -33,7 +33,7 @@ build_comparison(const term_t* target,
 
     counted_t<const func_t> comparison_function = arg->as_func(GET_FIELD_SHORTCUT);
     rcheck_target(target,
-                  comparison_function->is_deterministic() != deterministic_t::no,
+                  (comparison_function->is_deterministic() &~ DET_SINGLE_SERVER) == DET_DETERMINISTIC,
```
srh: Sorry, I don't know of any &~ operator in C++.

src/rdb_protocol/term.hpp
```diff
@@ -21,7 +21,7 @@ class table_t;
 class table_slice_t;
 class var_captures_t;
 class compile_env_t;
-enum class deterministic_t;
+typedef int deterministic_t;
```
srh: Instead this could be wrapped in a struct?

src/rdb_protocol/op.cc
```diff
@@ -356,5 +352,21 @@ bool bounded_op_term_t::open_bool(
     }
 }
 
+std::string deterministic_string(deterministic_t det) {
```
srh: Nothing uses this function `deterministic_string`.

src/rdb_protocol/op.hpp
```diff
@@ -18,22 +18,24 @@ namespace ql {
 
 class func_term_t;
 
-// Note that worst_determinism depends on the order of these; if you want to add
-// a new value, make sure worst_determinism is properly defined for it.
-enum class deterministic_t {
```
srh: Is it now possible for `DET_NONDET | DET_SINGLE_SERVER` to be a value of deterministic_t?  Is it possible for `DET_NONDET` to be a value?

src/rdb_protocol/order_util.cc
srh: So this was saying the value is either single-server or deterministic.

But now in a convoluted way, it's saying that the value is either DET_SINGLE_SERVER or DET_DETERMINISTIC, but not DET_CONSTANT_NOW.

AtnNn: And not DET_NONDET either.

src/rdb_protocol/term.hpp
AtnNn: Done.

src/rdb_protocol/op.cc
AtnNn: I renamed it to `debug_print`

src/rdb_protocol/op.hpp
AtnNn: Yes

srh:
I think the main problem afaict is that deterministic_t has multiple values that mean the same thing.  The encapsulation of deterministic_t doesn't really do anything because it's still bitmasking stuff.

AtnNn:
@srh `deterministic_t` is a set of possibly non-deterministic operation types. I use a bitset to represent them because `std::set` would be overkill.

It used to be an enum with a `worse_determinism` ordering, but that ordering won't be total anymore because of the introduction of non-deterministic `r.now`.

Given `det` the set of operation types present in a query and `mask` the set of deterministic operations in the current environment, the set `det & ~mask` is empty if the query is deterministic.

Because `mask` will never contain `NONDET`, some values of `det` will mean the same thing, but I don't think that's a problem.

srh:
I think that doing it this way is really confusing.  It's hard to keep track of what it means when these bitmask operations are done, and it's hard to understand what the meaning of an assert_deterministic(MASK, ...) call is supposed to mean.  You have to negate and bitwise and stuff in your head.

Could is_deterministic take a boolean indicating whether r.now() terms are to be deemed deterministic, and then produce a {no,single_server,always} result?

src/rdb_protocol/terms/seq.cc
```diff
@@ -461,18 +461,10 @@ struct rcheck_transform_visitor_t : public bt_rcheckable_t,
...
-                // ok
-                break;
-
-            case deterministic_t::no:
-                rfail_src(f.get_bt(),
-                          base_exc_t::LOGIC,
-                          "Cannot call `changes` after a non-deterministic function.");
-
-            default: unreachable();
+        if((f.compile_wire_func()->is_deterministic() & ~deterministic_t::SINGLE_SERVER) != deterministic_t::DETERMINISTIC) {
```
srh: Why is r.now() non-constant here?  (I have no idea whether it should be.)

src/rdb_protocol/order_util.cc
```diff
@@ -33,7 +33,7 @@ build_comparison(const term_t* target,
 
     counted_t<const func_t> comparison_function = arg->as_func(GET_FIELD_SHORTCUT);
     rcheck_target(target,
-                  comparison_function->is_deterministic() != deterministic_t::no,
+                  (comparison_function->is_deterministic() & ~deterministic_t::SINGLE_SERVER) == deterministic_t::DETERMINISTIC,
```
srh: Why does this not permit r.now() to be constant?  (I have no idea whether it should.)

srh:
I made a branch at https://github.com/srh/rethinkdb/commits/sam/deterministic_now which adds a few commits, that gives deterministic_t a tighter interface (and fixes a compile bug, and makes the names more like what they were originally).  The most important difference is that it removes operator== and operator!= so that functionally identical deterministic_t values are no longer distinguishable.

It doesn't address my last two questions, about the correctness of changes in order_util.cc or seq.cc.

thelinuxlich:
#3583

srh:
That's a ReQL proposal that this change partially implements.  `r.now()` becomes reevaluated for each change in .changes().  Note that as implemented it's *not* the timestamp at which the change happens -- it's some other timestamp that gets computed later.  That's pretty weird.

The other change that this introduces is that r.now() can't be used *before* .changes() is called.  If you want changes that refer to anything referring to a future time, or a past time, or anything over 3 weeks old, now you can't do that using r.now().

srh:
Another thing that's broken is queries that sort using r.now().

For example, if you want to order rows by the absolute distance of the time from some point in history:
`r.table('foo').orderBy(r.row('time').sub(r.now()).do(function (x) { return r.branch(x.gt(0), x, r.expr(0).sub(x)); }))`

thelinuxlich:
> That's a ReQL proposal that this change partially implements. r.now() becomes reevaluated for each change in .changes(). Note that as implemented it's not the timestamp at which the change happens -- it's some other timestamp that gets computed later. That's pretty weird.
The other change that this introduces is that r.now() can't be used before .changes() is called. If you want changes that refer to anything referring to a future time, or a past time, or anything over 3 weeks old, now you can't do that using 

Weirdness aside, this is very useful, it's actually a feature! But wouldn't the disallow of r.now() before changes() be very counter-intuitive(and also break production code)?

thelinuxlich:
If I understood correctly, this would allow something like the continuous queries of InfluxDB:

https://docs.influxdata.com/influxdb/v0.9/query_language/continuous_queries/

srh:
Yeah, `r.now()` before changes() would break stuff, if there's code that uses it.

After .changes(), it was always broken (it used a hard-coded value).  And then either `r.now()` or some other command name `r.write_timestamp()` could be used.  I think it should be disabled for now, because having `r.now()` change meaning depending on which side of `.changes()` is probably a bad idea.

srh:
Working on this.

AtnNn:
@srh are you still working on this?

After thinking about this for a while, I think the terminology is wrong. `is_deterministic` should return a `bool` and depend on the current environment.

If we implement that with a bitset, `term_t` could have a `nondet_set()` method and `env_t` could have a `det_set()` method, and `term->is_deterministic(env)` ⇔ `(bool)(term->nondet_set() & ~env->det_set())`

srh:
Oh sorry, I am still working on this and should be done today (using the technique of "actually" working on it).

I don't want to rename it just because I want to keep a clean diff for now.  But in the future I don't want to have to construct an `env_t` just to see if a query is deterministic.

srh:
I have fixes done in the server (in my branch: https://github.com/srh/rethinkdb/tree/sam/deterministic_now ).  Comparison functions are made to allow `r.now()`, and it is allowed before `changes()`, disallowed after `changes()`.

I guess some automatic test cases should be added too, but that's not done yet.

srh:
Added automatic test cases.  So that's done and up in that branch.

srh:
I've got a version of my branch above that has been rebased on next: https://github.com/srh/rethinkdb/tree/sam/deterministic_now_rebased

The rebasing process has some janitorial work done, related to the split-up of datum_stream.hpp.

AtnNn:
Thanks @srh, that `deterministic_t` looks easier to use. I'm taking a closer look at the code.

AtnNn:
@srh I added your changes to this PR. I also fixed a GCC 5.4 complaint about a `constant_now` name conflict between the enum  and method. Does it LGTY?

I like the idea of erroring if `r.now` is used after `changes`, but perhaps that should be a ReQL proposal first.

AtnNn:
I'm not sure disabling `r.now` after `.changes` is desired. There seems to be legitimate use cases for it, for example:

`.changes().map(r.row('new_val')('updatedAt').sub(r.now()))`

It also introduces confusion. With your patch `.changes().filter(_ => r.now())` fails but `.changes().filter(r.now())` works.

srh:
The problem with `.changes().map(r.row('new_val')('updatedAt').sub(r.now()))` is that you can come up with another example where people expect `r.now()` to refer to whatever time it is when the changefeed arrives.  And there's a workaround, `r.now().do(x => r.row('new_val')('updatedAt').sub(x)`.

AtnNn:
It's not innocuous (it might break existing queries) and it isn't obvious to me that is is a net improvement:

* Users might expect a different behaviour, but they can't have it. The current behaviour of `r.now()` after `changes` is documented, it is the same as everywhere else (except for write hooks) and has valid use cases.
* If users start seeing errors about `r.now()` after `changes` they might be more inclined to think that its use elsewhere isn't constant (such as after a long-running filter), because it doesn't error.
* Also, the undocumented evaluation order of ReQL combined with your change leads to inconsistent results, the like of which I would prefer to remove instead of add. For example:

Query | Result
-- | --
`.changes().filter(_ => r.now())` | Error
`.changes().filter(r.now())` | Okay
`.changes().merge(_ => ({t: r.now()}))` | Error
`.changes().merge({t: r.now()})` | Error

I would love to hear other opinions on this, which is why I suggested making it a ReQL proposal.

srh:
Then if you want to, make it so that `r.now()` behaves after `changes` the same as it does now.  I think the behavior's an un-thought-out artifact of implementation details of the interpreter though.

srh:
Note that to fix this, the commit you'll want to replace is everything in de3f25e1c5ab52a875d01e2e5d69cba2de8fa257.

srh:
Disallowing `r.now()` after `changes()` can get really gross when you realize that queries like this are legit:

`r.branch(r.random().gt(0.5), r.table('foo'), r.table('foo').changes()).merge({'z': r.now()})`

Disallowing `r.now()` after `changes()` would force the evaluation order, especially if we try to fix `.changes().filter(EXPR)` so that `EXPR` cannot use `r.now()` at all.  So that's another reason to just make it constant everywhere.

src/rdb_protocol/terms/writes.cc
```diff
@@ -313,7 +315,7 @@ class replace_term_t : public op_term_t {
             args->arg(env, 1)->as_func(CONSTANT_SHORTCUT);
         if (!nondet_ok) {
             // If this isn't true we should have caught it in the `rcheck` above.
-            rassert(f->is_deterministic() == deterministic_t::always);
+            rassert(f->is_deterministic().test(single_server::no, constant_now::yes));
```
srh: Compilation error here.

danielmewes:
Am I correct that this change ended up as part of https://github.com/rethinkdb/rethinkdb/pull/6396 and we no longer need this PR?

srh:
Yes, that's correct.

srh:
Closing.

PR #6198, opened 2016-12-16
Misc makefile fixes

AtnNn:
- Don't unnecessarily build the web assets when packaging.
- Don't build RethinkDB twice when building Debian and Ubuntu packages
- Ignore warnings in generated C++ files

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -11,7 +11,7 @@ clean:
 	dh_clean
 	if [ -e build ] ; then rm -rf build ; fi ;
 
-install: build
+install:
 	dh_clean
 	dh_installdirs
 	$(MAKE) DEBUG=$(DEBUG) DESTDIR=$(CURDIR)/debian/$(package) PVERSION=$(PVERSION) STRIP_ON_INSTALL=0 install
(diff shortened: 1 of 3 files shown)


packaging/debian/rules
```diff
@@ -11,7 +11,7 @@ clean:
 	dh_clean
 	if [ -e build ] ; then rm -rf build ; fi ;
 
-install: build
+install:
```
Qix-: Isn't this innocuous if running `make build install` anyway?

srh:
LGTM

packaging/debian/rules
AtnNn: Hi @qix-. Thanks for taking a look at this change. You should stop by http://slack.rethinkdb.com #engineering if you would like to get more involved.

`dpkg-buildpackage` calls both targets separately, which currently causes rethinkdb to be built twice when packaging. See http://man7.org/linux/man-pages/man1/dpkg-buildpackage.1.html

Qix-: Gotcha, that's good to know. I'm on Rethink's slack, too - didn't know about the #engineering channel though.

Thanks!

PR #6203, opened 2016-12-21
Fix use-after-move bug in lru_cache

### LINKED ISSUE #6202: `match' sometimes returns the wrong result

```
```

returns

```
[[
  null
], [
  null,
  null
]]
```

AtnNn:
This fixes #6202

This fix should make empty regexp patterns behave correctly and allow all other regexp patterns to hit the cache.

### DIFF
--- a/src/containers/lru_cache.hpp
+++ b/src/containers/lru_cache.hpp
@@ -4,6 +4,7 @@
 #include <map>
 #include <list>
 
+#include "debug.hpp"
 #include "errors.hpp"
 
 template <typename K, typename V>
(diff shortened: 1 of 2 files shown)


srh:
Does this mean that secondary indexes which had a .match could be in a broken, nonsensical state?

AtnNn:
I hadn't thought of that. Indexes that call `match` with an empty string might be broken If `env_t` gets reused to compute indexes. A quick look at the code makes me think that they aren't being reused.

srh:
Secondary indexes are supposed to use their own `env_t`, yeah.

Okay, I get it now -- empty regexes are going to get clobbered and point at the wrong iterator.  So secondary index functions that use an empty regex, followed by a non-empty regex, are going to run into trouble.

srh:
As far as I can tell, the patch in #6241 broke secondary index compatibility around this in 2.3.6, for those that were using empty regular expressions.

So now we can't make 2.4 offer a backwards-compatible implementation layer in either direction.

srh:
See #6562.

Issue #6137, opened 2016-10-06
Evaluate options for project stewardship, non-profit


eloquence:
A straightforward possibility is to apply for membership in the Software Freedom Conservancy, which offers important services such as the ability to accept donations:

https://sfconservancy.org/members/services/

Another path would be to join Apache:
https://incubator.apache.org/

(This would require relicensing, if I'm interpreting https://incubator.apache.org/ correctly, so should probably be taken care of ASAP if that's the path you want to take.)

dustinkingen:
Another possibility is joining The Eclipse Foundation.

https://eclipse.org/projects/handbook/#starting

mglukhovsky:
This is something we're exploring right now. Advice and introductions are welcome: mglukhovsky@gmail.com

nirgn975:
I think [Linux Foundation](https://www.linuxfoundation.org/projects) should be an option too.

lenovouser:
Linux Foundation sounds nice, not sure about Apache. They already have a lot of databases ([Cassandra](https://cassandra.apache.org/), [CouchDB](https://couchdb.apache.org/)) - which makes me unsure about how much time will be actually spent on continuing developing RethinkDB. Also loads of their projects who also had a decent amount of users started getting unmaintained for a few years or still are (e.g. [Shindig](https://shindig.apache.org/) or [Archiva](https://archiva.apache.org/)). Maybe even @stripe could support RethinkDB as they [contribute a lot](https://stripe.com/open-source) to Open Source already.

v3ss0n:
I am willing to contribute 5% of every project i have done to Rethinkdb. Please keep it going. I am a startup founder too . I know there will be hard times. But What we are doing with RethinkDB are big!

dvergeylen:
I am exactly in the same situation of @v3ss0n.

Maybe crowdfunded bug resolutions / feature requests could provide cash for short/mid term?
I personnally would be ready to fund some features I was waiting for or at least contribute to help your developers. 

Thank you for what you have done over these 7 years! You are inspiring.

bcantrill:
[...]

PR #6204, opened 2016-12-22
Server Name Indication (Let's Encrypt / Server only certificate) support for Ruby, Python and Java drivers

atomicules:

### Description

Enable Server Name Indication / Let's Encrypt / Server only certificate support in the Ruby, Python and Java drivers; The Node/Javascript driver already works with server only certificates using the `servername` property.

The changes are minimal and non-breaking to existing usage. These commits can be cleaned up, squashed, etc, but for now providing it as is to provide as much information and context as possible.

#### Ruby

For Ruby you'd specify an empty ssl hash to indicate SNI:

```ruby
    conn = r.connect(:host =>'rethink.domain.com',
        :port => 11111,
        :auth_key => '<auth_key>',
        :ssl => {}
    )
```

#### Python

For Python you'd do the same as for Ruby:

```python
    conn = r.connect(host='rethink.domain.com',
        port=11111,
        auth_key='<auth_key>',
        ssl=>{})
```

#### Java

For Java you have to use the sslContext option and initiate it using nulls (which means it uses the system trust manager):

```java
    Connection conn = null;
    Connection.Builder builder = r.connection().
        hostname(host).
	port(port).
	authKey(authKey);

    final SSLContext sslContext = SSLContext.getInstance(DEFAULT_SSL_PROTOCOL);
    sslContext.init(null, null, null);
    conn = builder.sslContext(sslContext).connect();
```

Also need to look at making corresponding changes to the rethinkdb client itself, i.e. for `rethinkdb dump` and `rethinkdb restore`, but I thought this would be a good place to start.

### DIFF
--- a/drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
+++ b/drivers/java/src/main/java/com/rethinkdb/net/SocketWrapper.java
@@ -3,7 +3,10 @@ package com.rethinkdb.net;
 import com.rethinkdb.gen.exc.ReqlDriverError;
 
 import javax.net.SocketFactory;
+import javax.net.ssl.SNIHostName;
+import javax.net.ssl.SNIServerName;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 import java.io.DataInputStream;
@@ -14,7 +17,9 @@ import java.net.Socket;
 import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.ArrayList;
 import java.util.Optional;
+import java.util.List;
 
 public class SocketWrapper {
     // networking stuff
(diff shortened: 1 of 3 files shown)


AtnNn:
Thanks! This looks like a good feature to have. It might take a while before we can review and merge your code, we're working on #6137.

Eventually your nice examples above will have to be added to the docs and test suite.

gabor-boros:
@adriantodt what do you think about this PR for the Java driver?

srh:
Pointing towards `v2.4.x`.  This is so we can make `v2.4.x` the main branch of the repository.  What I *should* do is close this PR and make a separate issue in each of the driver repositories (because they've been split out from the main repo) but... pointing to v2.4.x works for now and is slightly less work.  I intend to revisit this after v2.4.2 gets released.

atomicules:
I am still around. I seem to have a lot less time available for open-source than I used to, but _may_ be able to find time to re-work this PR if needed (i.e. across different driver repos).

_I'm also wondering if we get some of this supported for free now with updates to underlying languages?_

atomicules:
Closing due to age.

PR #6207, opened 2016-12-23
Make lru_cache smaller, simpler, and better.

### LINKED ISSUE #6202: `match' sometimes returns the wrong result

srh:
Fixes #6202.

### Description
In response to #6203.  Note the performance gain -- it doesn't allocate a list node every time you access it.


srh:
I don't know how this history got polluted.  But whatever.

PR #6218, opened 2017-01-07
Serializer cleanup

srh:

### Description
In making changes to the serializer, I put all the cleanup and janitorial commits up front.  So this is that set of commits.

I guess the commits fall into 5 categories:
- Non sequitur commits, don't really have much to do with the serializer
- Further cleanup as a result of the prior removal of the semantic checking serializer
- Moving stuff around so that changes can be tested or otherwise implemented
- Getting rid of all the silly little metablock typedefs
- Simplying some absolutely crazy way of doing stuff (e.g. metablock_manager_t::head_t)

### DIFF
--- a/src/rdb_protocol/changefeed.cc
+++ b/src/rdb_protocol/changefeed.cc
@@ -35,9 +35,8 @@ struct indexed_datum_t {
     }
     datum_t val;
     optional<std::string> btree_index_key;
-    // This should be true, but older versions of boost don't support `move`
-    // well in optionals.
-    // MOVABLE_BUT_NOT_COPYABLE(indexed_datum_t);
+
+    // TODO: It'd be nice if this was non-copyable.
 };
 
 struct stamped_range_t {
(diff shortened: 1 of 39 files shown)


src/rdb_protocol/changefeed.cc
```diff
@@ -1572,8 +1570,8 @@ class flat_sub_t : public subscription_t {
             queue->add(change_val_t(
                 std::make_pair(shard_uuid, stamp),
                 pkey,
-                old_val,
-                new_val
+                std::move(old_val),
+                std::move(new_val)
```
AtnNn: These two parameters are still being copied.

srh: This is now fixed -- the type is noncopyable now.  (And I think it was actually being moved before?)

AtnNn:
LGTM

AtnNn:
Pushed to`next` (68ea014)

PR #6219, opened 2017-01-09
fix zlib-1.2.8 source url

AtnNn:
The archive was moved, the current URL is now a 404


AtnNn:
Merging without review.

PR #6226, opened 2017-01-10
Fix Windows build error with tuple serialization.

srh:
Avoids complaint about zero size array.


srh:
This needs to be added to the 2.4 milestone, if you want 2.4 to build on Windows...

PR #6234, opened 2017-01-20
Issue#3762 Regex Match All

BadBuddha:
- [ ] I have read and agreed to the RethinkDB Contributor License Agreement http://rethinkdb.com/community/cla/

### Description
<Please fill in the description of your pull request here. Thank you for your contribution!>


marshall007:
@BadBuddha thanks for the PR, this looks good! We'll want to add some polyglot tests for this. See [`test/rql_test/src/match.yaml`](https://github.com/rethinkdb/rethinkdb/blob/c3f1bbe0ef5ef45fd63ef87a0c4ac2c94802d814/test/rql_test/src/match.yaml) as a starting point and let us know if you have any questions.

Once that's done we'll need to implement the new `match_all` term in the other drivers. That should be fairly trivial, but @AtnNn and I can help with that effort if necessary/desired.

@AtnNn or @srh can you review the server changes?

admin/static/js/reql_docs.js
```diff
@@ -2,2420 +2,2433 @@
 // hash: "186fb20ea9911710e910acfab0f4f221d59d7c04"
```
AtnNn: I don't see the documentation of matchAll in that branch. Did you edit this file manually?

src/pprint/js_pprint.cc
```diff
@@ -238,7 +240,7 @@ class js_pretty_printer_t
             stack->push_back(lparen);
             *last_is_dot = false;
             *last_should_r_wrap = false;
-            return make_optional(var.arg(0));
+            return boost::make_optional(var.arg(0));
```
AtnNn: What happened here? This file seems unrelated.

AtnNn:
The code looks good to me. However the commit includes a lot of unrelated changes to `reql_docs.js` and it partly reverts the removal of `boost::optional`.

With that fixed, I think it could be merged. However it would still be missing tests, and needs to be added to the other drivers.

admin/static/js/reql_docs.js
meggamind: I'll checkout reql_docs.js, I think there was issue with atom editor. Sorry should have seen this.

src/pprint/js_pprint.cc
meggamind: I don't remember changing this, will 'checkout' this file too

meggamind:
I'll add tests and add this to the other drivers. I should just search for all the places match is being used as matching all and replace it with match_all ?

AtnNn:
I would add separate tests for `match_all` in https://github.com/rethinkdb/rethinkdb/blob/next/test/rql_test/src/match.yaml

For the python driver it should be added in two places in https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/rethinkdb/ast.py

For ruby and java it should be similar.

srh:
Pointing at v2.4.x in anticipation of making that the new main branch of this repo.

I intend to revisit this after v2.4.2 is released.

gabor-boros:
FTR: protobuf file is missing the new term

PR #6241, opened 2017-02-02
Disable broken regex cache

### LINKED ISSUE #6202: `match' sometimes returns the wrong result

AtnNn:
This fixes #6202 for `2.3.x`. #6203 fixes the issue in `next`.

Disabling the cache entirely is the safer option for a point release.

### DIFF
--- a/src/rdb_protocol/terms/string.cc
+++ b/src/rdb_protocol/terms/string.cc
@@ -38,21 +38,14 @@ private:
         std::string str = args->arg(env, 0)->as_str().to_std();
         std::string re = args->arg(env, 1)->as_str().to_std();
         std::shared_ptr<re2::RE2> regexp;
-        regex_cache_t &cache = env->env->regex_cache();
-        auto search = cache.regexes.find(re);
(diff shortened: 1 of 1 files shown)


VeXocide:
LGTM

srh:
This doesn't look good to me.  How is this "safer"?

srh:
The changes in https://github.com/rethinkdb/rethinkdb/pull/6203/commits/4090d1cbb1dda0f3ff13bbe4fd931bdf62541305 are smaller and very straightforward.

The regex cache exists *for a reason*.  Throwing this performance regression into 2.3.6 is just broken.

AtnNn:
This isn't a regression. The cache was never being hit (other than by for empty strings). #6203 makes the cache work, which is a performance improvement, but exercises new code paths. That is why I called this a safer patch.

If I am wrong, then we should definitely use #6203 instead.

srh:
Oh I see.  OK.  LGTM!

PR #6245, opened 2017-02-06
Fix argument order in pretty-printed query

### LINKED ISSUE #6240: r.db("rethinkdb").table("jobs") return wrong value if the query use between
this problem happen for our new installed rethinkdb server with version 2.2.6~0trusty while it is ok in 2.1

when executing long running query using `between`, we can check the `jobs` table `r.db("rethinkdb").table("jobs")` to see the query but the `lowerkey` and `upperkey` are swapped

AtnNn:
Fixes #6240


marshall007:
LGTM

PR #6247, opened 2017-02-06
Add the packaging metadata to build the rethinkdb snap

elopio:

This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -15,7 +15,7 @@ confinement: strict
 apps:
   rethinkdb:
     command: usr/local/bin/rethinkdb
-    plugs: [network, network-bind]
+    plugs: [home, network, network-bind]
 
 parts:
   rethinkdb:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,7 +14,7 @@ confinement: strict
 
 apps:
   rethinkdb:
-    command: rethinkdb
+    command: usr/local/bin/rethinkdb
     plugs: [network, network-bind]
 
 parts:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,38 @@
+name: rethinkdb
+version: master
+summary: The open-source database for the realtime web
+description: |
+  RethinkDB is the first open-source scalable database built for realtime
+  applications. It exposes a new database access model -- instead of polling
+  for changes, the developer can tell the database to continuously push
+  updated query results to applications in realtime. RethinkDB allows
+  developers to build scalable realtime apps in a fraction of the time with
+  less effort.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  rethinkdb:
+    command: rethinkdb
+    plugs: [network, network-bind]
+
+parts:
+  rethinkdb:
+    source: .
+    plugin: autotools
+    build-packages:
+      - build-essential
+      - protobuf-compiler
+      - python
+      - libprotobuf-dev
+      - libcurl4-openssl-dev
+      - libboost-all-dev
+      - libncurses5-dev
+      - libjemalloc-dev
+      - wget
+      - m4
+      - g++
+      - npm
+      - nodejs-legacy
+    configflags: ['--allow-fetch']


elopio:
I loved how you are handling the transition to the community. Kudos everybody.

I thought that it would be nice to have this project easily installable in Ubuntu, so I tried to package it and it was simple enough.

You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/rethinkdb
    $ cd rethinkdb
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build yourself doesn't come signed from the trusted store. Once you push it to the store, your users will just have to do $ sudo snap install rethinkdb)

This requires more testing, but that's what the edge channel in the Ubuntu store is for. A working revision can be pushed to edge and then we can make a call for testing in the community to get feedback before making a stable release.

If you have any questions or comments, please let me know.

AtnNn:
Thanks!

* Is the intent of this patch that users can build their own snaps, or that RethinkDB be added to the snap repository?
* What is the advantage of keeping a `snapcraft.yaml` inside this repo instead of elsewhere?
* Can the process of building a snap use an existing package instead of rebuilding the entire project?

elopio:
Hey @AtnNn, thanks for reviewing it :)

The idea is to have rethinkdb easily installable from the Ubuntu store. Of course, if somebody needs to build the snap, this will make it easy too.

With the yaml in the main repo, you can easily configure your CI system [1] to push a new revision to the edge channel on the store every time that there's a change in master. This is really nice for early feedback from your community, before you make a stable release. We are using everything we learned from maintaining debs to make snaps as lightweight as possible, so you get some niceties and the only thing that you have to add is one yaml. But this is also flexible enough to let you keep the yaml in a different repo or branch, and you could use something like the daily travis cron jobs to make continuous delivery of new versions.

And it can use an existing package. For example, you could just dump the rethinkdb deb into a snap, or any other binary that you build without snapcraft. However, building the snap from the source code has a few advantages. You could use the launchpad ci-integration, and get a snap compiled from all the architectures supported, without changing anything. This is also in [1].

[1] https://snapcraft.io/docs/build-snaps/ci-integration

elopio:
@AtnNn ping. Would you like to meet to discuss more about snaps and how to integrate them in rethinkdb's release process? I would be really useful for us to learn about your process and get your feedback.

AtnNn:
@elopio At the moment, adding a snapcraft package is a low priority for us. I tried playing around with this PR, but snapcraft is not packaged for [my distro](http://nixos.org).

Our packaging process currently looks [like this](https://github.com/rethinkdb/rethinkdb/blob/next/mk/packaging.mk) and [like that](https://github.com/AtnNn/rethinkdb-nix/blob/master/lib/package.nix)

elopio:
@AtnNn you can use an ubuntu lxd in nixos to build it and try it.
We are working on adding more distros, but there's still some way until we reach nixos: https://github.com/snapcore/snapd/wiki/Distributions

elopio:
Hello @AtnNn, I have some news that might be interesting to you. We have a new service that's in beta right now: https://build.snapcraft.io/

It will take care of building the snap for you, and to continuously deliver it to your users in all the supported operating systems. Not yet in nixos, I'm sorry, but this is an easy way to get rethinkdb to all the millions of users out there using ubuntu trusty, xenial, yakkety and zesty, and all the ones that have set up snapd in the other distros.

As I said, it's in beta. So if you give it a try I would be very interested to hear your comments about the service.

pura vida

elopio:
Ping?

gabor-boros:
Hello @elopio,

First of all, thanks again for this contribution! 🙏 

We check the whole backlog of this repo (which is extremely lot). I'll get back to this PR as soon as I can 😇

srh:
Changing to target v2.4.x in anticipation of making that the main branch of this repo.  Unlike other PR's so far, I think this should be considered *before* v2.4.2 gets released.  I.e. sometime this weekend.

srh:
I have rebased and merged this PR into v2.4.x as of 084e782bbbec42a414c54fb2d4916370c77088b1 .

One change I made was to update the list of build-packages to match those of the newest Ubuntu releases and upcoming Jammy release.  This creates the issue that the file won't work with 20.04, but will (presumably) work with 22.04.

I also added libssl-dev to the build-packages list.

I have not tested this at all, and there's some chance it might bitrot, but it's in.

PR #6257, opened 2017-02-12
Make set_file_size_at_least take the real extent size.

srh:
Instead of using DEFAULT_EXTENT_SIZE.  Note: All releases of RethinkDB
since 1.13 have had an extent size of 2.0 MB, with no option to
reconfigure.  (This is just a small code cleanup, don't read too much
into it.)


AtnNn:
- [x] LGTM

PR #6263, opened 2017-02-15
Use cached dependencies in 2.3.x

AtnNn:
2.4.x and next store fetched tarballs in `external/.cache`. The CI server relies on this to avoid re-fetching dependencies and to isolate builds from the network.

This patch adds read-only support for the cache in 2.3.x


AtnNn:
Merged without review. It's small and I need it for packaging 2.3.6

PR #6264, opened 2017-02-16
Improve dirty check in gen-version

AtnNn:
The current check uses `update-index`, which doesn't work if `.git` is read-only


AtnNn:
Merged without review. It's small and I need it for packaging 2.3.6

PR #6282, opened 2017-03-02
jemalloc configure fails with missing --host option during cross compile

cthayer:

### Description
This change adds a space between the --libdir and --host arguments and removes the surrounding quotes so that the jemalloc configure script will process both options correctly.

Prior to this change, when cross compiling the jemalloc package, there was no space between the --libdir and --host arguments to the configure script and they were surrounded in quotes (causing the jemalloc configure script to fail with missing --host option).


AtnNn:
LGTM. Thanks!

Issue #5112, opened 2015-11-18
Bump the version of v8


AtnNn:
We currently use v8 3.30.33.16, but I wasn't able to build it easily on windows.

Newer versions of v8 fix the issues I had and include a lot of bugfixes and performance improvements.

Unfortunately, the API is not backwards-compatible.

analytik:
Is there any specific version of v8 you're aiming for?

Node.js 6 will be built on v8 5.0, which has an extensive ES6 support. Since Node.js 6 will be a long term support release, I think it would be great if RethinkDB had the same or newer v8 (probably not for 2.3, but some time during 2016 would be nice). 

This is just a nice to have feature, I guess mostly to keep syntax the same between newer ES6 Node.js projects and the Administration interface (if that's relevant at all, and the code entered there is processed by the v8 and not some driver translating the code before it gets to the server).

danielmewes:
@analytik The version we pick isn't all that important, since it only affects the `r.js` command. The Data Explorer relies on the browser to parse the JavaScript, so it will support whatever language features are available in your browser.

AtnNn:
V8 also fails to build on Ubuntu Yakkety (16.10) which has GCC 6.1 (see #5757).

The current stable version of V8 is 5.4.500.41 (ahttps://omahaproxy.appspot.com/)

We're still using V8 3.30.33.16 on Linux. On Windows we use 4.7.80.23, but the build process is more fragile, and it takes a lot longer.

AtnNn:
One of the obstacles of upgrading V8 is making their new build system compatible with ours.

Simply fetching the source of V8 requires over 500GiB of space, a lot of which is taken up by custom, platform specific versions of clang, python and other tools.

Our build and packaging system assumes it can fetch or create platform-independent tarballs of fetched dependencies, and that the subsequent build step can be performed offline.

srh:
> 500GiB

I assume this is supposed to say 500 MB.

srh:
See #6401 for an alternative.

gabor-boros:
@srh  I'm closing this as QuickJS replaced V8

PR #6295, opened 2017-03-10
Add Linux on z support (s390x)

mundaym:

(I think the CLA has been signed on my and Rishi's (@rposts here) behalf by an IBM representative. Let me know if this isn't your understanding).

### Description
This adds support for Linux on z (s390x). A couple of open questions:

 * Is it worth splitting up the context switching change (bcd07c8) into two commits? I've modified some of the comments to (hopefully) make what is going on a bit clearer as ports are added. I can break that out separately if you want to review it independently.
* Updating V8 to the lastest version for all non-Windows OSs would remove the need to pull in a s390x-specific version of it in 9f83100. We gave it a go but are seeing compilation errors. Are you planning on updating V8 at some point?

Thanks!

### DIFF
--- a/configure
+++ b/configure
@@ -105,6 +105,8 @@ configure_unixlike () {
         arm*)
             var_append LDFLAGS -ldl
             final_warning="ARM support is still experimental" ;;
+        s390x)
+            final_warning="s390x support is still experimental" ;;
         *)
             error "unsupported architecture: $MACHINE"
     esac
--- a/mk/support/pkg/node.sh
+++ b/mk/support/pkg/node.sh
@@ -1,5 +1,6 @@
+major=6
+minor=10.0
+version=$major.$minor
 
-version=0.12.2
-
-src_url=http://nodejs.org/dist/v$version/node-v$version.tar.gz
-src_url_sha1=a969f17a0a6c9238584f8946d96e8d39be8eb957
+src_url=http://nodejs.org/dist/latest-v$major.x/node-v$version.tar.gz
+src_url_sha1=3bb2629ed623f38b8c3011cf422333862d3653a3
--- a/mk/support/pkg/v8.sh
+++ b/mk/support/pkg/v8.sh
@@ -1,14 +1,6 @@
-# See http://omahaproxy.appspot.com/ for the current stable/beta/dev versions of v8
-if [[ "$OS" != Windows ]]; then
-
+if [[ "$OS" = Windows ]]; then
     # V8 3.30 doesn't play well with Visual Studio 2015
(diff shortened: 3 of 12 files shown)


AtnNn:
* I can confirm that the CLA has been signed.
* The context switching commit looks small enough to me, splitting it doesn't seem necessary.
* Yes, we do want to move to a newer version of V8 on all platforms. (See #5112. It would require a laborious rearchitecturing or gutting of our build system and a lot of multi-platform testing.)

Thanks for sharing your changes! They're very succinct and nicely organised.

As far as I can tell, these changes don't affect the build on other platforms and could safely be merged.

Do you have any recommendations for keeping the code compatible with s390x, perhaps something that we could add to the [style guide](https://github.com/rethinkdb/rethinkdb/blob/next/STYLE.md)?

mundaym:
> As far as I can tell, these changes don't affect the build on other platforms and could safely be merged.

Great! Thanks for your review.

> Do you have any recommendations for keeping the code compatible with s390x, perhaps something that we could add to the style guide?

Good question. The only recommendation I have is that future code avoid using reinterpret casts to convert from numbers (`int`, `float` etc.) to raw bytes (`void*` and `char*`) and vice versa. This style of code can cause not just endianness problems but also alignment problems on other architectures like ARM and MIPS.

Maybe something like this could be added:

<hr>

17. Do not use reinterpret casts to convert values to raw bytes and vice versa. Prefer portable code.

> E.g. don't do this:
>
>     int32_t v = ...;
>     ssize_t n = send(fd, reinterpret_cast<char*>(&v), sizeof(v), 0);
>
> Instead, perform the conversion explicitly (perhaps in a helper function):
>
>     int32_t v = ...;
>     char buf[4] = { v, v>>8, v>>16, v>>24 };
>     ssize_t n = send(fd, &buf[0], sizeof(buf), 0);
>
> Reason:  using reinterpret casts in this way can cause incompatibilities and performance issues on some CPU architectures.

<hr>

Unfortunately gcc and clang aren't always very good at merging little (or big) endian loads and stores. (For example: https://bugs.llvm.org//show_bug.cgi?id=31162). In some circumstances it may therefore be worth using the `#ifdef`s to choose between implementations (e.g. `reinterpret_cast` on x86, `__builtin_bswap` on s390x and so on). We did it this way to avoid introducing performance regressions. New code, however, can probably get away with the portable implementation when writing out headers to streams without any noticeable performance impact which is why I suggest this wording for the style guide.

AtnNn:
Merged into next and v2.4.x. These changes will be in 2.4

PR #6311, opened 2017-03-24
Fix Node.js fetch address.

mundaym:
The version of Node.js in the 'latest' directory changes over time.
Switch to using the version-specific directory.

### Description
A new version of Node.js has been released and it has broken the link we were using. This change should fix that (and stop it happening again).

### DIFF
--- a/mk/support/pkg/node.sh
+++ b/mk/support/pkg/node.sh
@@ -1,5 +1,4 @@
 version=6.10.0
-major=${version%%.*}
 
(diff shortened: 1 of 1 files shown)


marshall007:
LGTM

/cc @AtnNn

srh:
Should be merged to v2.4.x.

AtnNn:
Thanks, merging and chery-picking into v2.4.x as 75e8fa2

Issue #6315, opened 2017-03-29
Can we use V8 v4.7 on Linux.


asowani:
Hi,

File mk/support/pkg/v8.sh, lines 6-7 say that "4.7 has no source distribution, making it harder to build on Linux". I tried building RethinkDB with V8 v4,7 on Ubuntu 16.04/ppc64le and it seems to be working fine - all unit tests passed so far. Further testing is in progress - haven't hit any serious issue yet. I have introduced appropriate conditions in source code to pick V8 v4.7 if building on ppc64le.

So, I am wondering if it is okay to use V8 v4.7 on Linux if it works, or still there is a hard requirement to use 3.30 only? Any comments/advise about it?

Thanks,
Atul.

srh:
That's nice to hear, but how did you do it?  Is it done in a way that can be conveniently integrated with RethinkDB's build system?

asowani:
@srh Thanks! I have created PR for this today (https://github.com/rethinkdb/rethinkdb/pull/6317).

gabor-boros:
@srh v8 is dropped in favor of quickjs, should we close this?

PR #6317, opened 2017-04-03
Support for Power8/LE platform (ppc64le).

asowani:

Yes.

### Description
These code changes are done for:
1. Support for ppc64le.
2. Use of v8 v4.7 (https://github.com/rethinkdb/rethinkdb/issues/6315)

Thanks!

### DIFF
--- a/src/arch/runtime/context_switching.cc
+++ b/src/arch/runtime/context_switching.cc
@@ -260,6 +260,8 @@ artificial_stack_t::artificial_stack_t(void (*initial_fun)(void), size_t _stack_
 #elif defined(__arm__)
     // This slot is used to store r12.
     const size_t min_frame = 1;
+#elif defined(__powerpc64__)
+    const size_t min_frame = 1;
 #endif
     // Zero the caller stack frame. Prevents Valgrind complaining about uninitialized
     // value errors when throwing an uncaught exception.
@@ -270,6 +272,9 @@ artificial_stack_t::artificial_stack_t(void (*initial_fun)(void), size_t _stack_
 
     // (4) write the return address to the stack.
     sp--;
+#if defined (__powerpc64__)
+    sp -= 22;
+#endif
     *sp = reinterpret_cast<uintptr_t>(initial_fun);
 
     // (5) allocate space for callee-save registers. These will be popped off the
@@ -286,6 +291,8 @@ artificial_stack_t::artificial_stack_t(void (*initial_fun)(void), size_t _stack_
     sp -= 8; // r4-r11.
 #elif defined(__s390x__)
     sp -= 16; // r6-r13 and f8-f15.
+#elif defined (__powerpc64__)
+    sp += 24;
 #else
 #error "Unsupported architecture."
 #endif
@@ -444,7 +451,7 @@ void context_switch(artificial_stack_context_ref_t *current_context_out, artific
 }
 
 asm(
-#if defined(__i386__) || defined(__x86_64__) || defined(__arm__) || defined (__s390x__)
+#if defined(__i386__) || defined(__x86_64__) || defined(__arm__) || defined (__s390x__) || defined (__powerpc64__)
 // We keep architecture-specific code interleaved in order to enforce commonality.
 #if defined(__x86_64__)
 #if defined(__LP64__) || defined(__LLP64__)
@@ -465,6 +472,8 @@ asm(
     /* `current_pointer_out` is in `r0`. `dest_pointer` is in `r1` */
 #elif defined(__s390x_)
     /* `current_pointer_out` is in `%r2`. `dest_pointer` is in `%r3`. */
+#elif defined(__powerpc64__)
+    /* `current_pointer_out` is in `r3`. `dest_pointer` is in `r4` */
 #endif
 
     // Save preserved registers.
@@ -500,6 +509,38 @@ asm(
     "std %f13, 40(%r15)\n"
     "std %f14, 48(%r15)\n"
     "std %f15, 56(%r15)\n"
+#elif defined(__powerpc64__)
+    "std  1, 0(3)\n"
+    "addi 1,1,-200\n"
+    "mfcr     0\n"
+    "std      0, 0(1)\n"
+    "mflr     0\n"
+    "std      0, 8(1)\n"
+    "mfctr    0\n"
+    "std      0, 16(1)\n"
+    "mfxer    0\n"
+    "std      0, 24(1)\n"
+    "std       0, 32+(8*0)(1)\n"
+    "std       2, 32+(8*1)(1)\n"
+    "std       3, 32+(8*2)(1)\n"
+    "std      14, 32+(8*3)(1)\n"
+    "std      15, 32+(8*4)(1)\n"
+    "std      16, 32+(8*5)(1)\n"
+    "std      17, 32+(8*6)(1)\n"
+    "std      18, 32+(8*7)(1)\n"
+    "std      19, 32+(8*8)(1)\n"
+    "std      20, 32+(8*9)(1)\n"
+    "std      21, 32+(8*10)(1)\n"
+    "std      22, 32+(8*11)(1)\n"
+    "std      23, 32+(8*12)(1)\n"
+    "std      24, 32+(8*13)(1)\n"
+    "std      25, 32+(8*14)(1)\n"
+    "std      26, 32+(8*15)(1)\n"
+    "std      27, 32+(8*16)(1)\n"
+    "std      28, 32+(8*17)(1)\n"
+    "std      29, 32+(8*18)(1)\n"
+    "std      30, 32+(8*19)(1)\n"
+    "std      31, 32+(8*20)(1)\n"
 #endif
 
     /* Save old stack pointer. */
@@ -533,7 +574,9 @@ asm(
     "mov r13, r1\n"
 #elif defined(__s390x__)
     /* On s390x, the second argument is in r3 */
-    "lgr %r15, %r3\n"
+    "lgr %r15, %r3\n"a
+#elif defined(__powerpc64__)
+    "mr 1, 4\n"
 #endif
 
 #if defined(__i386__)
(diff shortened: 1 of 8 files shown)


srh:
In my opinion the changes look great, at least the code diff has a good smell to it.  (Some aspects, like build system decisions, I'm not qualified to judge.)

Why is the coroutine stack size increased from 128K to 256K on this platform?  I'd like to add a comment telling people why.

srh:
Another weird thing:  The change defining V8_NEEDS_BUFFER_ALLOCATOR to 1 in the middle of js_job.cc.  That should presumably be done in some build config setting.

srh:
Another potential bug is that in `src/build.mk` the change only affects things if the compiler is GCC.  What if it's another compiler?

mundaym:
Hi,

In the context_switching.cc change it looks like the new code saves registers below the stack pointer. All the other architectures decrement the stack pointer and then save the registers. Here are some diagrams of what the stacks look like after the `artificial_stack_t` constructor is called:

Existing architectures:
```
|   | <- unallocated memory (and maybe some allocated, but unaligned memory)
+---+
|   | <- fake caller stack frame (of size min_frame)
+---+
|   |
|   | <- register save area (saved stack pointer points to top of register save area)
|   |
+---+ <- stack pointer
```

Power:
```
|   | <- unallocated memory (and maybe some allocated, but unaligned memory)
+---+ <- stack pointer*
|   |
|   | <- register save area (saved stack pointer points to top of register save area)
|   |
+---+

 * sp += 24 removes the fake caller stack frame
```

Is there a reason power needs to behave differently? If so then it would be really good to add some comments.

Also I think there is a problem with removing the fake caller stack frame. If I understand the Power ELF V2 ABI correctly, then the `min_frame` constant should be 4 so that the `initial_fun` function can safely write LR to the LR save slot. As it stands I think `initial_fun` will in fact write LR into unallocated memory (beyond the top of the stack).
```

**Comment by asowani** [comment]
@srh @mundaym I appreciate your assessment and comments very much. I am working on the comments and suggestions. I will get back with appropriate code changes.

**Comment by asowani** [comment]
Implemented V8_NEEDS_BUFFER_ALLOCATOR related change. Pondering over the context switching discussion.

**Comment by srh** [comment]
The change there adds general linking flags.  I'd expect other compilers to need the same flags.

I don't expect you to test other compilers.  If people see flags like that in a compiler-specific branch, they're likely to think the flags are GCC-only for some specific reason.  And if Clang was used, you'd expect it to take the same flags.

Another bug in build.mk is that `#if ! defined(__powerpc64__)` isn't going to do anything there.  It's a comment.

A better way might be to make `pkg_link-flags` in v8.sh output -licui18n, -licuuc, and -ldl.  I believe the lines echoed out there get turned into linker parameters without any manipulation.  This keeps the v8 library logic in the same place.

**Comment by asowani** [comment]
I have made change for V8_NEEDS_BUFFER_ALLOCATOR and the changes suggested in v8.sh, build.mk files.

**Comment by asowani** [comment]
Marked the review comments which are implemented.

**Comment by asowani** [comment]
@srh @mundaym All the review comments have been incorporated now. Could you please check and let me know if I missed out something? Also please let me know if you have additional queries, comments. Thanks!

File: src/config/args.hpp [hunk 1]
```diff
@@ -145,6 +145,8 @@
 #define LBA_RECONSTRUCTION_BATCH_SIZE             1024
 
 #if defined (__powerpc64__)
+// POWER ABI requires more number of registers to be saved
+// in stack during context switch.
```
mundaym: This doesn't make sense to me. The context switch requires 168 bytes of stack space. I don't see why that would mean the stack size needs to be doubled. What error do you see? Does is still occur with the revised context switching code?

src/arch/runtime/context_switching.cc
```diff
@@ -292,7 +289,7 @@ artificial_stack_t::artificial_stack_t(void (*initial_fun)(void), size_t _stack_
 #elif defined(__s390x__)
     sp -= 16; // r6-r13 and f8-f15.
 #elif defined (__powerpc64__)
-    sp += 24;
+    sp -= 20; // r14-r31, toc, cr.
```
mundaym: Strictly speaking you also need to save f14-f31 and v20-v31 to meet the ABI requirements. The compiler could choose to save something in one of these registers and a coroutine switch might end up clobbering it. It's the sort of thing that is very hard to debug. At the very least there should be a comment here explaining the danger.

AtnNn:
Awesome work.

Ideally, if we set `use_system_icu=1` when building v8, the configure script should test for ICU.

Otherwise the v8-related changes look okay to me. They aren't uglier than what we already have. Someday we'll get around to do #5112

src/arch/runtime/context_switching.cc
RajalakshmiS: @mundaym Considering the factors performance and being light weight, I have not saved floating point and vector registers.There are no failures seen when non volatile floating point registers
are not saved and restored during context switch . I have tested on
ubuntu/RHEL powerpc64le systems. Also we can hit any bugs.

src/config/args.hpp
RajalakshmiS: @mundaym There are alloca() calls in glibc called by getifaddrs() which tries to
allocate 64KB memory in stack frame for powerpc. If I dont increase the stack
size, there are failures in test cases due to stack overflow checks in rethinkdb.
Hence increasing it for powerpc.

AtnNn: Can this explanation be added as a comment?

src/arch/runtime/context_switching.cc
AtnNn: There's a comment above that indicates that a context switch "doesn't swap the floating-point registers"

mundaym: Yeah, it's safe to do on x86 because none of the SSE/floating point registers are callee-save (and so aren't 'live' on entry to the swapcontext function). On POWER and s390x however the ABIs have callee-save floating point registers and the compiler assumes they will be restored after a function call.

In practice the compiler probably doesn't use the callee-saved floating point registers very often in code like this, so not saving/restoring them might not cause any issues. But it is a bit fragile IMO. Worth a comment anyway.

src/config/args.hpp
RajalakshmiS: @AtnNn Added comment.

src/arch/runtime/context_switching.cc
RajalakshmiS: @mundaym Added code comment.

src/config/args.hpp
RajalakshmiS: Added code comments.

asowani:
@AtnNn Thanks for adding this PR to 2.5 milestones. Any estimate on timeframe when 2.5 can be expected? Also, any plans to create a new branch for 2.5?
Thanks!

AtnNn:
@asowani The branch for 2.5 is `next`. We don't have a timeframe for 2.5. If I had to guess, I would say 4 to 12 months.

asowani:
@AtnNn Thanks for letting me know about the estimated time frame for 2.5 release. I don't know much about how much effort it generally takes for new release preparation (may be creating a new branch?), so please pardon if I sound too eager. But is there any way we can have 2.5 release a bit earlier? If there is any help required, I am willing to assist. Thanks!

AtnNn:
We're trying to get 2.4 ready right now. I'll try to get this patch into 2.4 if it doesn't delay the release.

asowani:
@AtnNn - is v2.4.x the correct branch to check for merged patches for upcoming 2.4 release?

AtnNn:
@asowani yes.

asowani:
@AtnNn I have setup a Jenkins job to test v2.4.x branch whenever ppc64le code gets merged. At the moment I have disabled the job as I don't see those changes in the branch, but it can start the moment the changes get merged in the branch. I guess I will get an automated mail from github when this PR is merged, but just in case, I would highly appreciate if you could intimate here once the code is merged. Thanks!

clnperez:
@AtnNn -- what's your feeling on whether this will make it in to the 2.4 release or not? It still has the milestone, which gives me hope that it will. 😆

mchasal:
Just following up on the previous comment. Any outlook on whether this will be merging into 2.4 at this point? Thanks!

srh:
I'd say it's likely to make it into 2.4.

srh:
I've rebased and cleaned up the changes in this PR and put them in branches sam/24x/powerpc64 (based off v2.4.x) and sam/powerpc64 (based off next) in my repo at <https://github.com/srh/rethinkdb/>.

There are two things I'd still like to understand before putting this into 2.4.

First, what is the point of setting V8_NEEDS_BUFFER_ALLOCATOR?

Second, why does getifaddrs use alloca on powerpc64, but not on amd64?  Is there an implementation I could look at in order to see this?

I don't want to throw in these mysterious configuration options without putting an explanation in the source code.

srh:
There's a third thing.  Why is v8 4.7 used for powerpc64?  Is it necessary?

srh:
I've updated the branches referenced above to compute the coroutine stack size based on the page size value.

So there are now still two things I'd like to understand:

- why V8_BUFFER_ALLOCATOR?

- why v8 4.7?

asowani:
@srh The selection of v8 4.7.x was already done in the script but was limited to Windows. Older versions of v8 are not ported to ppc64le and v8 4.7 has been ported to and tested on ppc64le, and is found to be working good on it. V8_NEEDS_BUFFER_ALLOCATOR is required at build time to have 4.7 version of v8.

RajalakshmiS:
@srh 
>Second, why does getifaddrs use alloca on powerpc64, but not on amd64? Is there an implementation I >could look at in order to see this?
    This really depends on the pagezise configured on that arch and for powerpc64 it is passed as libc_use_alloca(65536) and it uses alloca().

asowani:
@srh Any updates about code merge? Please let me know in case there are further queries. Thanks!

srh:
But what is it about ppc64le that makes v8 4.7 need V8_BUFFER_ALLOCATOR?

srh:
If I check out v8 4.7.80.23 and grep for V8_NEEDS_BUFFER_ALLOCATOR, I don't get any matches, and if I search on Google I don't get any results outside of this issue.

asowani:
@srh I think this is with reference to commit https://github.com/rethinkdb/rethinkdb/pull/6317/commits/e3de540eaf1a817c9bcdb89ab9de0f6ded2c9660

src/extproc/js_job.cc file has following code (around line 34):

#ifdef V8_NEEDS_BUFFER_ALLOCATOR
class array_buffer_allocator_t : public v8::ArrayBuffer::Allocator {
public: 
...

I think the change is required to take care of this.

srh:
Oh, I'm a retard. Will merge this when I get back to a PC.

srh:
In next as of c300a18b7027a70f9b50fcb44d017a3483514b46 and v2.4.x as of 4a71acffcd19ff4ae32a7f98cae9549aa6c6037f.

asowani:
@srh Thanks very much for merging the changes!

PR #6322, opened 2017-04-04
Fixes #6296 InvalidStateError after cancelling a RethinkDB operation.

### LINKED ISSUE #6296: Asyncio Driver Shouldn't Wrap CancelledError
The `connect()` method in the Asyncio driver has an exception handling clause like this:

    try:
        self._parent.handshake.reset()
        response = None
        with translate_timeout_errors():
            while True:
                request = self._parent.handshake.next_message(response)
                if request is None:
                    break
                # This may happen in the `V1_0` protocol where we send two requests as
                # an optimization, then need to read each separately
                if request is not "":
                    self._streamwriter.write(request)

                response = yield from asyncio.wait_for(
                    _read_until(self._streamreader, b'\0'),
                    timeout, loop=self._io_loop,
                )
                response = response[:-1]
    except ReqlAuthError:
        yield from self.close()
        raise
    except ReqlTimeoutError as err:
        yield from self.close()
        raise ReqlDriverError(
            'Connection interrupted during handshake with %s:%s. Error: %s' %
                (self._parent.host, self._parent.port, str(err)))
    except Exception as err:
        yield from self.close()
        raise ReqlDriverError('Could not connect to %s:%s. Error: %s' %
                              (self._parent.host, self._parent.port, str(err)))

The last `except` block is problematic, because if a task is canceled while it's suspended at `response = yield from asyncio.wait_for(...)`, that `asyncio.CancelledError` will get wrapped up in a `ReqlDriverError` with an empty error message. This is hard to debug (because of the empty error message), and it also breaks any cancel-friendly code. For example:

    try:
        await something_that_is_safe_to_cancel()
    except asyncio.CancelledError:
        pass
  
This says that if `something_that_is_safe_to_cancel()` is cancelled, we should just ignore it and move on. But if `something_that_is_safe_to_cancel()` calls RethinkDB's `connect()` method, then the raised exception is actually `ReqlDriverError`, not `CancelledError`. After a lengthy debugging session, I finally figured this out and wrote this workaround code:

    try:
        await something_that_is_safe_to_cancel()
    except r.ReqlDriverError as rde:
        if isinstance(rde.__context__, asyncio.CancelledError):
            pass
        else:
            raise

I consider this a bug in the RethinkDB driver, and my suggested fix is to add a penultimate `except` block:

    ...
    except asyncio.CancelledError:
      yield from self.close()
      raise
    except Exception as err:
        yield from self.close()
        raise ReqlDriverError('Could not connect to %s:%s. Error: %s' %
                              (self._parent.host, self._parent.port, str(err)))

I can submit a patchfile or PR if you agree. Thanks!

mehaase:

### Description

Fixes #6296.

If a Python coroutine is cancelled while it is suspended inside a RethinkDB query, the query
future is cancelled. When the connection is later closed, the `close()` method tries to set the
future result, which throws `InvalidStateError` because the future was already cancelled. This
commit adds a guard condition to avoid this.

### DIFF
--- a/drivers/python/rethinkdb/asyncio_net/net_asyncio.py
+++ b/drivers/python/rethinkdb/asyncio_net/net_asyncio.py
@@ -230,7 +230,8 @@ class ConnectionInstance(object):
             cursor._error(err_message)
 
         for query, future in iter(self._user_queries.values()):
-            future.set_exception(ReqlDriverError(err_message))
+            if not future.done():
+                future.set_exception(ReqlDriverError(err_message))
 
         self._user_queries = { }
         self._cursor_cache = { }


mehaase:
Sorry, tagged

marshall007:
LGTM

/cc @AtnNn

AtnNn:
LGTM too.

Thanks @mehaase! If you agree with our CLA could you please fill-in the form at the bottom of the agreement?

mehaase:
@AtnNn Done, thanks!

srh:
Rebased onto next, cherry-picked into v2.4.x, v2.3.x.  Thank you!  Sorry about the delay.

Now we have contributors named Marc Hesse and Mark Haase.

Issue #6333, opened 2017-04-18
Guarantee failed: [entry->state == entry_t::state_t::STREAM]


analytik:
Hi,

Got this error on one of the nodes in production, I haven't seen it before. It's hard to say if there was anything unusual going on, but besides the usual production load, we've been doing two things:
```
r.db("foo")
 .table("bar")
 .between(r.iso8601("2016-07-15T00:00Z"), r.iso8601("2016-08-01T00:00Z"), {"index": "created"})
 .filter(function (var127) {             return r.not(r.row.hasFields("updated"));         })
 .update(function (var128) {             return r.object("updated", r.row("created"));         })
```
Which was updating hundreds of thousands of docs in batches, and

```
r.db("ormFoo")
 .table("ormBar")
 .update({
          // 20 or so fields omitted
          "id": "a87aef60-3533-415a-9769-5b16c5959380",
          "created": r.iso8601("2017-04-18T12:46:40.462Z"),
          "updated": r.iso8601("2017-04-18T12:46:40.660Z"))
```
which, of course, is a horrible thing as it tries to update ~20k documents one by one, and fails on all except for the one correct ID.

Here's the crash trace:
```
Version: rethinkdb 2.3.5~0jessie (GCC 4.9.2)
error: Error in src/rdb_protocol/query_cache.cc at line 325:
error: Guarantee failed: [entry->state == entry_t::state_t::STREAM]
error: Backtrace:
error: Tue Apr 18 12:31:15 2017

       1 [0xae7500]: backtrace_t::backtrace_t() at ??:?
       2 [0xae7879]: format_backtrace(bool) at ??:?
       3 [0xd9f6c3]: report_fatal_error(char const*, int, char const*, ...) at ??:?
       4 [0x72ffac]: ql::query_cache_t::ref_t::serve(ql::env_t*, ql::response_t*) at ??:?
       5 [0x7323f2]: ql::query_cache_t::ref_t::fill_response(ql::response_t*) at ??:?
       6 [0x805e63]: rdb_query_server_t::run_query(ql::query_params_t*, ql::response_t*, signal_t*) at ??:?
[...]

PR #6343, opened 2017-04-29
Fix potential race condition in query cache

AtnNn:
`terminate_internal` modifies the `entry->state` asynchronously, without holding a lock on the entry.

This might be what caused the guarantee failure in #6333

- [x] I have read and agreed to the CLA


srh:
In principle this looks good to me.  But note that in `~ref_t` there is a `guarantee(entry->state != entry_t::state_t::START)` there.  That seems like it might now fail, no?

srh:
Fixed by #6564 (which is built off these changes).

PR #6354, opened 2017-05-07
Fix metadata migration for non-v2_3->v2_4.

srh:
Two bugs.

1. The v2_1->v2_3 auth metadata migration was getting passed over.
All that "serialization_version" means in the function
migrate_metadata_v2_1_to_v2_3 is what version the structs were last
serialized at -- what cluster_version_t W parameter was used.  Auth
metadata _still_ needs to get migrated.

2. Because the v2_2 case in metadata_file_t::metadata_file_t did not
set metadata_version to the value cluster_version_t::LATEST_DISK, the
subsequent call to migrate_metadata_v2_3_to_v2_4 would hit
unreachable() in migrate_metadata_v2_3_to_v2_4.

This should also be cherry-picked into the v2.4.x branch and make it into v2.4.


AtnNn:
I don't know this code too well, but these changes LGTM.

Issue #6358, opened 2017-05-10
Build failure, cannot find v8.h


srh:
I get
```
src/extproc/js_job.cc:4:16: fatal error: v8.h: No such file or directory
```
This happens after the recent web UI merge and #6357 fix.

I'm getting it on Xubuntu 17.04 now, but I also saw it on Jessie after tweaking paths before the recent fix to #6357.

srh:
Also later I'm getting

```
./build.sh: 25: ./build.sh: [[: not found
./build.sh: 43: ./build.sh: [[: not found
./build.sh: 30: ./build.sh: browserify: not found
./build.sh: 48: ./build.sh: lessc: not found

**Comment by AtnNn** [comment]
I've seen that `v8.h` error before but I don't know what the cause is. A possible workaround is to run `make install-include-v8`.

Regarding `[[: not found`, I think `admin/build.sh` is missing `#!/usr/bin/env bash`. I'll test and fix that right now

**Comment by srh** [comment]
Also seeing stuff like

```
    [1/18] BUILD coffee-script_1.10.0
touch: cannot touch 'build/external/bluebird_2.9.32/node_modules/packed-bluebird/node_modules/bluebird': No such file or directory
drivers/javascript/build.mk:62: recipe for target 'build/external/bluebird_2.9.3
make[1]: *** [build/external/bluebird_2.9.32/node_modules/packed-bluebird/node_modules/bluebird] Error 1
```

srh:
Fixed by reverting changes in #6363.

PR #6361, opened 2017-05-12
Add missing shebang to admin/build.sh

AtnNn:
This should fix the `[[: not found` issue (https://github.com/rethinkdb/rethinkdb/issues/6358#issuecomment-300606330)


srh:
I have merged this. But the build is still broken for other reasons.

PR #6392, opened 2017-06-13
Make soft durability writes use a flush interval.

srh:

### Description
This changes the cache's flush strategy for soft durability writes to be to accumulate changes, and then flush the changes every `t` seconds.  It has the effect of greatly reducing the number of write operations.

The feature is described in the notes for https://github.com/srh/rethinkdb/releases/tag/v2.3.5-srh-extra except that the configuration option is moved to a top-level `flush_interval` table config.  Also, the default interval is 1.0 seconds.

This addresses some suggestions in #1771.  It's easy to imagine people finding something to complain about here, so I'm going to leave this up for a while.

### DIFF
--- a/src/arch/timer.cc
+++ b/src/arch/timer.cc
@@ -58,7 +58,7 @@ void timer_handler_t::on_oneshot() {
             token_queue.push(token);
         }
 
-        token->callback->on_timer();
+        token->callback->on_timer(ticks_t{real_ticks});
 
         // Delete nonrepeating timer tokens.
         if (token->interval_nanos == 0) {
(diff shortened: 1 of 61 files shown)


danielmewes:
Haven't looked at the code, but from the description of it, this should be awesome to reduce the amount of write amplification! I've always been somewhat unhappy with how bad RethinkDB was at that in some scenarios.

adallow:
This would be very useful! Can it be part of the next release?

srh:
It's released as 2.3.5-extra, link in the OP.  And 2.4 won't have it, but I will release a 2.4-extra.  Then 2.5 will (hopefully) have it.

srh:
I see no complaints -- in next with commit 011bea8a3fa7244f56d4ef1c236f3b9402daf924.

adamierymenko:
FYI -- we finally got around to taking this live. Our writes dropped from ~200 **MEGABYTES** per second to around 400 **KILOBYTES** per second. That's with only two tables configured to have delayed flush, so the gain is not entirely from configuring that. This patch improves I/O write load so enormously that the previous level of write load could almost be considered a bug.

adallow:
Hi any idea when this will be released officially for rethink? (not sure if this is the right place to ask, but we would love this feature!). I assume it comes when what is in next becomes a release, but do you have any indication as to when that may be? And many thanks for the great work on this.

lbguilherme:
Looks like current `next` will become v2.5. As v2.4 has no ETA yet... this is the only way to go: https://github.com/srh/rethinkdb/releases/tag/v2.3.6-srh-extra

srh:
Note that a pretty nasty memory leak has been reported, so you might want to hold off trying it until an update.

adallow:
@srh thanks for the heads up, will keep an eye out for updates. Appreciate your work.

thelinuxlich:
Does this helps with single-write performances, and those who are using it in production, did you step into the memory leak?

adallow:
Any news on this feature, would be cool to have it in the next release.

PR #6394, opened 2017-06-14
Shrink size of gc_entry_t

srh:

### Description
This reduces the memory overhead of gc_entry_t.  This reduces the constant factor of memory overhead per unit of data on disk.


srh:
The most controversial change here is the change of the block size from uint32_t to uint16_t, everywhere.

srh:
In next as of ceb207cf60a7b4794576f0d1b987ac5640aa2add.  Changing block size to uint16_t is not particularly controversial, because we can always change it back later.

Issue #6267, opened 2017-02-17
Proposal: r.write_time() in write hooks.


srh:
This proposal is that
(1) Like in v2.3, `r.now()` in a particular query should never change what value it evaluates to.
(2) `r.write_time()` should evaluate to the "write time" in certain contexts:  In write hooks, and maybe in changefeed maps and filters later.
(3) `r.now()` should be disallowed in write hooks (like with secondary index functions).

----

The reason for (1) is, having `r.now()` change meaning depending on exactly where it is, and having this depend on evaluation order, can be confusing.  Evaluation order is not exactly well-defined in RethinkDB.  This is a particular problem in changefeeds.  In next, `.changes().merge({'time': r.now()})` will compute different values of `r.now()`, because `.merge` takes its argument with call-by-name behavior.  (*That* is fixed by pull request #6191.)

The reason for (2) is, we need some way to refer to the write's timestamp in write hooks.  Another option would be to make write hooks take a fourth function parameter.  The main reason I went with `r.write_time()` is because before the plan was to use `r.now()`.

The reasons for (3) is,
  - first of all, setting a write hook that uses `r.now()` would be weird
  - people might do it by accident, expecting `r.write_time()` behavior, and never get a warning
  - if you really want to use `r.now()`, you can work around it with `.do`.

----

This is implemented in https://github.com/srh/rethinkdb/tree/sam/write_hook  as of commit c9bf5074a06f4eac59a3babb1510a9819140e341.  (Except the Java driver, for which I have no freakin' clue.)  It uses the command name `r.write_timestamp()`.

----

Handy link to the initial write hooks proposal: It has some discussion of this. #5813

marshall007:
I really like this proposal and cheeky agree on all three points. My only suggestion would be that we consider introducing a term for exposing arbitrary metadata related to the query and execution context. This would lend itself well to also exposing the secondary index name and its value associated with the current row, shard the row is being written to/read from, etc without introducing new terms in the future.

Something like `r.context` and/or `r.ctx`?

mlucy:
+1 on keeping `r.now` like v2.3 and disallowing it in write hooks.

I'm torn on `r.write_time`.  I worry a little bit about ending up in a world where there are a bunch of terms with unusual behavior that are only legal in certain contexts.  But on the other hand we already did that with `r.literal` and it's been basically fine; people seem to treat it as syntax rather than a magic term which is sometimes illegal.

[...]

PR #6396, opened 2017-06-15
Write hooks with `{'primary_key': , 'timestamp':}`, also fixes r.now determinism problems

srh:

### Description

This implements #6267, which is a ReQL proposal.  Has it resolved to an agreement?  Who exactly is in charge here?  This PR also fixes deterministic r.now problems addressed by PR #6191, in fact it includes its commit.

Somebody else has to merge this PR, I'm not going to.  If you want to use `r.write_timestamp()` instead, an implementation for that is two commits up, as of "Removed duplicate write hook checking logic."

This PR should be merged or rebased onto v2.4.x and then cherry-picked onto next.


drivers/javascript/ast.coffee
```diff
@@ -1288,6 +1288,8 @@ rethinkdb.tableList = (args...) -> new TableList {}, args...
 
 rethinkdb.grant = (args...) -> new Grant {}, args...
 
+rethinkdb.writeTimestamp = (args...) -> new WriteTimestamp {}, args...
+
```
marshall007: Need to revert these driver changes.

marshall007: @srh need to revert these driver changes.

drivers/python/rethinkdb/query.py
```diff
@@ -95,6 +96,10 @@ def grant(*args, **kwargs):
     return ast.GrantTL(*args, **kwargs)
 
 
+def write_timestamp(*args):
+    return ast.WriteTimestamp(*args)
+
+
```
marshall007: Revert

srh:
Agh, just dropped the WRITE_TIMESTAMP stuff from the drivers...

srh:
Okay those are reverted...

src/pprint/js_pprint.cc
```diff
@@ -684,6 +684,7 @@ counted_t<const document_t> render_as_javascript(const ql::raw_term_t &t) {
 // - `should_continue_string` should add the new Term to the list if
 //   it should be used in a string-of-dotted-expressions context.  So
 //   like `r.foo(1).bar(2).baz(4)` instead of `r.eq(1, 2)`.
+//   ^^ What does this mean??
```
marshall007: I think it's referring to cases where you'd naively pretty-print `r.eq(var_0, 1)`, but you'd prefer `var_0.eq(1)`. The inverse example would be that you don't want `r.expr('foo').db()`, it should be `r.db('foo')`.

marshall007: Actually I'm wrong about the `r.eq()` example. It appears the pretty-printer always prefers `r.eq(args...)`, but I think the `.db()` one holds.

srh: Hopefully that comment will provide emotional support to other people that are confused.

danielmewes:
API and design sound great to me. I've only done a quick pass over the code so far. Will review more closely in the next days.

Can we add a test case for the timestamp API?

srh:
We can, if by we you mean somebody other than me, because I'm taking the next week off from working on RethinkDB.

marshall007:
@srh yea I'll take care of the polyglot tests. Should have something up later tonight.

srh:
Okay, that's great news.  I'm basically iPad-only for the next week so consider me to be mildly incapacitated.

srh:
Your changes look good to me @marshall007.

danielmewes:
Code looks good.
Thanks for making the `read_all_rows_filtered` thing and `deterministic_t` cleaner.

I say :shipit: 

Can you merge it or do you want me (or someone else) to do that?

srh:
I can't merge it, I needed to give an external software access to my Github account and all my orgs, so I removed myself from the RethinkDB org.

danielmewes:
Merged into v2.4.x, and cherry-picked into `next` as well.

srh:
Thanks Daniel.

robertjpayne:
Just running the polygot tests on v2.4.x branch is this still a valid test:

```
desc: Test that r.now() is allowed before changes() and after.
table_variable_name: tbl
tests:

    - py: tbl.changes().merge({'a': r.now()})
      ot: err('ReqlQueryLogicError')

    - py: tbl.merge({'a': r.now()}).changes()
```

It seems to fail the test because it works fine…

marshall007:
@robertjpayne that test was added as part of this PR when the behavior of `r.now()` in the context of changefeeds was still somewhat undefined, but that was prior to https://github.com/rethinkdb/rethinkdb/pull/6396/commits/e6c15e48f8945987b4c6afe48ac9333c1ac512ba, which made `r.now()` deterministic across batches.

I think that failing test should now be removed.

PR #6399, opened 2017-06-16
Fix compilation on macOS

robertjpayne:

### Description

macOS compilation on the `next` branch is currently broken regardless of how you configure it. These small tweaks fix it and also update a few dependencies for good will:

- Updates the min os deployment target to 10.9, this is necessary to be compatible with node 6.x due to issues with how c++ and c++11 work.
- Update curl to avoid compilation issues with SecureTransport and also ensure it doesn't attempt to compile http2 support even if libnghttp2 is on the system path
- Update openssl just for security reasons it should always be kept as up to date as possible
- Include icu4c headers because macOS doesn't ship with them

With these updates you can once again do `./configure --fetch all` and then `make -j8` and it compiles with no issues on macOS Sierra.

### DIFF
--- a/configure
+++ b/configure
@@ -21,7 +21,7 @@ init () {
     max_browserify_version=13.1.0
     min_python_version=2.6.0
     
-    osx_min_version=10.7
+    osx_min_version=10.9
 
     min_msbuild_version=14.0
     min_nmake_version=14.0
--- a/mk/support/pkg/curl.sh
+++ b/mk/support/pkg/curl.sh
@@ -7,15 +7,10 @@ src_url_sha1=f5193316e4b5ff23505cb09bc946763d35d02cd6
 pkg_configure () {
     local prefix
     prefix="$(niceabspath "$install_dir")"
-    local ssl_command
-    ssl_command="--with-ssl"
-    if [[ "$OS" = "Darwin" ]]; then
-        ssl_command="--with-darwinssl --without-ssl --without-nghttp2"
-    fi
     if [[ "$CROSS_COMPILING" = 1 ]]; then
         configure_flags="--host=$($CXX -dumpmachine)"
     fi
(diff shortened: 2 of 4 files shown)


srh:
Okay.  I like that SSL is upgraded.  I'd been putting that off...

Also, I'm glad that the min OS version is updated.  Personally I'd make it be 10.11.

Also, I thought v8 came with its own copy of ICU -- that's what we end up using.  So we fetch it as part of v8.  Is that no good anymore?

Anyway, regarding ICU, check out https://github.com/srh/rethinkdb/commit/94190efd1f4838e448e463c84a4ce697a1ed40ad -- we don't need ICU as a dependency, period.  I think that is the right solution for ICU build problems.

So I think you should drop your ICU changes, removing them from history so that cloning the repo doesn't take any longer, and cherry-pick my change that removes our dependency on ICU.

robertjpayne:
@srh oh interesting regarding v8's ICU stuff, I'll have a look and see if we can re-use any headers.

Do you have a PR for removing ICU yet?

robertjpayne:
@srh have another look, was able to get rid of ICU not sure why it constantly was giving me errors before.

I left deployment target as 10.9, I don't see any reason to go higher than what is necessary if it just works fine.

robertjpayne:
Also one final change, darwinssl in curl crashes for some reason -- I can't see a reason not to just link to openssl that is fetched or on the system path?

srh:
>  have another look, was able to get rid of ICU not sure why it constantly was giving me errors before.

You might want to `git clean -fxd` and rebuild again to make sure it really still works.

srh:
Okay, `next` still compiles on macOS for me!  I just merged this into next as of f5d98a1e5dccf64771b84c332b18731bdce5e956.

Could you do v2.3.x now? :-)

But I forgot to test it out on Linux so I'll do that first before closing.

PR #6400, opened 2017-06-16
Drop ICU dependency.

srh:

### Description
Removes need of ICU from RethinkDB.  We barely used it at all.


srh:
Okay, I merged it.  I also cherry-picked it into v2.4.x.  The reason is, I think it has a good chance of reducing build headaches in the future (as operating systems and libraries move forward and stuff breaks, next moves with it, while v2.4.x remains relatively unchanged).

PR #6405, opened 2017-06-18
Refactor pretty print code.

srh:

### Description
This refactors the pretty print logic.


srh:
This seems quite non-controversial (Who here has a problem with fewer allocations, avoiding atomic operations, fewer passes through the data, more concrete execution flow?  Raise yo hand!), so it's in `next` as of 6593452be283436b63d2dd67f4bf86db73abf609.

PR #6407, opened 2017-06-20
OpenSSL-related compilation fixes

srh:

### Description
This fixes some compatibility issues with OpenSSL versions.  In a new OpenSSL version, some struct definitions are now removed from header files, their implementation details abstracted away.

I suppose this fix should be backported to v2.3.x and v2.4.x as well as merged into next.

I think that BN_ext_get_uint64 and BN_ext_count_low_zero_bits could use a unit test before this change gets accepted, because I actually have not tested them in any way.

### DIFF
--- a/src/client_protocol/server.cc
+++ b/src/client_protocol/server.cc
@@ -174,12 +174,13 @@ void http_conn_cache_t::on_ring() {
 }
 
 size_t http_conn_cache_t::sha_hasher_t::operator()(const conn_key_t &x) const {
-    EVP_MD_CTX c;
-    EVP_DigestInit(&c, EVP_sha256());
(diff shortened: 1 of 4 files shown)


srh:
I strongly recommend you get this merged because it'll clearly reduce future compatibility pain (which I encountered in Debian Stretch).  Writing unit tests for this should be simple enough, and I recommend doing that too.

elfgoh:
This may be a fix for armhf compilation on Debian Jessie as detailed in #6427

srh:
Merged.
In next as of 624561559daf692583c375d84838d34e28798c66.
In v2.3.x as of 58c3a7c09b7060e605fad333745480f2661ecd8c.
In v2.4.x as of 3156820e058cd410078a4c39d5a2a85f6b714c07.

PR #6410, opened 2017-06-22
Serializer checksums

srh:

### Description
Adds serializer checksumming as described in http://samuelhughes.com/blog/7.html

Note that we checksum block writes when they come into the serializer less than 64 KB at a time.  Particularly noteworthy is that with the new flush interval behavior for soft durability writes (which is sitting in pull request #6392), we won't wastefully checksum blocks of small soft durability writes.

The parameter 64 KB has not been tuned.  You might want to adjust it.

Also, the advantages of using 2x32-bit Fletcher checksums versus 4x16-bit Fletcher is something I have not investigated.  I think ZFS or XFS or something uses 4x16-bit -- maybe they had a reason.

In the long-run, note that you could checksum blocks "earlier" when they've been recently processed by the CPU (so they're fresh in cache), and we could differentially update a checksum when modifying portions of a b-tree node (so we don't have to load the entire 4K block into L1 cache).  That's one thing Fletcher checksums get for you.

### DIFF
--- a/src/arch/arch.cc
+++ b/src/arch/arch.cc
@@ -18,8 +18,8 @@ void co_read(file_t *file, int64_t offset, size_t length, void *buf, file_accoun
 }
 
 void co_write(file_t *file, int64_t offset, size_t length, void *buf,
-              file_account_t *account, file_t::wrap_in_datasyncs_t wrap_in_datasyncs) {
+              file_account_t *account, datasync_op datasync) {
     io_coroutine_adapter_t adapter;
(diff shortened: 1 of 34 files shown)


srh:
In next as of 7be1b283fb05d3a7b7a385935d180d7ab85b93fb.

Stuck with 2x32-bit Fletcher checksums instead of 4x16-bit.
Stuck with 64 KB as the threshold above which we don't checksum stuff.

PR #6411, opened 2017-06-22
Avoid some uses of wall-clock time

srh:

### Description
This replaces some inappropriate uses of wall-clock time, users of `current_microtime`.  Instead they use a monotonic timer.

This fixes honest-to-god bugs in the presence of backwards-stepping wallclock time.


srh:
Pushed to next with commit 37617ab2d17e54180ab2e40ece43f63c6633b0c8.

PR #6431, opened 2017-07-11
Memory pool

thedrow:

### Description
This PR introduces a memory pool for segments in segmented vector. The pool allows reuse of segments and should improve performance,


srh:
segmented_vector_t already allocates in chunks of 16K elements, which means we're usually doing 64K-size or 128K-size allocations.  Object pools and such are hypothetically useful performance optimizations when doing lots of *small* allocations, which is the opposite of what segmented_vector_t does, except when adding/removing an element, setting the size from 0 to not-0, which it already handles.

Another problem with this is that who knows if object_pool actually frees memory when you've deallocated stuff.  Users of segmented_vector generally expect the segments to get freed when the segmented_vector is shrunk or cleared.

thedrow:
Regardless of what @srh has pointed out I came up with a more severe issue.
jemalloc/tcmalloc will allocate memory in the arena local to the thread. 
If we provide a chunk of memory from another thread's area will hurt performance for sure.
Any allocator we have needs to be jemalloc aware at the very least.

PR #6379, opened 2017-06-02
Add v8_2-HandleScope-protected.patch for OS X/modern clang

srh:
Edit: Disregard... somehow it seems this patch is now unnecessary.

Clang now complains about accessing a protected member function.
Maybe it now does this before instantiating.

- [x] I have read and agreed to the RethinkDB Contributor License Agreement http://rethinkdb.com/community/cla/

PR #6435, opened 2017-07-13
Fix OS X build in v2.3.x

AtnNn:
This is a port of #6379 for the v2.3.x branch


marshall007:
LGTM, fixes macOS build on my machine (noting that you have to `--allow-fetch --fetch openssl` and link your system version of Node).

AtnNn:
Squashed and merged as 1398daf

PR #6436, opened 2017-07-13
Make the free list use a single linked list

thedrow:

### Description
I removed the usage of segmented vector from the free list. Instead I replaced it with a single linked list that uses a pool allocator.
I don't think a segmented vector should have been used here in the first place because we don't use it as a vector.

I need to measure this somehow.

### DIFF
--- a/configure
+++ b/configure
@@ -20,7 +20,7 @@ init () {
     min_browserify_version=13.0.0
     max_browserify_version=13.1.0
     min_python_version=2.6.0
-    
+
     osx_min_version=10.9
 
     min_msbuild_version=14.0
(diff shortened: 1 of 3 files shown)


AtnNn:
Please do measure and share your results.

thedrow:
I first want to see it compile on CI before I do. It seems that requiring boost_system causes the build system to think it should fetch it :/

srh:
What is the benefit of this change?

srh:
In fact, this change should be rejected.  The segmented vector implementation releases unused memory if we shrink or empty out the free list.  But using a boost pool_allocator does promise that behavior.

thedrow:
Again, closing for the same reasoning as #6431.

PR #6450, opened 2017-07-22
Add v8 patch for Clang 9.0

robertjpayne:

Adds v8 patch to future proof compilation when Xcode 9 ships with Clang 9.0 in a month or so

### DIFF
--- a/mk/support/pkg/patch/v8_3-clang9.patch
+++ b/mk/support/pkg/patch/v8_3-clang9.patch
@@ -1,0 +1,12 @@
+diff -ruN --label original original ./src/compiler/instruction.h
+--- original
++++ ./src/compiler/instruction.h
+@@ -865,7 +865,7 @@ class InstructionBlock FINAL : public ZoneObject {
+ 
+ typedef ZoneDeque<Constant> ConstantDeque;
+ typedef std::map<int, Constant, std::less<int>,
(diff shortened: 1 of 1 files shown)


robertjpayne:
The only thing I could think of that is bad with this is `--without-nghttp2` on the curl configure. If you have system nghttp2 installed it'll take advantage of it and use http2.

But given that RethinkDB runs all HTTP calls in subprocesses I'm not sure we'd gain any benefit here? Does it keep the process alive to the point where using a single TCP connection for multiple requests would even work?

robertjpayne:
I have no idea why GitHub is showing diffs the two cherry picked commits that @srh already merged in #6399

robertjpayne:
If we commit this and then backport it and  #6399 into v2.4.x we can compile 2.4.x on macOS!

srh:
This looks good to me -- I say merge it.

srh:
This looks good to me -- I say merge it.

srh:
In next as of 88c1a8508504127391ff5df49128c30ae9d6b164.  Still needs backporting into v2.4.x as mentioned above.

srh:
Backported to v2.4.x in 07ae49f1c8c6d24c85c3d3ea9601e061ebe010df and v2.4.x compiles on macOS.

Thank you!

PR #6451, opened 2017-07-23
Fix some slowdowns when using get_field and default

### LINKED ISSUE #4271: datum_t::get_field() exhibits pathological performance on missing attributes
I have a database that consists of ~200,000 JSON documents.  I have noticed
vast differences on query time depending on the presence of a property.
For example, here is a query that looks for a match on a property named
"node".  This property is present in all documents, and the filter
returns a selection of 116 of them:

```
$ time rql 'r.table("dumps").filter({"node": "486Y9R1"}).pluck("dump")'

real    0m0.710s
user    0m0.121s
sys     0m0.051s
```

Considering that this property is unindexed, this performance is great.  If,
however, the query is on a property that is present in many fewer of the
documents, the query time is increased by an order of magnitude.  This
is a query of a property that is only present in ~12,000 of the documents:

```
$ time rql 'r.table("dumps").filter({"ticket": "OS-1028"}).pluck("dump")'

real    0m15.891s
user    0m0.125s
sys     0m0.054s
```

And if the filter is phrased as a function, the performance degrades by
another order of magnitude:

```
$ time rql 'r.table("dumps").filter(lambda dump: dump["ticket"] == "OS-1028").pluck("dump")'

real    3m8.122s
user    0m0.090s
sys     0m0.038s
```

Note that this cannot simply be explained by the additional overhead of
executing an arbitrary function; returning to the first query but rephrasing
it in terms of a function results in a much more modest hit:

```
$ time rql 'r.table("dumps").filter(lambda dump: dump["node"] == "486Y9R1").pluck("dump")'

real    0m1.314s
user    0m0.120s
sys     0m0.050s
```

I am running this on a DTrace-capable Linux platform (namely, a Linux
container on Joyent's Triton stack); looking into this with DTrace reveals
that in the most pathological case, we're spending a significant amount of
time in vfprintf():

```
# dtrace -n profile-1234hz'/arg1 != NULL && execname == "rethinkdb"/{@[usym(arg1)] = count()}' -n tick-1sec'/++i == 10/{trunc(@, 30); printa("%10@d %A\n", @); exit(0)}' -q
       195 libc.so.6`0x7ffffbc86e5f
       210 rethinkdb`0x972d10
       210 rethinkdb`_Z10force_readP13read_stream_tPvl
       220 rethinkdb`_ZN2ql7datum_t14data_wrapper_t11assign_copyERKS1_
       233 rethinkdb`cJSON_CreateStringN
       289 libc.so.6`0x7ffffbc86ea6
       299 rethinkdb`_ZN2ql20datum_get_array_sizeERK16shared_buf_ref_tIcE
       322 libpthread.so.0`__lll_unlock_wake
       325 rethinkdb`_ZN2ql31datum_deserialize_pair_from_bufERK16shared_buf_ref_tIcEm
       328 rethinkdb`_ZNK2ql7datum_t11as_json_rawEv
       369 libc.so.6`__printf_fp
       382 libpthread.so.0`pthread_mutex_unlock
       417 libc.so.6`0x7ffffbc724a8
       436 libc.so.6`0x7ffffbc724f8
       438 libc.so.6`0x7ffffbc724b7
       449 libc.so.6`0x7ffffbc724f0
       466 rethinkdb`_ZN2ql26datum_deserialize_from_bufERK16shared_buf_ref_tIcEm
       480 libstdc++.so.6`__gxx_personality_v0
       487 libpthread.so.0`__lll_lock_wait
       559 rethinkdb`_ZN2ql24datum_get_element_offsetERK16shared_buf_ref_tIcEm
       583 rethinkdb`_ZN2ql7datum_t14data_wrapper_t8destructEv
       637 libc.so.6`0x7ffffbc86ea2
       642 libpthread.so.0`pthread_mutex_lock
       841 rethinkdb`_ZNK14datum_string_t4sizeEv
       861 libc.so.6`strlen
      1068 libc.so.6`0x7ffffbc724a0
      1443 libc.so.6`_IO_default_xsputn
      1652 rethinkdb`malloc
      1666 rethinkdb`free
      2408 libc.so.6`vfprintf
```

Here's where those calls are coming from:

```
# dtrace -n pid14912::vfprintf:entry'{@[usym(ucaller)] = count()}'
dtrace: description 'pid14912::vfprintf:entry' matched 2 probes
^C

  libc.so.6`__vsnprintf_chk                                    142145
  libc.so.6`__vasprintf_chk                                    370300
```

Digging into where these are coming from:

```
#pragma D option strsize=128

pid$target::__vsnprintf_chk:entry
{
	self->buf = arg0;
}

pid$target::__vsnprintf_chk:return
/self->buf/
{
	@[copyinstr(self->buf)] = count();
	self->buf = NULL;
}

END
{
	trunc(@, 10);
}
```

Here's the output running for a few seconds on the pathological query:

```
$ dtrace -s ./re-snprintf.d -p 14912
dtrace: script './re.d' matched 3 probes
^C
CPU     ID                    FUNCTION:NAME
  2      2                             :END 

  No attribute `ticket` in object:
{
	"cmd":	"zabbix_sender",
	"dump":	"/thoth/stor/thoth/b811bc0736b2cf885d700ec47d33821f/core.z               28
  No attribute `ticket` in object:
{
	"cmd":	"zabbix_sender",
	"dump":	"/thoth/stor/thoth/b86a105f641545827b96ade8977c8753/core.z               28
  No attribute `ticket` in object:
{
	"cmd":	"zdb",
	"dump":	"/thoth/stor/thoth/b795da70280bccb539826e4555ce1354/core.zdb.87172",               28
  No attribute `ticket` in object:
{
	"cmd":	"zdb",
	"dump":	"/thoth/stor/thoth/b890ccbbd943a35c76753f3f5c4c049c/core.zdb.54093",               28
  No attribute `ticket` in object:
{
	"dump":	"/thoth/stor/thoth/b80c1a837b2faf8ae44b5e77dda9831e/vmcore.0",
	"id":	"b80c1a837b2f               28
  Evaluating pluck.                                                40
  Evaluating bracket.                                          175367
  Evaluating EQ.                                               175368
  Evaluating var.                                              175407
  Evaluating datum.                                            185719
```

If it isn't clear, the objects themselves that lack the specified property are
clearly be snprintf()'d in a human-readable way!  Looking for the "No
attribute" string in the source reveals the source:

```
datum_t datum_t::get_field(const datum_string_t &key, throw_bool_t throw_bool) const {
    // Use binary search on top of unchecked_get_pair()
    size_t range_beg = 0;
    // The obj_size() also makes sure that this has the right type (R_OBJECT)
    size_t range_end = obj_size();
    while (range_beg < range_end) {
        const size_t center = range_beg + ((range_end - range_beg) / 2);
        auto center_pair = unchecked_get_pair(center);
        const int cmp = key.compare(center_pair.first);
        if (cmp == 0) {
            // Found it
            return center_pair.second;
        } else if (cmp < 0) {
            range_end = center;
        } else {
            range_beg = center + 1;
        }
        rassert(range_beg <= range_end);
    }

    // Didn't find it
    if (throw_bool == THROW) {
        rfail(base_exc_t::NON_EXISTENCE,
              "No attribute `%s` in object:\n%s", key.to_std().c_str(), print().c_str());
    }
    return datum_t();
}
```

Here's a script which explores the amount of time we're actually spending in
datum_t::get_field() for missing properties:

```
#pragma D option quiet

BEGIN
{
	start = timestamp;
}

pid$target::_ZNK2ql7datum_t9get_fieldERK14datum_string_tNS_12throw_bool_tE:entry
{
	self->ts = vtimestamp;
	@ttl = count();
}

pid$target::__cxa_throw:entry
/self->ts/
{
	@missing = sum(vtimestamp - self->ts);
	@ttlmissing = count();
	self->ts = 0;
}

pid$target::_ZNK2ql7datum_t9get_fieldERK14datum_string_tNS_12throw_bool_tE:return
/self->ts/
{
	@found = sum(vtimestamp - self->ts);
	@ttlfound = count();
	self->ts = 0;
}

sched:::on-cpu
/pid == $target/
{
	self->on = vtimestamp;
}

sched:::off-cpu
/pid == $target && self->on/
{
	@alltime = sum(vtimestamp - self->on);
}

tick-100ms
/timestamp - start > 60 * 1000000000/
{
	normalize(@found, 1000000);
	normalize(@missing, 1000000);
	normalize(@alltime, 1000000);

	printf("%10d => wall clock ms\n", (timestamp - start) / 1000000);
	printa("%10@d => total CPU ms\n\n", @alltime);
	printa("%10@d => total calls to datum_t::get_field()\n", @ttl);
	printa("%10@d => calls on found attributes\n", @ttlfound);
	printa("%10@d => calls on missing attributes\n\n", @ttlmissing);
	printa("%10@d => total CPU ms on found attributes\n", @found);
	printa("%10@d => total CPU ms on missing attributes\n", @missing);
	exit(0);
}
```

Running this for one of the (several) minutes that the pathological query is
running yields:

```
     60081 => wall clock ms
    239799 => total CPU ms

    316125 => total calls to datum_t::get_field()
     19194 => calls on found attributes
    296931 => calls on missing attributes

       103 => total CPU ms on found attributes
    112481 => total CPU ms on missing attributes
```

This is telling us that at least half of our CPU time is being spent in
datum_t::get_field() on missing attributes -- with a whopping ~380 usecs per
call.  Further, it seems quite possible (if not likely) that the other half of
our CPU time is being spent on the mechanics of throwing and then (presumably)
discarding the exception.  It seems at the very, very least
datum_t::get_field() should not be printing an entire verbatim, human-readable
copy of the object before it throws that an attribute is missing.  Beyond
that, it seems likely that performance could be substantially improved by
using (or introducing, as necessary) a method that would get a property
without throwing an exception on its absence (that is, returning failure
instead of throwing an exception).

robertjpayne:

### Description

I know v2.4.x is somewhat frozen but this is a small tweak for big performance gains when using `default` in conjunction with `get_field`.

It somewhat fixes #4271. I'd appreciate a code review here as I'm unsure the implications of doing a `catch (...)` to reset the env is there a better way to do this?

### DIFF
--- a/src/rdb_protocol/env.hpp
+++ b/src/rdb_protocol/env.hpp
@@ -157,6 +157,9 @@ public:
 
     rdb_context_t *get_rdb_ctx() { return rdb_ctx_; }
 
+    // This is set when inside a default term
+    bool eval_in_default = false;
+
 private:
     static const uint32_t EVALS_BEFORE_YIELD = 256;
     uint32_t evals_since_yield_;
--- a/src/rdb_protocol/terms/error.cc
+++ b/src/rdb_protocol/terms/error.cc
@@ -32,7 +32,9 @@ private:
         scoped_ptr_t<exc_t> err;
         scoped_ptr_t<val_t> v;
         try {
+            env->env->eval_in_default = true;
             v = args->arg(env, 0);
+            env->env->eval_in_default = false;
             if (v->get_type().is_convertible(val_t::type_t::DATUM)) {
                 func_arg = v->as_datum();
                 if (func_arg.get_type() != datum_t::R_NULL) {
@@ -42,6 +44,7 @@ private:
                 return v;
             }
         } catch (const exc_t &e) {
+            env->env->eval_in_default = false;
             if (e.get_type() == base_exc_t::NON_EXISTENCE) {
                 err.init(new exc_t(e));
                 func_arg = datum_t(e.what());
@@ -49,16 +52,20 @@ private:
                 throw;
             }
         } catch (const datum_exc_t &e) {
+            env->env->eval_in_default = false;
             if (e.get_type() == base_exc_t::NON_EXISTENCE) {
                 err.init(new exc_t(e.get_type(), e.what(), backtrace()));
                 func_arg = datum_t(e.what());
             } else {
                 throw;
             }
+        } catch (...) {
+            env->env->eval_in_default = false;
+            throw;
         }
+
         r_sanity_check(func_arg.has());
-        r_sanity_check(func_arg.get_type() == datum_t::R_NULL
-                       || func_arg.get_type() == datum_t::R_STR);
+        r_sanity_check(func_arg.get_type() == datum_t::R_NULL || func_arg.get_type() == datum_t::R_STR);
         try {
             scoped_ptr_t<val_t> def = args->arg(env, 1);
             if (def->get_type().is_convertible(val_t::type_t::FUNC)) {
--- a/src/rdb_protocol/terms/obj_or_seq.cc
+++ b/src/rdb_protocol/terms/obj_or_seq.cc
@@ -285,7 +285,16 @@ private:
     virtual scoped_ptr_t<val_t> obj_eval(
         scope_env_t *env, args_t *args, const scoped_ptr_t<val_t> &v0) const {
         datum_t d = v0->as_datum();
-        return new_val(d.get_field(args->arg(env, 1)->as_str()));
+        datum_t r;
+        if (env->env->eval_in_default) {
+            r = d.get_field(args->arg(env, 1)->as_str(), NOTHROW);
+            if (r.get_type() == datum_t::UNINITIALIZED) {
+                r = datum_t::null();
+            }
+        } else {
+            r = d.get_field(args->arg(env, 1)->as_str(), THROW);
+        }
+        return new_val(r);
     }
     virtual const char *name() const { return "get_field"; }
 };


robertjpayne:
I'll keep working on this, it only really solves a `get_field` that is followed by a `default`.

robertjpayne:
This actually sadly is not going to work, it fails on filters like `x.getField("role").do(y => 1).default("other")`

PR #6453, opened 2017-07-23
Use https urls

NicoHood:


gabor-boros:
This issue was solved in the extracted python driver.

Issue #6352, opened 2017-05-05
`.default()` is a potentially giant performance pitfall


cristicbz:
`some_object(key).default(value)` is _extremely_ slow if `some_object` is a large object (presumably because the non-existence error serializes the entire object to a string):

In IPython:
```
In [70]: z = r.expr({os.urandom(18).encode('hex'): os.urandom(4).encode('hex') for _ in xrange(1000)})

In [71]: %timeit -r1 -n1 r.table('comments').filter(lambda x: (z[x['ci']] != x['di']).default(True)).count().run()
1 loop, best of 1: 3m 43.01 s per loop

In [72]: %timeit -r1 -n1 z.do(lambda e: r.table('comments').filter(lambda x: e.has_fields(x['ci']).not_().or_(e[x['ci']] != x['di'])).count()).run()
1 loop, best of 1: 2.08 s per loop
```

Where comments is a table with about 700K rows. The difference is HUGE! We had this both in some queries, but also in some index functions which made inserts extremely slow.

I think `default` should come with performance warning in the docs, but also there's a case for a `get_or_default` function on objects, to avoid having to write the `has_fields` version all the time.

I think this is the issue in #6119 too.

AtnNn:
Thanks for opening the issue. This seems closely related to #4271, which seems to indicate that a failed `get_field` is slow even without the `default`.

robertjpayne:
Yea just digging into this specifically I don't think the `default` is actually the issue. Rather it's the `get_field` lookup which is implicit via brackets on `z[x['ci']]`.

I'd close this in favour of #4271 as well.

cristicbz:
Fair enough. The thing is this issue makes `get_field.default`into an antipattern (with quadratic perf), compared to the linear `branch.has_fields.get_field`. It feels like at least a documented perf warning on `default`'s page is warranted.

robertjpayne:
@cristicbz I mean really this should just be fixed, it'd be a HUGE performance win for a lot of queries, in my use of RethinkDB I used `default` *a lot*.

thelinuxlich:
@srh, duplicate of #1232

PR #6454, opened 2017-07-25
Avoid throwing exceptions on non existence in get_field & bracket

robertjpayne:

### Description

#6352 and #4271 outline a lot more detail but currently non existence in `get_field` and `bracket` terms throw exceptions, these exceptions are caught by `default` but at the cost of about 10-20 microseconds per exception.

When doing large filters this can add up to huge overheads.

Unfortunately it's impossible to optimise this away in all cases, there really are cases where we need the exception to unwind the stack and back out to the `default` term. However, most users are probably using simple queries like `r.expr({name: {first_name: "Robert", last_name: "Payne"}})("first")("middle").default("Unknown")` where the `get_field` and `bracket` terms are immediately followed by the `default` term.

This PR addresses those use cases by re-writing the `default` term when it is preceded by either a `get_field` or `bracket`. It adds an internal optarg that is used at eval time to return null rather than throw an exception.

In addition the `get_field` and `bracket` terms also re-write themselves copying this optarg if their preceeding term is either a `get_field` or `bracket`. This allows chained uninterrupted `get_field` and `bracket` terms to take advantage of this optimisation.

In my case this cuts down execution time by roughly half and since it uses the optarg it's impact is fairly small as it defaults back to throwing exceptions when it cannot optimise it away such as `r.expr({name: {first_name: "Robert", last_name: "Payne"}})("first").map(x => x.toUpperCase()).default("Unknown")`

### DIFF
--- a/src/rdb_protocol/optargs.cc
+++ b/src/rdb_protocol/optargs.cc
@@ -39,6 +39,7 @@ scoped_ptr_t<val_t> global_optargs_t::get_optarg(env_t *env, const std::string &
 static const std::set<std::string> acceptable_optargs({
     "_EVAL_FLAGS_",
     "_NO_RECURSE_",
+    "_NON_EXISTENCE_NULL_",
     "_SHORTCUT_",
     "array_limit",
     "attempts",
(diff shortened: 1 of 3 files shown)


robertjpayne:
This PR can also be back ported to 2.4 if desired. And before it's completed I'm going to try and get this to work for `nth` terms as well.

robertjpayne:
Hm I may hold off on merging this, it causes a lot of the RQL tests to fail

Issue #6095, opened 2016-09-09
Segmentation fault in RDBInterrupt unit test (related to write hooks?)


danielmewes:
I'm getting this reliably on `next` using a debug-mode binary:

```
rethinkdb/build/debug_clang$ ./rethinkdb-unittest --gtest_filter=RDBInterrupt.InsertOp
Note: Google Test filter = RDBInterrupt.InsertOp
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from RDBInterrupt
[ RUN      ] RDBInterrupt.InsertOp
Version: rethinkdb 2.3.2-windows-beta-277-gccc952-dirty (debug) (CLANG 3.5.0 (tags/RELEASE_350/final))
error: Error in thread 1 in src/arch/runtime/thread_pool.cc at line 420:
Previous message may not have been written to the log file (logging module is not yet initialized).
error: Segmentation fault from reading the address (nil).
Previous message may not have been written to the log file (logging module is not yet initialized).
error: Backtrace:
Previous message may not have been written to the log file (logging module is not yet initialized).
error: Thu Sep  8 18:44:12 2016

       1 [0x26cde2b]: rethinkdb_backtrace(void**, int) at rethinkdb_backtrace.cc:123
       2 [0x268cc40]: backtrace_t::backtrace_t() at backtrace.cc:213 (discriminator 1)
       3 [0x268d7bc]: lazy_backtrace_formatter_t::lazy_backtrace_formatter_t() at backtrace.cc:299
       4 [0x268c8c7]: format_backtrace(bool) at backtrace.cc:208
       5 [0x2690ecc]: report_fatal_error(char const*, int, char const*, ...) at errors.cc:86
       6 [0x17f275c]: linux_thread_pool_t::fatal_signal_handler(int, siginfo_t*, void*) at thread_pool.cc:420
       7 [0x7fad747d8330]: /lib/x86_64-linux-gnu/libpthread.so.0(+0x10330) [0x7fad747d8330] at 0x7fad747d8330 (/lib/x86_64-linux-gnu/libpthread.so.0)
       8 [0x17af4dc]: home_thread_mixin_t::home_thread() const at threading.hpp:54
       9 [0x21b5f09]: table_meta_client_t::get_config(uuid_u const&, signal_t*, table_config_and_shards_t*) at table_meta_client.cc:98
[...]

PR #6461, opened 2017-08-01
Fix interrupt tests #6095

robertjpayne:

### Description

This fixes the disabled interrupt tests that were causing segfaults. The segfaults occurs because the unit test environment doesn't currently mock the `table_meta_client_t` so a get write hook causes a null pointer dereference and crashes the process.

Mocking the `table_meta_client_t` looked extremely extensive so instead I opted to alter the get_write_hook method to exit early if ignoring write hooks which prevents the nullptr from being used.

Addresses #6095

### DIFF
--- a/src/rdb_protocol/real_table.cc
+++ b/src/rdb_protocol/real_table.cc
@@ -224,11 +224,13 @@ optional<counted_t<const ql::func_t> > real_table_t::get_write_hook(
     ignore_write_hook_t ignore_write_hook) {
     optional<counted_t<const ql::func_t> > write_hook;
     table_config_and_shards_t config;
+    if (ignore_write_hook == ignore_write_hook_t::YES) {
+        return write_hook;
+    }
 
     m_table_meta_client->get_config(uuid, env->interruptor, &config);
 
-    if (config.config.write_hook &&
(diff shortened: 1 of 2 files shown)


marshall007:
@robertjpayne I'm not sure if returning early from `get_write_hook` is necessarily valid. It looks like there might be potential side effects from calling `m_table_meta_client->get_config(...)` that would no longer occur when passing the `ignore_write_hook: true` optarg.

See [src/clustering/table_manager/table_meta_client.cc#L114](https://github.com/rethinkdb/rethinkdb/blob/v2.4.x/src/clustering/table_manager/table_meta_client.cc#L114)

robertjpayne:
Ugh yea I totally missed that (why code reviews are important!) Well I have no idea how we get this into the test rig so we can re-enable these tests then.

robertjpayne:
Actually the side-effects here are probably not necessary. The side effects is that if the lookup fails it'd throw an exception.

But nothing else uses the data returned from the lookup and thus if something else needs to get the table config it should throw the same exceptions.

It's possibly even smarter to skip the `get_config` call to avoid the overhead.

marshall007:
After discussion on Slack, I agree it shouldn't matter and is perhaps an optimization.

PR #6489, opened 2017-09-07
Enable use of native Promise library for Node.js driver

### LINKED ISSUE #6459: Use native ES6 Promises (move away from bluebird) for the javascript driver
Since ES6 Promises are available for all important node versions ([http://node.green/#ES2015-built-ins-Promise](http://node.green/#ES2015-built-ins-Promise)) I think it would be possible to move away from bluebird  (which is currently the only dependency for the javascript driver!)

Some implementation where you can easily plug in your favorite promise library would be nice too ([like something mongoose did](http://mongoosejs.com/docs/promises.html))

ralphholzmann:
fixes #6459

### Description
Adds a `setPromise` method to the exported `r` object which overrides the promise implementation in the `ast`, `cursor`, and `net` files of the JavaScript driver.

### DIFF
--- a/drivers/javascript/util.coffee
+++ b/drivers/javascript/util.coffee
@@ -1,3 +1,4 @@
+Promise = require('bluebird')
 err = require('./errors')
 protodef = require('./proto-def')
 
@@ -122,6 +123,23 @@ mkSeq = (response, opts) -> recursivelyConvertPseudotype(response.r, opts)
 mkErr = (ErrClass, response, root) ->
     new ErrClass(mkAtom(response), root, response.b)
 
+module.exports.asCallback = (promise, callback) ->
+  if typeof callback !== 'function' return promise
+  promise.then((value) -> callback(null, value), callback)
+  promise
+
+module.exports.fromCallback = (fn) ->
+  new Promise((resolve, reject) ->
+    try
+      fn (error, result) ->
+        if error
+          reject error
+        else
+          resolve result
+    catch error
+      reject error
+  )
+
 errorClass = (errorType) =>
     switch errorType
         when protoErrorType.QUERY_LOGIC         then err.ReqlQueryLogicError
@@ -140,3 +158,4 @@ module.exports.mkAtom = mkAtom
 module.exports.mkSeq = mkSeq
 module.exports.mkErr = mkErr
 module.exports.errorClass = errorClass
+module.exports._setPromise = (PromiseImplementation) -> Promise = PromiseImplementation;
--- a/drivers/javascript/ast.coffee
+++ b/drivers/javascript/ast.coffee
@@ -9,6 +9,7 @@ Promise = require('bluebird')
 ar = util.ar
 varar = util.varar
 aropt = util.aropt
+asCallback = util.asCallback
 
 # rethinkdb is both the main export object for the module
 # and a function that shortcuts `r.expr`.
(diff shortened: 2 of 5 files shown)


marshall007:
@ralphholzmann modifying the global `Promise` in each of these files is a breaking change. The driver depends on several utility methods exposed by `bluebird`.

In order to make this work you would need to wrap the returned promises in a custom implementation, not simply modify it globally.

ralphholzmann:
Got it -- I'll make some updates

ralphholzmann:
@marshall007 I only saw the `fromNode` (now, [fromCallback](http://bluebirdjs.com/docs/api/promise.fromcallback.html)) and `nodeify` (now, [asCallback](http://bluebirdjs.com/docs/api/ascallback.html)) utility methods being used from bluebird, so I added them as utils. Did I miss anything here? Let me know if this approach will work.

marshall007:
@ralphholzmann here are a [couple more examples](https://github.com/rethinkdb/rethinkdb/blob/e48cfb868c32d6549dcb392a177effd0d8a1687e/drivers/javascript/cursor.coffee#L290-L291). In this case using `Promise.any(<array>)` and `.catch(<ErrorClass>, <callback>)` respectively.

In any case, I think internally relying on the user-specified promise implementation is too unpredictable and would be very inconvenient for driver development. It's useful to be able to rely on a specific implementation internally and take advantage of its documented features.

drivers/javascript/util.coffee
```diff
@@ -122,6 +123,23 @@ mkSeq = (response, opts) -> recursivelyConvertPseudotype(response.r, opts)
 mkErr = (ErrClass, response, root) ->
     new ErrClass(mkAtom(response), root, response.b)
 
+module.exports.asCallback = (promise, callback) ->
+  if typeof callback !== 'function' return promise
+  promise.then((value) -> callback(null, value), callback)
```
chrisvariety: Should we consider `.catch` here and include error handling? e.g. https://github.com/petkaantonov/bluebird/blob/3746b7eca90dd8b11af73db5d30cf46d7dd90f9b/src/nodeify.js#L39

marshall007:
@ralphholzmann here are a [couple more examples](https://github.com/rethinkdb/rethinkdb/blob/e48cfb868c32d6549dcb392a177effd0d8a1687e/drivers/javascript/cursor.coffee#L290-L291). In this case using `Promise.any(<array>)` and `.catch(<ErrorClass>, <callback>)` respectively.

In any case, I think internally relying on the user-specified promise implementation is too unpredictable and would be very inconvenient for driver development. It's useful to be able to rely on a specific implementation internally and take advantage of its documented features.

I think the only reliable way to implement this is to simply wrap all returned promises with the custom implementation (if specified).

drivers/javascript/util.coffee
ralphholzmann: curious how errors aren't being handled here?

ralphholzmann:
@marshall007 the goal here (as stated by #6459) is to enable the opt-in use of native ES6 promises that ship with V8, because they're not only faster but they also produce a much cleaner stack trace when debugging. Now it's possible that you could swap in *any* Promises A+ standard compliant promise library (aka "custom implementation") if you'd like, but that's tangential to the goal.

If we went the route of simply wrapping bluebird promises with native promises, the change would be pointless -- speed would be *worse* and the stack traces would be the same.

I'm more than happy to convert the rest of the bluebird utility methods into methods in `util.js`, but I also don't want to waste my time if the PR will just be rejected. Let me know your thoughts when you can.

ralphholzmann:
Added last commits and updated the issue title for accuracy

marshall007:
> If we went the route of simply wrapping bluebird promises with native promises, the change would be pointless -- speed would be worse and the stack traces would be the same.

[Last I checked](https://github.com/petkaantonov/bluebird/tree/master/benchmark) bluebird's implementation is still significantly faster than native promises. If stack traces are the primary concern, take a look at [`Promise.longStackTraces()`](http://bluebirdjs.com/docs/api/promise.longstacktraces.html).

I don't think the goal of this PR should be to replace the internal promise implementation. Wrapping all returned promises in the user-specified implementation would be a useful feature, but only if it doesn't modify internal behavior.

ralphholzmann:
Alright, I'll abandon this effort then. Thanks for your time.

PR #6492, opened 2017-09-10
Fix auth permissions bug

robertjpayne:

### Description

When conversion from boost optional to homegrown optional was done an existence gaurantee check was changed to a truthy check which causes a crasher.

### DIFF
--- a/src/clustering/administration/auth/permissions.cc
+++ b/src/clustering/administration/auth/permissions.cc
@@ -145,7 +145,7 @@ void permissions_t::set_config(tribool config) {
 }
 
 void permissions_t::set_connect(tribool connect) {
-    guarantee(m_connect == make_optional(tribool::True));
+    guarantee(m_connect.has_value());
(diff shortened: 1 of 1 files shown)


src/clustering/administration/auth/permissions.cc
```diff
@@ -145,7 +145,7 @@ void permissions_t::set_config(tribool config) {
 }
 
 void permissions_t::set_connect(tribool connect) {
-    guarantee(m_connect == make_optional(tribool::True));
+    guarantee(m_connect.has_value());
```
marshall007: For future reference, the bug was introduced in https://github.com/rethinkdb/rethinkdb/commit/e705583d8c6dfd87ac90cbd9946b1810fee13552#diff-5f4c6e389529139aa0cef0416cf8bcf4R149 which changed the behavior from `m_connect != null` to `m_connect == true`.

PR #6493, opened 2017-09-10
v2.4.x tests fixes

robertjpayne:

### Description

Some small fixes for various RQL tests so they can run.

### DIFF
--- a/test/common/http_support/server.py
+++ b/test/common/http_support/server.py
@@ -11,7 +11,7 @@ import utils
 
 def __runServer(httpbinPort=0, httpPort=0, sslPort=0):
     
-    import twisted.internet
+    import twisted.internet.reactor
     import twisted.internet.ssl
     import twisted.web
     import twisted.web.static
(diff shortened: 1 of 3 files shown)


test/rql_test/connections/r_http.py.httpbin
```diff
@@ -137,8 +137,8 @@ class TestHttpTerm(rdb_unittest.RdbTestCase):
     def test_redirects(self):
         url = self.getHttpBinURL('redirect', '2')
         
-        self.assertRaisesRegex(r.ReqlRuntimeError, self.err_string('GET', url, 'status code 302'), r.http(url, redirects=0).run, self.conn)
-        self.assertRaisesRegex(r.ReqlRuntimeError, self.err_string('GET', url, 'Number of redirects hit maximum amount'), r.http(url, redirects=1).run, self.conn)
```
robertjpayne: These APIs were saying they were undefined/missing so updating them to the newer interface

test/rql_test/src/regression/6108.yaml
```diff
@@ -1,4 +1,5 @@
 desc: Test eqJoin on artificial tables 
 tests:
-  - cd: r.db('rethinkdb').table('table_status', identifier_format='uuid').eq_join('raft_leader', r.db('rethinkdb').table('server_status'))
```
robertjpayne: These tests need to be language specific

test/common/http_support/server.py
```diff
@@ -23,7 +23,7 @@ def __runServer(httpbinPort=0, httpPort=0, sslPort=0):
     # -- generate self signed SSL certificate
     
     privateKey = tempfile.NamedTemporaryFile(suffix='.pem')
-    subprocess.check_call(['openssl', 'genrsa'], stdout=privateKey, stderr=subprocess.PIPE)
+    subprocess.check_call(['openssl', 'genrsa', '2048'], stdout=privateKey, stderr=subprocess.PIPE)
```
robertjpayne: This test was failing with "key too short" so increasing key length to 2048 which is pretty standard. It doesn't slow down that much since RSA keys are pretty quick.

PR #6498, opened 2017-09-18
Export hidden cursor classes

ralphholzmann:

### Description
In building a new Node.js based model library for RethinkDB, we've found that it can sometimes be beneficial to be able to test what class of response is being returned from `rethinkdb`. This commit exports all of the cursor classes (under `r.net.cursors`) so we'll be able to do things such as

```js
if (result instanceof ArrayResult) {
  // something
}
```


srh:
What are you using this for?  Is it to see whether a result is an array (which, I presume, contains all returned data and doesn't need to fetch more) or not?

This is an API change so we don't want to expose driver internals unnecessarily.

srh:
Hey @ralphholzmann, Can you please confirm whether what you want is specifically to see if the result is an array or if it's a cursor?  That you don't specifically want more?

ralphholzmann:
Heyo -- sorry for the late reply here. Yes, that's essentially what I need, some way to differentiate.

srh:
I'll make an issue before closing if not a patch that sees if a cursor is an array vs stream.

PR #6535, opened 2017-11-05
Fix building on Windows in next

srh:

### Description
This fixes building on Windows.

Some things broke from the recent ticks_t refactor, and also the removal of the ICU dependency made us mistakenly not link against ICU's libraries (which we should do because v8 still uses them).


srh:
In next as of b93054889bbf.

Issue #6545, opened 2017-11-13
Java driver write hook support


srh:
Right now write hook support in the Java driver is incomplete.  There are no write hook commands, and the command `make update-driver`, which you run in the directory `drivers/java`, will add the setWriteHook and addWriteHook term types (but not driver commands).  I'm 99% sure we need to add the commands to `drivers/java/term_info.json`.

srh:
It turns out the Java driver tests don't actually work.  `make java-test` doesn't work.

srh:
The current status on this is that there's one write hook test to fix so that the tests can compile:

```
    py: tbl.set_write_hook(lambda ctx: 1)
```

The problem is, this doesn't type-check when converted into Java code, because the function must have three arguments.

PR #6565, opened 2017-12-12
Java driver write hook support for 2.4

srh:

### Description
This adds write hook support for the Java Driver, fixing #6545.

It also fixes the Java Driver testing framework so that it's close to working.

The commit before the last makes the test framework run (and 3 fail), but right now it doesn't, with a compilation error.

A lot of generated files are modified.


srh:
Squashed a bit, removed the "test_user" commit which is probably wrong, and pushed to v2.4.x with

1c343f3e8d40364eeabd8846bb7205641884308f
5786ae1f3878aa77ca6f9aebe8fd59b229864904

and next with

43c47df2a6c443af86e611d82ca6da238f734178
b2412fe679d06cee05a18ebd5ec13ed1c6509e3c

PR #6574, opened 2017-12-21
Make r.iso8601 round, not truncate.

### LINKED ISSUE #6069: Python datetime not rounded like r.time()
r.time() will round the number of milliseconds, but Python datetime does not do the same thing. This results in different results when using the same inputs through the different interfaces. It can also cause problems during floating point conversions when reading values and then writing them back (i.e. the timestamp can become X.ABC999 in the Python datetime when it should be X.ABD000 where C is D-1).

Here's a simple script to demonstrate the difference in behavior:

```
#!/usr/bin/env python3

import datetime

import pytz
import rethinkdb as r

conn = r.connect(host="localhost", db="test")

r.table_create("test_time").run(conn)
truth = datetime.datetime(2016, 1, 1, 0, 0, 0, 1000, pytz.utc)
r.table("test_time").insert({"time": truth}).run(conn)
r.table("test_time").insert({"time": datetime.datetime(2016, 1, 1, 0, 0, 0, 999, pytz.utc)}).run(conn)
r.table("test_time").insert({"time": r.time(2016, 1, 1, 0, 0, 0.000999, "Z")}).run(conn)
for row in r.table("test_time").run(conn):
    print(row["time"], row["time"] == truth)
```

srh:

### Description
Fixes #6069.

### DIFF
--- a/src/rdb_protocol/pseudo_time.cc
+++ b/src/rdb_protocol/pseudo_time.cc
@@ -198,7 +198,7 @@ std::string date(const std::string &s, date_format_t *df_out) {
 }
 
 // Sanitize a time.
-std::string time(const std::string &s) {
+std::string time(reql_version_t reql_version, const std::string &s) {
     std::string out;
     size_t at = 0;
     mandatory_digits(s, 2, &at, &out);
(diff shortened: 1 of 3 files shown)


srh:
Pushed
v2.4.x 72d500d9bdb509b0d70263cb22b3cac1f3d9bbb8
next 89276127e710ff9a90b63765019eb36b6822da6c

src/rdb_protocol/pseudo_time.cc
```diff
@@ -215,16 +215,22 @@ std::string time(const std::string &s) {
...
-        while (at < s.size() && read < 3) {
-            mandatory_digits(s, 1, &at, &out);
-            read += 1;
-        }
-        while (at < s.size()) {
-            mandatory_digits(s, 1, &at, NULL);
-        }
-        while (read++ < 3) {
-            out += '0';
+        if (reql_version < reql_version_t::v2_4) {
```
daveisfera: Why does version matter here? Can't it just be changed to always do this behavior?

srh: Regular queries run with the latest reql_version value, but old secondary indexes' functions need to run with whatever reql_version value they were created with, in order to work properly.

daveisfera: The upgrade process states that [secondary indexes need to be rebuilt during upgrade](https://rethinkdb.com/docs/migration/#rebuild-indexes), so I don't see why this needs to be mainted since it's part of 2.4 and will require an upgrade.

srh: It says that secondary indexes *should* be rebuilt.

daveisfera: Just trying to keep the code simpler and avoid potential bugs down the road. Sorry for the noise.

srh: It's no problem.  Eventually the old pre-2.4 behavior will be dropped (being optimistic about future releases happening here...)

Issue #6568, opened 2017-12-17
Failed build (cannot bind packed field) on Ubuntu


p00temkin:
OS: Ubuntu 16.04

Steps:
git clone https://github.com/rethinkdb/rethinkdb.git
apt-get install -y gcc g++
apt-get install -y protobuf-compiler
apt-get install -y nodejs
apt-get install -y npm
apt-get install -y coffeescript
ln -s /usr/bin/nodejs /usr/bin/node
cd rethinkdb
./configure --allow-fetch
make

Error:
    [318/468] CC build/release/obj/serializer/log/metablock_manager.o
src/serializer/log/metablock_manager.cc: In member function ‘void metablock_manager_t::co_start_existing(file_t*, bool*, log_serializer_metablock_t*)’:
src/serializer/log/metablock_manager.cc:359:61: error: cannot bind packed field ‘mb->crc_metablock_t::version’ to ‘long int&’
             indices_by_version.push_back(std::make_pair(mb->version, i));
                                                             ^
src/build.mk:379: recipe for target 'build/release/obj/serializer/log/metablock_manager.o' failed
make[1]: *** [build/release/obj/serializer/log/metablock_manager.o] Error 1
Makefile:52: recipe for target 'make' failed
make: *** [make] Error 2

Additional notes: 
Same result using "make ALLOW_WARNINGS=1"

srh:
Thank you for the report.  The following instructions work for me, and there's no reason I'd expect a different result for you, having run the steps you followed, if you're running on amd64 like I am.

```
apt-get install build-essential protobuf-compiler python \
    libprotobuf-dev libcurl4-openssl-dev libboost-all-dev \
    libncurses5-dev libjemalloc-dev wget m4 g++ libssl-dev
./configure --allow-fetch
make
```

Are you running this on a 32-bit installation?  That might explain why you get a different result.  If so I'll be happy to fix it,

jtc3161:
I the same build error as p00temkin. OS information:
```
$ uname -a
Linux XXXXXXXXX 4.4.0-103-generic #126~14.04.1-Ubuntu SMP Mon Dec 4 19:33:04 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

p00temkin:
Hi @srh,

So far I am getting the same build error on two 64-bit hosts:
host-1$ uname -a
[...]

PR #6580, opened 2017-12-27
Avoid making reference to packed field.

srh:
Fixes build on Xenial, Trusty, older GCCs.


srh:
Fixes #6568.  In next as of 0e91f4908ed249c8be88bf844a1a73d2386c38d1.

Issue #4600, opened 2015-07-30
Print a message if `getaddrinfo` is slow on startup


danielmewes:
On server startup, we call `get_local_ips` which in turn calls `getaddrinfo` for a few host names.

Under some circumstances, this step can take a while. It is done before we print any messages to the log or console, so this can be confusing. @deontologician ran into a case where the server was taking 15-30 seconds to come up, and we didn't know what was going on at first.

It would be nice if we could detect that the process is taking a while, and print a message about what's going on in that case.

This probably doesn't come up much in normal configurations.
In our case it happened because the OS X machine just had its hostname changed. After a reboot things were fine again. Hence pretty low priority...

rmg:
This also comes up when offline, as I discovered during a power outage. My laptop still runs, but WiFi is down, so my rethink docker container went from 2s startup to 24s startup. Fun to debug with access to the Internet only through a 3g tablet. But I did manage to come to the same conclusion as this issue through code inspection on github before landing here!

pdonorio:
I can confirm the problem offline, around ~30s (kinda frozen) if wifi is off or unreachable. Would it be possible to add an option on the command line to skip the operation?

danielmewes:
@pdonorio That sounds reasonable I think. I'm not 100% sure what we currently do the address lookup for. It might be important somehow, but most likely we can do without it.

pdonorio:
Ah great! Thanks! 
Make sure that is really nothing important, but if you want my guess 
it probably checks rethinkdb installed binary version against your website to tell the user when he should upgrade.

danielmewes:
@pdonorio I don't think the version check is the issue. The version check runs in the background and will not block a RethinkDB server from starting up (you can start with `--no-version-check` to verify this). The `getaddrinfo` that's blocking is one that's related to figuring out the interfaces to bind on I think.

pdonorio:
I can confirm that with wifi disabled and `rethinkdb --no-update-check` the issue is still there.

Now that I recall, I had a similar app hanging with `socket.gethostbyaddr` in python, of course when offline. Since there were no options to use a shorter timeout it ended up with skipping the call when you know you are disconnected.

nenuko:
Hi! I've removed Google Public DNS on Network preferences and it worked

irae:
Thanks @nenuko! Based on your finding, although I was not using Google DNS, it got me trying stuff. Finally, this is what I did to solve it:

```
sudo hostname [new name]
[...]

PR #6587, opened 2018-01-05
Add the startup flag --no-dns-resolution to fix issue #4600.

jtc3161:

### Description
This pull request adds a startup flag to tell rethinkdb to skip resolving hostnames via the dns server. This is being submitted as a possible fix to Issue 4600 where a misconfigured DNS server causes a startup delay for the database.

### DIFF
--- a/src/arch/address.cc
+++ b/src/arch/address.cc
@@ -149,14 +149,17 @@ std::set<ip_address_t> hostname_to_ips(const std::string &host) {
 }
 
 std::set<ip_address_t> get_local_ips(const std::set<ip_address_t> &filter,
-                                     local_ip_filter_t filter_type) {
+                                     local_ip_filter_t filter_type, 
+                                     bool ifaddr_only) {
     std::set<ip_address_t> all_ips;
     std::set<ip_address_t> filtered_ips;
 
-    try {
-        all_ips = hostname_to_ips(str_gethostname());
-    } catch (const host_lookup_exc_t &ex) {
-        // Continue on, this probably means there's no DNS entry for this host
+    if (!ifaddr_only) {
+        try {
+            all_ips = hostname_to_ips(str_gethostname());
+        } catch (const host_lookup_exc_t &ex) {
+            // Continue on, this probably means there's no DNS entry for this host
+        }
     }
 
 #ifdef _WIN32
--- a/src/arch/address.hpp
+++ b/src/arch/address.hpp
@@ -109,7 +109,8 @@ enum class local_ip_filter_t {
 };
 
 std::set<ip_address_t> get_local_ips(const std::set<ip_address_t> &filter,
-                                     local_ip_filter_t filter_type);
+                                     local_ip_filter_t filter_type, 
+                                     bool ifaddr_only);
 
 class port_t {
 public:
--- a/src/clustering/administration/main/command_line.cc
+++ b/src/clustering/administration/main/command_line.cc
@@ -593,7 +593,8 @@ private:
 std::set<ip_address_t> get_local_addresses(
     const std::vector<std::string> &specific_options,
     const std::vector<std::string> &default_options,
-    local_ip_filter_t filter_type) {
+    local_ip_filter_t filter_type, 
+    bool ifaddr_only) {
     std::set<ip_address_t> set_filter;
 
     const std::vector<std::string> &bind_options =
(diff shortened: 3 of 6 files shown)


srh:
I can confirm it works!  Thank you.

I'll look at the code, figure out what this is all about and what should be the default behavior, and merge this into next and v2.4.x this weekend.

srh:
It looks like the practice of calling `getaddrinfo` on the result of `gethostname` was introduced with commit 9fa17c5fff16de8f27362f8406ee22658ba09f52.

Also, I'll mention issue 4600 **here** -> #4600 <- so that it has a backlink to this PR.

I'm probably going to add a tweak to make the testing system pass this flag, because I've been annoyed in the past by invocations of `./test/run` just resulting in a bunch of timed out tests.

srh:
It looks like we do a few things in `get_local_ips`:
1. We call `getaddrinfo` on the result of `gethostname` to get some IP addresses to bind to.  And we filter out the loopback IP's we see from that.
2. We use `getifaddrs` to to get some IP addresses to bind to, which we add to the set we're building.  This provides us with the loopback IP's.
3. Then we *filter* that set of addresses with a filter provided to the `--bind` argument.

After that returns we:
4. "`// Make sure that all specified addresses were found`".  We then go through the filter, and check that an interface was found for every address we can bind to in that filter.
5. Then, finally, if the user passed a `--bind all`, we ignore the results of that work entirely and return an empty set of IP addresses, which gets treated specially by `linux_nonthrowing_tcp_listener_t`'s constructor (see "`// If no addresses were supplied, default to 'any'`").

----
Here's what I think we should do instead:

- If the user passes `--bind all`, we will just bind on '0.0.0.0' like we end up doing.
- If the user passes bind options with certain IP addresses, we'll just try binding on those IP addresses and see if the `bind()` call succeeds.  We'll do this *instead* of item 3 above where we read the set of interfaces and filter through them.
- If we want to bind to loopback but nothing else (e.g. if the user passes no `--bind` options), then we'll use `getifaddrs` to get the loopback IP's.  Looking at commits f1bade75d750345811766bd47c2bdc4e7942a594 and 0116bb768a94198fc1ddf7a9d389a31699a38312 in the history of `src/arch/address.cc`, I'm going to conclude that getting loopback IP's is at least conservative, non-breaking behavior.

So, what would be different?  Only the case when the user passes certain IP addresses for us to bind to.  Instead of using getaddrinfo and getifaddrs to get interface IP's, then filtering them, and then binding to the IP addresses the user supplied anyway (and checking the error anyway, as we do), we'll just bind to the addresses straight away.

srh:
Just noticed this from commit 3d2e9ae1fa754b2adb61955401c32c66c9cfd5c8:

```
    // Ignore loopback addresses - those will be returned by getifaddrs, and
    // getaddrinfo is not so trustworthy.
    // See https://github.com/rethinkdb/rethinkdb/issues/2405
    for (auto it = all_ips.begin(); it != all_ips.end();) {
        auto curr = it++;
        if (!curr->is_loopback()) {
            all_ips.erase(curr);
        }
    }
```

The `curr->is_loopback()` test is mistakenly negated.

srh:
It turns out my plan above has one flaw:  We still use `get_local_ips` to compute our own canonical addresses in the rpc/ directory.  I haven't looked at that code *too* closely but it does mean one use of get_local_ips will stick around.  I wonder why we compute our own canonical addresses, instead of just having the node receiving a connection use the IP address that it was told connected to it.  Well, maybe because then it might start sharing that peer info with other peers.

I don't see any intelligent design behind using DNS resolution to find our network addresses, so I'm going to remove it and solely use getifaddrs instead.  So my plan is now:

----

To remove the `--no-dns-resolution` flag and just drop DNS resolution entirely (in other words, make the code behave as if that flag was turned on).  The exception is Windows, where DNS resolution of our own hostname will remain turned on, for now.

srh:
Oh, I should add:  Because `curr->is_loopback()` was negated, we actually discarded all the non-loopback addresses returned by DNS resolution, and kept the loopback addresses.  So our DNS resolution was all for nothing.  I'm confident this won't break things.

srh:
These changes have been merged with #6588.  Note that I dropped the `--no-dns-resolution` flag entirely, and made the code act as if the flag were always turned on.

PR #6593, opened 2018-01-14
Remove unmaintained Scala and PHP drivers

ShalokShalom:

### Description
This commit removes the Scala and PHP drivers, since they are unmaintained since nearly 2 years.


lbguilherme:
Even if they didn't get any commit for 2 years, I wouldn't call them unmainteined because:

1. They support 2.3.x, which is *the latest release* of RethinkDB
2. They work
3. There are no better replacement on that language (I didn't actually search, but if there is, then better replace, not remove)

srh:
RethinkDB is stable enough that they don't need to be maintained.  I think we shouldn't remove any drivers from that list unless their authors request it.

ShalokShalom:
ok

PR #6610, opened 2018-02-07
Modify error info when creating index failed

Shawn-ye:

### Description
When creating a secondary index failed, the error log would be something like this:
> The server(s) hosting table abc.in are currently unreachable. The secondary index was not created. If youdo not expect the server(s) to recover, you can use emergency_repair to restore availability of the table.

Where actually `abc` is the name of the db and `in` is the name of the index, but the log starts with 'server(s) hosting table'. This really confuses users and I would think that `in` is the name of the table, not the index.


lbguilherme:
merge?

gabor-boros:
@srh What do you think about this? can we merge?

srh:
I think this won't compile, because c_str() is already part of the macro.  I'll get it compiling then we'll see.

srh:
Amended the commit and merged to next as of 6b1f8238272.

PR #6654, opened 2018-04-17
Update net_tornado.py

krull:

### Description
Compatibility for `tornado 5`, remove line `117`, and `141`, per mentioned in the [release notes of 5.0](http://www.tornadoweb.org/en/stable/releases/v5.0.0.html#backwards-compatibility-notes)

> `io_loop` arguments to many Tornado functions have been removed. Use `IOLoop.current()` instead of passing `IOLoop` objects explicitly.

### DIFF
--- a/drivers/python/rethinkdb/tornado_net/net_tornado.py
+++ b/drivers/python/rethinkdb/tornado_net/net_tornado.py
@@ -114,7 +114,6 @@ class ConnectionInstance(object):
             self._stream = yield with_absolute_timeout(
                 deadline,
                 stream_future,
-                io_loop=self._io_loop,
                 quiet_exceptions=(iostream.StreamClosedError))
         except Exception as err:
             raise ReqlDriverError('Could not connect to %s:%s. Error: %s' %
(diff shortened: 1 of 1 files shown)


srh:
What happens to people using Tornado 4?  Does this break for them?  Is there a way to support both?

srh:
One possibility:  Check the Tornado version at runtime.  Another possibility:  Fork the library.

krull:
hi @srh

> What happens to people using Tornado 4? Does this break for them? Is there a way to support both?

I think it is broken in this form for tornado 4 users. So yes.

> One possibility: Check the Tornado version at runtime. Another possibility: Fork the library. Another: Nobody uses earlier versions of Tornado and we should just support the latest. Meanwhile, the 2.3.6 versions of the driver do not...

My suggestion is to check at runtime, one version back, with a nice Warning of outdated version used. This way you don't get stale installs, with forward moving mindsets. YMMV

BR

krull

codeskyblue:
Any  progress?

gabor-boros:
FYI: @codeskyblue 
CC: @grandquista 
This issue should be moved to the `rethinkdb-python` repo.

gabor-boros:
Due to the issue resolved in the python driver I’m closing this issue

PR #6667, opened 2018-05-21
Use patch -p1 instead of patch -fp1

### LINKED ISSUE #6666: `patch -fp1` doesn't work on Alpine Linux
RethinkDB 2.3.5 would build on Alpine Linux, but then [this change](https://github.com/rethinkdb/rethinkdb/commit/41e11c740b32493deefd668ab343fabb15d4b6b4#diff-2f99a770a02afe004be6e92b2cc92161R126) added `patch -fp1` and that isn't a valid command on Alpine Linux, so it fails to build.

srh:

### Description
Fixes #6666.

### DIFF
--- a/mk/support/pkg/v8.sh
+++ b/mk/support/pkg/v8.sh
@@ -6,13 +6,13 @@ src_url=http://commondatastorage.googleapis.com/chromium-browser-official/v8-${v
 
 pkg_install-include () {
     pkg_copy_src_to_build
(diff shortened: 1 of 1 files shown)


srh:
next 98fd78409095a25b04dcaf0eadd6719f381a7d7e
v2.4.x d602ca096da20c136a90d0eb6fa0ad0622a87d86
v2.3.x 700a7b9a96ede2c89718b2551c1e80cb22c253fd

PR #6686, opened 2018-07-02
Support Python 3.7

Ars2014:
- [✔] I have read and agreed to the RethinkDB Contributor License Agreement http://rethinkdb.com/community/cla/

### Description
Change 'asyncio.async' to 'asyncio.ensure_future' since the first was removed in Python 3.7.


srh:
This will break stuff before 3.4.4, right?

Ars2014:
@srh, Yep, but I don't think this will be such a problem.

PR #6709, opened 2019-02-07
Sam/24x/fixups

srh:

### Description
This adds a few fixups to the v2.4.x branch.

- disables a test that, annoyingly, is failing all the time
- turns off debug-mode throws() checks, which are an obsolete C++ feature in C++17.

Most importantly:
- Re-introduces proper use of the superblock write semaphore!!!

This bug has been around since version 2.0.0, but it was not present in 1.16.

The consequences of this change is, if there is a big queue of writers lined up to write on a given table, readers are allowed to jump the line.

### DIFF
--- a/src/btree/reql_specific.cc
+++ b/src/btree/reql_specific.cc
@@ -59,6 +59,7 @@ real_superblock_t::real_superblock_t(
 
 void real_superblock_t::release() {
     sb_buf_.reset_buf_lock();
+    write_semaphore_acq_.reset();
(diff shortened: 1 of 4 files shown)


srh:
Cherry-picked two of the changes to next, but not the CrashedJob test disabling, as of 09708889394bd632a13105061e1ed1c9f5e61e2b and 62cbaa0021487a2bece7f15556cc42abbdd4d332.

PR #6737, opened 2019-04-22
Add libssl-dev

gabor-boros:

### Description
The `libssl-dev` package is not listed for install but needed for the builds.


srh:
Looks good to me.  Should also be in v2.4.x.  Will merge when I get to the right PC.

PR #6757, opened 2019-08-08
Update CONTRIBUTING.md

gabor-boros:
### Description
This PR is opened because we need to match contribution guidelines.

My email address is **temporary** until we gain access over emailing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ a project may be further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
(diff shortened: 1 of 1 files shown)


srh:
> This PR is opened because we need to match contribution guidelines.

What?

gabor-boros:
@srh sorry it wasn't really explicit or meaningful.

So we needed this change because the CII guidelines required us to have a proper contribution guideline. This guideline was suggested by them. The previous contribution guide contained things which - I think - is not needed like signing the contribution agreement and so on.

We had to provide an email address in the contribution guide, and because we have no common email address yet, I added by own to ensure we covered the checklist on their side.

Issue #5537, opened 2016-03-17
We should allow `-`s in db names, table names etc.


mlucy:
It seems like the rest of the world is moving to that in place of `_`, and it's pretty annoying.

(@danielmewes, is there any way we could schedule this for 2.4?)

felixis-1:
+1.

Just to add, we have a number of applications where we store some metadata into RethinkDB and create databases and tables by naming them as the auto generated ID of these metadata records. This use case occurs too often for us. And just because of this restriction on `-` character, we need to do string manipulation all over the place. This will be a very nice modification for us.

danielmewes:
Yeah, good idea.

VeXocide:
In favour.

oliverjanik:
Yes please,


My use case is generating table names on the fly based on sanitized input. That input however does contain `-` character often which is very common in urls and slugs.

mlucy:
Actually, as long as we're here and changing this, why do we have any restrictions on db/table names?  Would people object to drastically expanding the set of allowed characters rather than just supporting `-`?

marshall007:
I'm suspicious @mlucy is only saying that because he wants to use emoji for the system table names.

danielmewes:
We should make sure that all db and table names are valid filenames, because our backup tools rely on that (see https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words ). @VeXocide also mentioned that we should disallow quotes and control characters.

deontologician:
So all of unicode, minus `"'<>{}()[]\/*?%` ?

VeXocide:
☃

danielmewes:
I think maybe we should just add `-` and call it a day.

oliverjanik:
Yeah, just whitelist few more commonly used chars that have low chance of colliding with filename restrictions in any OS and job done.

mlucy:
It would be really annoying to have to go back and add more legal characters later.  I've personally wanted "."s and "/"s in table names before (when I wanted a table per website and a table per nested directory, respectively).

Personally I would vote for allowing any db/table name at all and having our backup tool do escaping.

If people think that's overkill, my second vote would be that anything you can type on a qwerty keyboard should be legal (and our backup tool should still do escaping -- not being able to have ":" or "/" in db/table names would be unfortunate).

If adding escaping to our backup tool is too much work, I would vote for just adding support for "-" and opening another issue to support any db/table name after the backup tool is changed.

[...]

PR #6843, opened 2020-03-03
Allow dash in DB and table names when using '-e' option in rethinkdb-dump

hchac:

### Description
Hello,

it seems from https://github.com/rethinkdb/rethinkdb/issues/5537 that '-' is allowed in DB and table names. However, the `-e` option when using `rethinkdb-dump` breaks when specifying a DB or table with '-' in its name due to this regex.

Opening this pull request to hopefully fix that.

### DIFF
--- a/drivers/python/rethinkdb/utils_common.py
+++ b/drivers/python/rethinkdb/utils_common.py
@@ -112,7 +112,7 @@ def check_minimum_version(options, minimum_version='1.6'):
 
 
 DbTable = collections.namedtuple('DbTable', ['db', 'table'])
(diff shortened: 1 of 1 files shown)


gabor-boros:
@hchac Thank you for your PR 🙏We extracted the python driver to rethinkdb/rethinkdb-python, but not deleted the driver folder from here (yet). If you wouldn't like to do the same change there, I'm happy to do that.

gabor-boros:
Closing since the related PR is merged and the change will be release on Tuesday (19th of May)

Issue #6863, opened 2020-05-07
Rethinkdb crash - Changefeed Guarantee failed: [inserted]


tommilligan:
We've recently seen a Rethinkdb crash that recurs and we're not sure what the root cause is.

The most recent similar ticket is https://github.com/rethinkdb/rethinkdb/issues/6695, which was closed a couple of months ago, and included in 2.4.0. Our crash is consistent with the behaviour detailed there, but is **not** a limit changefeed. It does appear to crash inconsistently, suggesting (similar to several other issues) that it is affected by other load on the cluster.

**Version**: 2.3.7
**OS**: Ubuntu 18.04

Trback:

```log
Version: rethinkdb 2.3.7~0bionic (CLANG 6.0.0 (tags/RELEASE_600/final))
error: Error in src/rdb_protocol/changefeed.cc at line 1164:
error: Backtrace:
error: Guarantee failed: [inserted] 
error: Exiting.ace
       10 [0xd0de24]: coro_t::run() at 0xd0de24 (rethinkdb)
[...]

PR #6710, opened 2019-02-07
Fix limit change feed secondary index bug

srh:
…ffecting changefeed limit queries

- [x] I have read and agreed to the RethinkDB Contributor License Agreement http://rethinkdb.com/community/cla/

### Description
Fixes a bug where a so-called "superblock" gets released too early.  Affects behavior of limit change feeds defined on secondary indexes.

This might fix issue #6695.

PR #6865, opened 2020-05-09
Backport 'Treat release_superblock properly in do_read' to 2.3.x

tommilligan:

### Description

The PR backports the fix described in https://github.com/rethinkdb/rethinkdb/pull/6710 to the 2.3.x branch, to attempt mitigation of https://github.com/rethinkdb/rethinkdb/issues/6863

### DIFF
--- a/src/rdb_protocol/btree_store.cc
+++ b/src/rdb_protocol/btree_store.cc
@@ -974,6 +974,7 @@ MUST_USE bool store_t::acquire_sindex_superblock_for_read(
         const sindex_name_t &name,
         const std::string &table_name,
         real_superblock_t *superblock,
+        release_superblock_t release_superblock,
         scoped_ptr_t<sindex_superblock_t> *sindex_sb_out,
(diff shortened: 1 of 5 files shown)


gabor-boros:
@tommilligan why would you like to backport that change to 2.3.x instead of using 2.4?

tommilligan:
It's a fair point - we have investigated upgrading to 2.4, but as it requires data file migrations, it wouldn't be a reversible upgrade for us. It's also just not a high priority for us at the moment. As there has been considerable work on the `next` branch since the 2.4 release, we're thinking of waiting for a point release unless we have a really good reason to upgrade.

FWIW we've been running this patch in production since I opened the PR, and haven't seen any issues (although this is a non-reproducible bug, so :shrug:).

gabor-boros:
@srh if you agree we could merge this

srh:
Please never hesitate to ping me multiple times if I don't respond.

Merging.

Issue #1740, opened 2013-12-05
Move drivers out of the main tree


mmalecki:
I noticed that all the drivers are kept in the main tree. This is problematic for developers who want to contribute to them or even just check out the source code locally.

I'd be happy to help with moving the node driver out of the tree.

knownasilya:
Would love them moved out as well, makes it that much easier to contribute, and find the code. This has a lot to do with productivity, commit history, issues, PRs, etc. It's all merged together and that's a HUGE pain! Like I was just looking through the commit history, trying to find where Promise support was added to the Javascript driver, and I couldn't find anything.. almost 30k commits.

deontologician:
@knownasilya If you're looking for changes that only affect a single directory, you can specify it to git log:

```
$ git log --name-status --no-merges drivers/javascript
```

AtnNn:
For the specific problem of finding commits and tracking changes, you can use the following `git` commands:

`git blame` to find the commit that introduced a change:

```
$ git blame -- drivers/javascript/net.coffee | grep -i bluebird
02b70c7f drivers/javascript/net.coffee     (neumino       2014-03-31 17:18:38 -0700  15) Promise = require('bluebird')
```

02b70c7f (Add promises to the JS driver) is the commit you are looking for.

`git log` to show all changes in a specific subfolder:

```
$ git log --graph --oneline --decorate -- drivers/javascript/
* 4d2fd72 (tag: v1.15.0) Bump driver version numbers to 1.15.0.0
|  drivers/javascript/package.json |    2 +-
|  1 file changed, 1 insertion(+), 1 deletion(-)
* 29ab2e5 (origin/michel_3087_stack_traces_2) Create a stack trace only if Error.captureStackTrace is available
|  drivers/javascript/errors.coffee |    6 ++++--
|  1 file changed, 4 insertions(+), 2 deletions(-)
*   dc0cff0 Merge branch 'michel_3054_backtraces' into v1.15.x
|\  
| * 1b48d71 Backtrace should prefix geo terms
| |  drivers/javascript/ast.coffee |   10 +++++-----
| |  1 file changed, 5 insertions(+), 5 deletions(-)
* | 4fb062f Capture stack trace
| |  drivers/javascript/errors.coffee |    2 ++
| |  1 file changed, 2 insertions(+)
* | 23c52ce (origin/gchpaco_2185) Fix ReQL tests to reflect new style for batch operations.
|/  
[...]

PR #6870, opened 2020-05-17
Extract drivers

gabor-boros:
##  Background

Right now, the Python, Java, Javascript and Ruby drivers are living outside of this repository. We could remove the drivers folder and do a little cleanup. Although it would be easy to remove the 
drivers folder, we need to count with and resolve the following "issues":

* Rql tests are using the builtin python driver
* Admin UI using the builtin javascript driver

## Description

This PR removes all the drivers from the main repository and make sure we can still build the admin ui with any version of the javascript driver. As a "side effect" we remove bluebird dependency and add the javascript driver as an external dependency.

Since the JavaScript driver got [extracted](https://github.com/rethinkdb/rethinkdb-javascript) too (finally), we had to solve the Admin UI build issues. To mitigate the issue, we are replacing bluebird external dependency by RethinkDB's latest JavaScript driver.

_Note: During the extraction of the python driver (some years ago) we were not aware the rql tests, so this PR fixes the compatibility issues with the extracted driver too. Right now, all the python scripts in the `rql_test` directory were migrated to fully support Python 3 and the code was formatted. Python 3 support for other Python files in the repo will come later since that would blow up the scope of this PR._

## TODO before being ready for the review

- [x] Remove drivers folder
- [x] Remove related Make targets
- [x] Fix importing builtin drivers in rql tests and helpers
- [x] During build time we should prepare the requirements for the Admin UI

This PR would resolve  https://github.com/rethinkdb/rethinkdb/issues/1740 (moving to rethinkdash is out of scope right now).


gabor-boros:
@srh can we merge this?

Issue #6822, opened 2020-01-23
Re-introduce webui update checker, point it at download.rethinkdb.com


srh:
See discussion in #6791.

Basically, having a static `latest_version.txt` at download.rethinkdb.com should be super easy and allow us to avoid having to maintain update.rethinkdb.com.

This is the one other fix I'd like to get into 2.4.1.

PR #6879, opened 2020-05-24
Re introduce update checker

gabor-boros:

### Description
resolves https://github.com/rethinkdb/rethinkdb/issues/6822

Re-introducing the update checker pointing to the new download server which will be released on 29th of May (to avoid changing the url).

Screenshot 2020-05-24 at 9 28 59

_note: The real latest version is still 2.4.0, just manipulated the latest version on the new download server temporarily for the demo._


sholladay:
Thank you. I always thought this was a good way to discover an update, where I'm in the process of actually using the database.

gabor-boros:
@srh can we merge this?

PR #6911, opened 2020-08-18
Run ruby tests in circle ci

tinco:
Implement some fixes to automatically run Ruby tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ commands:
     steps:
       - run: |
           << parameters.pre_hook >>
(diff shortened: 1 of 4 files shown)


tinco:
Why is it not running circle ci on this pull request?

srh:
I've merged this into `next`.  Generally the plan is to make `v2.4.x` the main repo branch, then consider what in `next` to merge into it.  Because this change is built atop other changes in `next` branch (it has conflicts if rebased onto v2.4.x), I'm merging this into `next` (because the changes "look good") and the changes' getting added to `v2.4.x` will be downstream of that, if it happens.

PR #6912, opened 2020-08-21
fix environment syntax

tinco:
This pull request is an attempt to get all the tests to work in continuous integration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,9 +60,12 @@ commands:
       - run: |
           curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
           python get-pip.py
-          python -m pip install -r /tmp/rethinkdb-python/requirements.txt
           python3.7 get-pip.py
-          python3.7 -m pip install -r /tmp/rethinkdb-python/requirements.txt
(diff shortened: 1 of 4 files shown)


tinco:
@srh any idea why the integration test would be failing on the artificial tables test with the message:

```
/root/project/test/interface/../rql_test/test-runner:1895: UserWarning: Worker failure: Unable to insert user permissions ('test_user',): {u'first_error': u'The `connect` permission is only valid at the global scope.', u'errors': 1, u'deleted': 0, u'unchanged': 0, u'skipped': 0, u'replaced': 0, u'inserted': 0}
```

I don't really know what artificial tables are supposed to be or why it's inserting user permissions.

srh:
The artificial tables are the system tables in the system database `rethinkdb`, such as `r.db('rethinkdb').table('users')` and `r.db('rethinkdb').table('permissions')`.  That name gets used a lot in the code and tests.

srh:
Closing.  Why now: Because it targets `next` and `v2.4.x` will become the main branch of this repo.  Also it's not in a state plausibly ready to be merged to `next`.

I have made issue #7048 to look at this later.

PR #6934, opened 2020-12-26
fix: patch for rethinkdb 2.4.1 osx build

chenrui333:
👋  adding a patch for osx builds (see https://github.com/Homebrew/homebrew-core/pull/59643 and https://github.com/Homebrew/homebrew-core/pull/67782)

### DIFF
--- a/src/rdb_protocol/datum.cc
+++ b/src/rdb_protocol/datum.cc
@@ -12,7 +12,7 @@
 #include <iterator>
 
 #include "errors.hpp"
-#include <boost/detail/endian.hpp>
+#include <boost/predef/other/endian.h>
 
 #include "arch/runtime/coroutines.hpp"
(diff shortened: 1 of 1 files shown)


srh:
Thank you.

It looks like it'll work without updating the vendored boost version.

That's a good thing, because the date type is based on the boost library, and I'm not keen on changing how that gets built (though I'm considering that probably we should *always* vendor it for that reason).

Anyway, this patch seems like it'll drop in and build.  I'll test it out first and then merge it.

chenrui333:
Thanks @srh!

stakach:
any chance of movement on this one @srh ?

srh:
@stakach Yes.

Also, cherry-picked into v2.4.x with
d7e2706bf28bda2ef88666819173acd249a8a96e

PR #6949, opened 2021-02-23
Add AARCH64 Flag in Architecture Clause

odidev:
1) Mapped aarch64 with arm64 architecture in **mk/support/pkg/v8.sh**
2) Added arm64 and aarch64 flag in the declaration on minframe initialization and register definition in **src/arch/runtime/context_switching.cc**

Signed-off-by: odidev <odidev@puresoftware.com>

### DIFF
--- a/mk/support/pkg/v8.sh
+++ b/mk/support/pkg/v8.sh
@@ -136,7 +136,7 @@ pkg_install () {
         i?86)    arch=ia32 ;;
         x86_64)  arch=x64 ;;
         arm64)   arch=arm64; arch_gypflags=$arm_gypflags ;;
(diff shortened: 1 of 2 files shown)


AGSaidi:
I'd like to see Arm64 support!

srh:
Thank you for the fix.

In v2.4.x branch, an equivalent fix was applied in 01683d3e92a3280183235df6137825dc93fc7305 in context_switching.cc.

In v2.4.x, we have dropped v8 and replaced it with QuickJS.

We are going to abandon the `next` branch and are going to carefully apply its changes to v2.4.x (or some newly named branch) in the future.

So I am closing this PR.

Issue #6956, opened 2021-04-22
Memory bugs reported by ASan


zouyonghao:
system: Ubuntu 18.04 amd64
RethinkDB version: code on next branch (commit ec80a184092f5484326b9a581c21b227274099d3)

1. stack-use-after-scope

    ```
    Server got SIGINT from pid 0, uid 0; shutting down...
    =================================================================
    ==17041==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7f40ddefe880 at pc 0x000002b2a55f bp 0x7f40ddefe5b0 sp 0x7f40ddefe5a8
    WRITE of size 8 at 0x7f40ddefe880 thread T9
        #0 0x2b2a55e in epoll_event_queue_t::forget_resource(int, linux_event_callback_t*) /home/zyh/rethinkdb/src/arch/runtime/event_queue/epoll.cc:184:5
        #1 0x2ad0357 in linux_event_watcher_t::remask() /home/zyh/rethinkdb/src/arch/io/event_watcher.cc:204:54
        #2 0x2ad0aa7 in linux_event_watcher_t::stop_watching_for_errors() /home/zyh/rethinkdb/src/arch/io/event_watcher.cc:150:9
        #3 0x2ad093c in linux_event_watcher_t::~linux_event_watcher_t() /home/zyh/rethinkdb/src/arch/io/event_watcher.cc:144:5
        #4 0x2d2def6 in linux_event_fd_watcher_t::~linux_event_fd_watcher_t() /home/zyh/rethinkdb/src/containers/archive/socket_stream.cc:223:1
        #5 0x2d2e04f in linux_event_fd_watcher_t::~linux_event_fd_watcher_t() /home/zyh/rethinkdb/src/containers/archive/socket_stream.cc:221:55
        #6 0x2d0abbe in scoped_ptr_t<fd_watcher_t>::reset() /home/zyh/rethinkdb/./src/containers/scoped.hpp:73:9
        #7 0x2d0a8e8 in scoped_ptr_t<fd_watcher_t>::~scoped_ptr_t() /home/zyh/rethinkdb/./src/containers/scoped.hpp:35:9
        #8 0x2d2ee62 in socket_stream_t::~socket_stream_t() /home/zyh/rethinkdb/src/containers/archive/socket_stream.cc:247:1
[...]

PR #6957, opened 2021-04-22
try to fix memory leak

zouyonghao:
**Reason for the change**

for https://github.com/rethinkdb/rethinkdb/issues/6956

**Description**

use unique_ptr to deallocate the object

**Checklist**
- [x] I have read and agreed to the [RethinkDB Contributor License Agreement](http://rethinkdb.com/community/cla/)

### DIFF
--- a/src/random.cc
+++ b/src/random.cc
@@ -3,6 +3,7 @@
 
 #include "containers/scoped.hpp"
 #include "thread_local.hpp"
+#include <memory>
 
 rng_t::rng_t()
     : m_mt19937(std::random_device{}()) {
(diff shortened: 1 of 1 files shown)


src/random.cc
```diff
@@ -41,16 +42,19 @@ double rng_t::randdouble() {
...
-    if (TLS_get_rng() == nullptr) {
-        TLS_set_rng(new rng_t());
-    }
-    return TLS_get_rng();
+    // if (TLS_get_rng() == nullptr) {
+    //     TLS_set_rng(new rng_t());
+    // }
+    // return TLS_get_rng();
+    
+    static thread_local std::unique_ptr<rng_t> tls_rng(new rng_t());
```
srh, line 56: Well.  That looks great.  I'll check that this works on the oldest platforms we support, and if it does, we'll merge this and backport it to v2.4.x.

Thanks for the PR.

srh:
Thank you again for the PR.

I am closing in favor of #7010, which merges your change, and another, to `v2.4.x`.  We are not going to be using the `next` branch in the future, so this PR does not need to be merged to `next`.

If you are interested, see the changes in the second commit of #7010.  We need NOINLINE to stop the compiler from optimizing access to thread local storage in this circumstance:

```
auto x = TLS_get_foo();
on_thread_t th(other);
auto y = TLS_get_foo();
```

The compiler (or link time optimizer) does not know that `on_thread_t th(other)` will switch the coroutine to a different POSIX thread.  Thus it may assume that both calls to TLS_get_foo(), after inlining, both access a thread_local, which on_thread_t's constructor does not modify, thus they return the same value.  Using NOINLINE, we prevent this optimization.

srh:
I'll add, because `thread_local` is a C++11 feature,  I've decided to not worry about platforms which do not support it (which might be CentOS 6 with an old system gcc 4.8 compiler).

PR #6965, opened 2021-04-28
Fix link flags order for curl, openssl and zlib

Skywalker13:
The flags order is very important especially in case of static linking. This commit fixes the static link of rethinkdb binary by providing the flags in the appropriate order.

**Reason for the change**
Static linking is broken

**Description**
Linking needs appropriate flags order

**Checklist**
- [X] I have read and agreed to the [RethinkDB Contributor License Agreement](http://rethinkdb.com/community/cla/)


srh:
Thank you for the fix.  But I think static linking is working for me, e.g. with `./configure --fetch all --static all CXX=clang++` on Ubuntu 20.04.

What configuration do you have which gets fixed by this patch?

Skywalker13:
Hello

I build myself (no `--fetch all`) most dependencies (jemalloc zlib protobuf openssl libcurl) only in static and I configure by this way:  
image

When libs are installed in my sysroot only in static (without .so), it reveals the static linking problems.

In order to reproduce, I must prepare a script for you.. let me some days for that because I can't provide our whole toolchain just for this problem..

srh:
Closing because `v2.4.x` is to become the main branch of this repo (once I've cleared out the PR's targeting `next`).  This change is in v2.4.x already (via your other PR, as you've seen).

PR #6966, opened 2021-04-28
Fix link flags order for curl, openssl and zlib (v2.4)

Skywalker13:
Like https://github.com/rethinkdb/rethinkdb/pull/6965 but for v2.4 branch


srh:
Thank you for the changes.  This has been rebased onto v2.4.x in commit 3f653078d11d8601639f212e136318444890ea86.

PR #6978, opened 2021-06-26
Replace v8 with QuickJS

srh:
Replaces v8 with QuickJS.

Keeps the Duktape replacement in the git history.

The build system now git clones a vendored repo at https://github.com/srh/rethinkdb-vendored, and we should change that to https://github.com/rethinkdb/rethinkdb-vendored , provided that we do want to make the build system clone a vendored repo for the dependencies this way.

Also, we need to get this working on Windows.


srh:
After a make clean and rebuild, we already see a problem with this PR:

```
rm: cannot remove './build/vendored/quickjs': No such file or directory
cp: cannot create directory './build/vendored/quickjs': No such file or directory
make[1]: *** [src/build.mk:393: build/vendored/quickjs] Error 1
make[1]: *** Waiting for unfinished jobs....
```

srh:
> Also, we need to get this working on Windows.

I have some time freed up for this.  Reviewing the situation, this is the main (or only?) TODO item.

srh:
Well, that and addressing this:

```
VENDORED_REMOTE_REPO := https://github.com/srh/rethinkdb-vendored.git
```

srh:
I have merged this to v2.4.x.  I will make a separate issue for Windows support.

The reason I merged without Windows support was that Linux ARM builds were hitting v8 problems in #6999.

I still have not tested ARM but that is on the to-do.

PR #6987, opened 2021-09-16
Create SECURITY.md

zidingz:
A simple instruction for security researchers.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


srh:
Closing.  Why now?  Because `v2.4.x` is to become the main branch of the repo, and PR's targeting `next` have to be handled.

I'm not sure I want this change either.  I didn't sign up to become permanent security contact point for some RethinkDB release.  Probably the best bet is to make an issue, and if it's sensitive, ask who to contact, like before.

PR #7000, opened 2021-11-30
Fix build on Ubuntu 20.04 with clang++-9

zouyonghao:
**Reason for the change**
I got some compilation fails recently on Ubuntu 20.04 with clang++-9, this is a fix.

**Description**
Add a missing header and namespace.

**Checklist**
- [x] I have read and agreed to the [RethinkDB Contributor License Agreement](http://rethinkdb.com/community/cla/)

### DIFF
--- a/src/clustering/immediate_consistency/history.hpp
+++ b/src/clustering/immediate_consistency/history.hpp
@@ -4,6 +4,7 @@
 
 #include <map>
 #include <set>
+#include <stdexcept>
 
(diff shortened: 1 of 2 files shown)


srh:
First, thank you for the PR.

I am going to merge this, but note that the `next` branch is deprecated and v2.4.x should already have this fix (though it includes stddef.h instead of using std::size_t).

PR #7051, opened 2022-04-18
Add RDB_NO_BACKTRACE flag for musl, alpine users

srh:
Add CXXFLAGS=-DRDB_NO_BACKTRACE to your ./configure line, or your
environment when running make.

I have not (yet?) tested this on Alpine or with Musl libc.  (In the prior month or two, I got reliance on backtrace() being the one error on Alpine.)  This will make it possible or at least easier for musl users to use RethinkDB.

### DIFF
--- a/src/backtrace.cc
+++ b/src/backtrace.cc
@@ -332,7 +332,10 @@ void initialize_dbghelp() {
 #endif
 
 std::string backtrace_t::print_frames(bool use_addr2line) const {
-#ifdef _WIN32
+#if defined(RDB_NO_BACKTRACE)
+    (void)use_addr2line;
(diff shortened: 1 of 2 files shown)


srh:
Here's what it looks like (adding a `guarantee(false)` line in uuid_term_t and running an `r.uuid()` query to see the behavior):
```
error: Error in thread 16 in src/rdb_protocol/terms/uuid.cc at line 24:
error: Guarantee failed: [false] 
error: Backtrace:
error: Sun Apr 17 23:33:47 2022
       
       backtrace not supported by this binary
error: Exiting.
Trace/breakpoint trap (core dumped)
```

Issue #4647, opened 2015-08-08
Crash when starting service related to log directory set in conf


p-e-s-t-s:
I went from running rethinkdb on the command line to starting it as a service. I used the default conf except for changing the data and log locations. This crash happened after trying to (re)start the service.

The problem ended up being the log config change. I commented it back out to let rethinkdb use its default and then it started up just fine.

Probably an issue with permissions but @mglukhovsky asked me log the crash here.

```
cat /proc/version
Linux version 3.13.0-48-generic (buildd@orlo) (gcc version 4.8.2 (Ubuntu 4.8.2-19ubuntu1) ) #80-Ubuntu SMP Thu Mar 12 11:16:15 UTC 2015
```

```
rethinkdb -v
rethinkdb 2.0.4~0trusty (GCC 4.8.2)
```

```
sudo /etc/init.d/rethinkdb start
rethinkdb: instance1: The instance is not running (there is no pid file)
rethinkdb: instance1: will only listen on local network interfaces.
rethinkdb: instance1: To expose rethinkdb on the network, add the 'bind=all' option to /etc/rethinkdb/instances.d/instance1.conf
rethinkdb: instance1: Starting instance. (logging to `/var/log/rethinkdb')
Recursively removing directory /var/lib/rethinkdb/default
Previous message may not have been written to the log file (cannot open or find log file).
In recursion: removing file /var/lib/rethinkdb/default
Previous message may not have been written to the log file (cannot open or find log file).
Version: rethinkdb 2.0.4~0trusty (GCC 4.8.2)
error: Error in src/utils.cc at line 525:
Previous message may not have been written to the log file (cannot open or find log file).
error: Guarantee failed: [res == 0]  (errno 13 - Permission denied) Fatal error: failed to delete '/var/lib/rethinkdb/default'.
Previous message may not have been written to the log file (cannot open or find log file).
error: Backtrace:
Previous message may not have been written to the log file (cannot open or find log file).
error: Fri Aug  7 23:56:15 2015

       1: backtrace_t::backtrace_t() at ??:?
       2: format_backtrace(bool) at ??:?
       3: report_fatal_error(char const*, int, char const*, ...) at ??:?
       4: remove_directory_helper(char const*, stat const*, int, FTW*) at ??:?
[...]

PR #7057, opened 2022-05-02
refactor: avoid backtrace for user error

gabor-boros:

### Description

This PR resolves https://github.com/rethinkdb/rethinkdb/issues/4647 by using `fail_due_to_user_error` over `guarantee_err` and `guarantee_winerr` for `remove_directory_helper`.

### DIFF
--- a/src/paths.cc
+++ b/src/paths.cc
@@ -37,7 +37,7 @@ int remove_directory_helper(const char *path) {
     } else {
         res = DeleteFile(path);
     }
-    if(res != 0) {
+    if (res != 0) {
         fail_due_to_user_error("failed to remove: '%s': %s", path, winerr_string(GetLastError()).c_str());
     }
     return 0;
@@ -48,7 +48,7 @@ int remove_directory_helper(const char *path) {
 int remove_directory_helper(const char *path, UNUSED const struct stat *, UNUSED int, UNUSED struct FTW *) {
     logNTC("In recursion: removing file '%s'\n", path);
     int res = ::remove(path);
-    if(res != 0) {
+    if (res != 0) {
         fail_due_to_user_error("Fatal error: failed to delete '%s'.", path);
     }
     return 0;
--- a/src/paths.cc
+++ b/src/paths.cc
@@ -17,6 +17,7 @@
 
 #include "arch/io/disk.hpp"
 #include "clustering/administration/main/directory_lock.hpp"
+#include "errors.hpp"
 #include "logger.hpp"
 
 #ifdef _MSC_VER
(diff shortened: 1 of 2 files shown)


src/random.cc
```diff
@@ -3,7 +3,6 @@
 
 #include "containers/scoped.hpp"
 #include "thread_local.hpp"
-#include <memory>
```
srh, line 6: We don't want to remove this include -- we use std::unique_ptr in this file.  zouyonghao might have added it because it was necessary to compile, on some platform whose stdlib does not include <memory> via one of the other included headers.  So we should keep it.

Also, generally, it is preferable to have include lines in the same file for any library feature used in that file, even if a header already has it. One reason for this is that it makes it easier to remove include lines from frequently used header files (because there will later be fewer .cc files that need to be updated). That would reduce compilation time.  (This is a very unimportant, weak benefit.)

src/paths.cc
```diff
@@ -36,7 +37,9 @@ int remove_directory_helper(const char *path) {
     } else {
         res = DeleteFile(path);
     }
-    guarantee_winerr(res, "failed to remove: '%s': %s", path, winerr_string(GetLastError()).c_str());
+    if(res != 0) {
```
srh, line 40: Generally the coding style should be consistent with the rest of the codebase.

```suggestion
    if (res != 0) {
```

src/paths.cc
```diff
@@ -45,7 +48,9 @@ int remove_directory_helper(const char *path) {
 int remove_directory_helper(const char *path, UNUSED const struct stat *, UNUSED int, UNUSED struct FTW *) {
     logNTC("In recursion: removing file '%s'\n", path);
     int res = ::remove(path);
-    guarantee_err(res == 0, "Fatal error: failed to delete '%s'.", path);
+    if(res != 0) {
```
srh, line 51:
```suggestion
    if (res != 0) {
```

src/random.cc
```diff
@@ -3,7 +3,6 @@
 
 #include "containers/scoped.hpp"
```
srh, line 4: Having includes in a specific order is mostly superstitious, beyond just being organized, but if we're going to put `<memory>` back, this is where it would go (in this project).  As a system include it's at the top of the list, except for the cpp's own header file.

```suggestion
#include <memory>

#include "containers/scoped.hpp"
```

srh (changes requested):
Removing `#include <memory>` could be considered a defect, and the other changes requested are coding style.  Otherwise the changes look correct.

src/random.cc
gabor-boros: It turned out that compilation in Jammy fails without this include. I’m going to put this back.

srh (approved):

Issue #1279, opened 2013-08-12
Have rethinkdb --version print a bit more info.


srh:
RethinkDB, when starting up normally with a serve or porcelain command, prints the RethinkDB version and `uname -r` info.  Running `rethinkdb --version` just gets you the RethinkDB version, and it doesn't tell you what type of machine you're running on.
- `rethinkdb --version` should also print the `uname -r` info that tells what machine you're on.
- Also, the RethinkDB version should tell what the target architecture of the RethinkDB executable is, since that's not necessarily the same as the machine architecture.  You can run 32-bit executables on 64-bit systems.

PR #7060, opened 2022-05-05
feat: add machine arch and uname to rethinkdb version

gabor-boros:
### Description

This PR resolves https://github.com/rethinkdb/rethinkdb/issues/1279 by adding the `uname -r` info and target architecture to the `rethinkdb --version` output.

### Example output

```shell
root@8dc5774dc4a2:/workspace# build/debug/rethinkdb --version
rethinkdb 2.4.2-3-g0bad09-dirty (debug) (aarch64-linux-gnu) (GCC 9.4.0)
Linux 5.10.104-linuxkit aarch64
```

### Notes

Moved `std::string windows_version_string` and `std::string run_uname` before its first use for clarity. I can revert that if does not make sense.


gabor-boros:
_FTR: merging as we discussed with @srh that there are no concerns against this PR_

Issue #3353, opened 2014-11-19
ReQL proposal: string templating


marshall007:
String building is a bit of a pain right now, at least in the JS driver. This is loosely related to #3048.

``` js
r.table('foo').map({
  id: r.row('id'),
  bars: r.http(r.add(r.args(['http://example.com/foos/', r.row('id'), '/bars?type=', r.row('type')])))
})
```

It would be nice to have something like this:

``` js
r.table('foo').map({
  id: r.row('id'),
  bars: r.http(r.fmt('http://example.com/foos/{id}/bars?type={type}', r.row))
})
```

marshall007:
Here's a quick and dirty client-side implementation I've been using in case anyone else finds it useful:

```js
r.fmt = function (template, params) {
  var parts = template.split(/(\{\w+\})/g).map(function (part) {
    if (part.indexOf('{') === 0) {
      var name = part.slice(1, -1);
      return params(name).coerceTo('string');
    }
    return part;
  });
  
  return r.add(r.args(parts));
};
```

Example usage:

```js
r.fmt('http://example.com/foos/{id}/bars?type={type}', r.expr({
  id: 123,
  type: 'tag'
}))
// => "http://example.com/foos/123/bars?type=tag"
```

danielmewes:
@marshall007 That's a very neat work-around!

PR #7066, opened 2022-05-08
Gabor/add reql fmt

gabor-boros:

### Description

This PR resolves https://github.com/rethinkdb/rethinkdb/issues/3353 by adding the mentioned `r.fmt` command. The command formats the template string, passed as the first parameter, using the object provided as the second parameter.

### Related PRs

* https://github.com/rethinkdb/rethinkdb/pull/7065
* https://github.com/rethinkdb/docs/pull/1332

### Limitations

* The object must have `string` keys and `string` values.

### Questions

* Should we support `int` and `double` in the object? If yes, what would be the expected way of lexical casting of doubles to strings withouth adding extra trailing zeros or truncating the doubles?

### Leftovers

_The drivers shouldn't release `fmt` support until 2.5.0._

- [ ] Implement `fmt` in the Javascript driver
- [ ] Implement `fmt` in the Python driver (cc: @lsabi)
- [ ] Implement `fmt` in the Java driver (cc: @adriantodt)
- [ ] Implement `fmt` in the Ruby driver (cc: @tinco)

Other (maintained) drivers:

- [ ] Implement `fmt` in the Go driver (cc: @CMogilko)
- [ ] Implement `fmt` in the Typescript driver (cc: @atassis)


srh:
Is it going to be `r.fmt` or `r.format`?

I would expect it to support non-string values and use `r.coerceTo('string')`.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,55 @@ class match_term_t : public op_term_t {
...
+                while (pos + param_end < templ.size() && templ[pos + param_end] != '}') {
+                    param.push_back(templ[pos + param_end]);
+                    ++param_end;
+                }
+
+                const datum_string_t datum_param {datum_string_t(param)};
+                if (params.count(datum_param)) {
+                    formatted += params.at(datum_param).as_str().to_std();
+                } else {
+                    formatted += param;
```
srh, line 165: This is highly questionable fallback behavior.  It think we should throw an error if the field is absent.

If some kind of field-or-fallback-if-doesn't-exist logic is desired, it could be specified in the format syntax, or the user could r.merge the document into an object with default values.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,55 @@ class match_term_t : public op_term_t {
...
+        std::map <datum_string_t, datum_t> params{};
+        for (size_t i = 0; i < datum.obj_size(); ++i) {
+            const auto datum_pair{datum.get_pair(i)};
+            rcheck_target(&datum_pair.second, datum_pair.second.get_type() == datum_t::R_STR,
+                          base_exc_t::LOGIC,
+                          strprintf("Cannot call `%s` on objects of type `%s`.",
+                                    name(), datum_pair.second.get_type_name().c_str()));
+
+            params.insert(datum_pair);
+        }
```
srh, line 148: I don't see why we should copy out all the objects from the datum.  Instead, just use get_field on the datum.

Also, if we were to mandate all formatted fields be strings (or some other requirement), why should we mandate all the properties be strings and not just the ones we're formatting?  Generally, I think people might pass in documents that are not specially trimmed for the format command.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,55 @@ class match_term_t : public op_term_t {
...
+        std::string formatted{};
+        for (size_t pos = 0; pos < templ.size(); ++pos) {
+            if (templ[pos] == '{') {
+                size_t param_end {1};
+                std::string param {};
+
+                while (pos + param_end < templ.size() && templ[pos + param_end] != '}') {
+                    param.push_back(templ[pos + param_end]);
+                    ++param_end;
+                }
```
srh, line 159: A bug here is that if we reach the end of the format string without reaching a closing brace, we'd still try doing the formatting.

Also, if we have a string with extra curly braces, like `{abc{def}` it would try formatting with `abc{def`.  This is probably a user error.  You might want to carefully specify what characters are allowed inside a formatting specifier.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,55 @@ class match_term_t : public op_term_t {
...
+
+                const datum_string_t datum_param {datum_string_t(param)};
+                if (params.count(datum_param)) {
+                    formatted += params.at(datum_param).as_str().to_std();
+                } else {
+                    formatted += param;
+                }
+
+                pos += param_end;
```
srh, line 168: If `templ[pos + param_end]` is `'}'`, this means we would include `}` in the formatted string instead of skipping past it, so there is a bug here.

src/rdb_protocol/terms/string.cc
srh: Similarly, we probably want to prohibit stray `}` characters in the format string.  Or at least think about it.

srh (changes requested):
There are some bugs, and I think the treatment of format strings with unmatched curly braces is bad.

srh:
> Is it going to be r.fmt or r.format?

You wrote format or FORMAT the whole time in the code, which is why I'm asking.  I think r.format would be fine.

src/rdb_protocol/terms/string.cc
gabor-boros: That's a fair point. I'll refactor it to throw an error instead.

src/rdb_protocol/terms/string.cc
gabor-boros: Oh! I missed the get_field. Also, fair point to validate only necessary params.

If we would coerce the params to strings how would you do that to not loose/add extra precision for doubles?

src/rdb_protocol/terms/string.cc
gabor-boros: Thinking about it again, allowing missing closing brackets isn't a good idea. Will refactor this too.

src/rdb_protocol/terms/string.cc
gabor-boros: The } is not included in the param name as the while loop wont enter and increment the param_end variable. Or did I overlook something?

gabor-boros:
Thank you for the review, @srh . I'll go through the suggestions in the upcoming days. Also, I'll use the r.format version instead of r.fmt. That seems a bit more correct to me.

gabor-boros:
@srh except coercing all comments are addressed. The error message wording is probably not the best, but I'm opened for any suggestions.

**Regarding the coercing**

As I figured out, we could probably use the `minidriver` to evaluate the following, but I'm struggling how to properly do that. Do you have any suggestions?

```cpp
datum_t field = datum.get_field(datum_string_t{param_name});
r.expr(field).coerce_to("STRING")
```

gabor-boros:
Now, this is ready for review.

src/rdb_protocol/terms/string.cc
srh: Oh, I overlooked that the for loop itself increments with `++pos`.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,64 @@ class match_term_t : public op_term_t {
...
+                       base_exc_t::LOGIC,
+                       "Parameter tag must be closed");
+
+                minidriver_t r{src_term.bt()};
+                datum_t field{datum.get_field(datum_string_t{param_name})};
+                counted_t<const term_t> term = compile_term(
+                        compile_env,
+                        r.expr(field).coerce_to("STRING").root_term());
+
+                formatted += term->eval(env)->as_datum().as_str().to_std();
```
srh, line 172: This will copy the string in the result twice.  Instead, this line could use `formatted.append(result.data(), result.size())` to avoid the extra copy.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,64 @@ class match_term_t : public op_term_t {
...
+
+                    param_name.push_back(next_char);
+                    ++param_end_pos;
+                }
+
+                rcheck(pos + param_end_pos != template_length,
+                       base_exc_t::LOGIC,
+                       "Parameter tag must be closed");
+
+                minidriver_t r{src_term.bt()};
```
srh, line 166: The `raw_term_t` parameter to the class constructor gets passed up to the `runtime_term_t` base class, and that calls bt() and passes the backtrace id to the `bt_rcheckable_t` base class.  So this could call `backtrace()` instead of `src_term.bt()`.

We should get rid of `src_term` and use `backtrace()`.

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,64 @@ class match_term_t : public op_term_t {
...
+                datum_t field{datum.get_field(datum_string_t{param_name})};
```
srh, line 167: A question here, also with coerceTo('STRING'), is what behavior we want if both the format string has parse errors, and some early fields are missing.  Do we want to parse the whole format string first, and produce an error that there is an invalid format string?  Or is having errors on field lookups or coerceTo('STRING') behavior (if that can error) desirable?

src/rdb_protocol/terms/string.cc
gabor-boros: 👍 Good tip, fixing soon.

src/rdb_protocol/terms/string.cc
gabor-boros: Oh, I didn't realize it's available. Replacing the call soon.

src/rdb_protocol/terms/string.cc
gabor-boros: I think throwing an error on the case whichever happens first is desirable and expected.

If someone wants to format the content of _War and Peace_ from Tolstoy in a way that it adds a pre- and postfix to the book's content, and the first parameter is missing, we don't want to parse the whole text if we already know there is an error. So we raise an error at the time we realize a parameter is missing (, obviously, we cannot do much about the case if the suffix is missing). My thinking is the same for `coerceTo`, if we know we cannot coerce the field for some reason, we should let the caller know it as soon as possible.

srh: If the format string has parse errors, we would be saving work by not coercing some of the parameters before finding the parse error.

But I'm not concerned about the performance here.  What I'm concerned about is providing a useful error message to the user, and avoiding behavior changes in the future.

srh:
We might also want to reserve a character for use in format strings for use in specifying how the data is formatted.  C# uses `,` and `:`, while Rust uses `:`.  It might be that `:` is used in some field names.  Maybe there is a better character to use.

We might also want to reserve `.` to allow `field.subfield` syntax in the future.

src/rdb_protocol/terms/string.cc
srh: `to_std()` is what creates the copy here.  What I meant was something like

```
datum_string_t result = term->eval(env)->as_str();  // Or use ->as_datum(), then use a const datum_string_t& variable?
formatted.append(result.data(), result.size());
```

Note that `val_t` says to prefer `val_t::as_str()` with its comment, "prefer the forms below".  It's unclear to me why, but it catches the `datum_exc_t` and turns it into an `rfail`.

gabor-boros: @srh The copying issue is resolved by [45d0277](https://github.com/rethinkdb/rethinkdb/pull/7066/commits/45d02778807a2036b46433c0ced945e689d2d31a). Regarding the formatting and https://github.com/rethinkdb/rethinkdb/pull/7066#issuecomment-1133498128, what would you suggests?

In terms of keeping characters for formatting, I would personally prefer the colon approach or percent sign. These two characters are rarely used in document keys -- I hope -- and both are used by other languages like Python (`%`), Rust (`:`), Java (`%`), etc.

As of error messages, what are you suggesting? Should we parse the whole template string before trying to look for parameter replacements?

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,66 @@ class match_term_t : public op_term_t {
...
+                       "Parameter tag must be closed");
+
+                const datum_t &field{datum.get_field(datum_string_t{param_name})};
+
+                minidriver_t r{backtrace()};
+                counted_t<const term_t> term = compile_term(
+                        compile_env,
+                        r.expr(field).coerce_to("STRING").root_term());
+
+                const datum_string_t &result{term->eval(env)->as_str()};
```
srh, line 173: This should be not a const reference.  The existing code is not broken, but making it a const reference offers no advantage, because the return value of as_str() is returned by value and will be copy-elided or at worst move-constructed into the local.

Using a const reference for the variable type has a downside, too: it looks like there *might* be a bug.  What if eval(env) returned a temporary but `as_str()` returned a const reference pointing into its parameter?  That's generally plausible.  In that case, the temporary would get destructed and `result` would be pointing into freed memory.  If the return type of `as_str()` was changed in the future, then there would be a bug.  So this line increases the rate of bugs by some small number.

Generally, you don't want to have local variables be const references.  Unless there is something specific you want to accomplish.

```suggestion
                const datum_string_t result{term->eval(env)->as_str()};
```

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,66 @@ class match_term_t : public op_term_t {
...
+
+                    param_name.push_back(next_char);
+                    ++param_end_pos;
+                }
+
+                rcheck(pos + param_end_pos != template_length,
+                       base_exc_t::LOGIC,
+                       "Parameter tag must be closed");
+
+                const datum_t &field{datum.get_field(datum_string_t{param_name})};
```
srh, line 166: Also, get_field returns a datum_t, so having this be a `const datum_t &` reference doesn't avoid any copying.  The same comment about safety does not apply as strongly (because `datum` is not a temporary) but we still do not want a gratuitous reference.

```suggestion
                const datum_t field{datum.get_field(datum_string_t{param_name})};
```

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,66 @@ class match_term_t : public op_term_t {
...
+
+                minidriver_t r{backtrace()};
+                counted_t<const term_t> term = compile_term(
+                        compile_env,
+                        r.expr(field).coerce_to("STRING").root_term());
```
srh, line 171: Also, `field` *would* be copied an extra time here.  We could use `std::move(field)`.

But I double-checked and copying `datum_t` or a `datum_string_t` just increments a non-atomic refcount.  So there was no reason for me to be super concerned about one extra copy operation.

srh (changes requested):
Const reference nitpicking.

src/rdb_protocol/terms/string.cc
srh: Let me talk only about the error behavior issue here.

My question is, which error behavior is better?  Does it even matter?  The implementation might be fine.

srh:
> In terms of keeping characters for formatting, I would personally prefer the colon approach or percent sign. These two characters are rarely used in document keys -- I hope -- and both are used by other languages like Python (%), Rust (:), Java (%), etc.

What do you think nested fields might look like?

Maybe we should reserve an escape or quoting character that could be used for field names.

src/rdb_protocol/terms/string.cc
gabor-boros: > which error behavior is better? Does it even matter? 

Honestly speaking, I don't know which behaviour would be better. I think this is something that can be changed later on if the community prefers the other way around that we implement. I would leave it as it is now and change if someone has serious issues with this approach.

gabor-boros:
@srh 

> What do you think nested fields might look like?
> Maybe we should reserve an escape or quoting character that could be used for field names.

I had some time today to think about it, though I couldn't come up with a better idea. I think there are two questions:

* What character(s) do we want to reserve for later use for formatting?
* Do we want to support nested field access? If yes, how would we access the nested field?

> "What character(s) do we want to reserve for later use for formatting?"

The original proposal was to have a string replacement like this:

```js
// Replace the placeholder with string representation
r.format("Simple formatting: {x}", {"x": "})
```

To define formatting, we could use the `%`, `:`, or `,` sign, but using that could make the placeholder a bit unreadable:

```js
// Format the replacement as a floating point number v1
r.format("Temperature: {temp%f.2}", {temp: 32.123})

// Format the replacement as a floating point number v2
r.format("Temperature: {temp:f.2}", {temp: 32.123})

// Format the replacement as a floating point number v3
r.format("Temperature: {temp,f.2}", {temp: 32.123})
```

Out of the three the third variant, using `,`, maybe the most readable. Although I personally prefer `%` it is more important to keep the placeholder readable, so I'd say to reserve the `,` character.

> "Do we want to support nested field access? If yes, how would we access the nested field?"

If I understand correctly, this would be the usage example (assuming the subfield access character is `.`):

```js
// Replace the placeholder with string representation
r.format("Simple formatting: {x.y}", {x: {y: 1}})
```

Well, I would **not** support this. It could bring too many edge cases to the table. Consider the following example:

```js
r.format("Simple formatting: {x.y}", {"x.y": 1})
```

In this case, the key has the access character (`.`) in its name, furthermore, the value is not an object, but a number. Or even worse:

```js
r.format("Simple formatting: {x.y}", {"x.y": {y: 1}})
```

What would we render here? The key is `x.y`, but the value of `x.y` has a key `y` as well.

I would simply say that nested field's shouldn't be allowed as of now. If someone brings a proposal for solving all issues mentioned above, we could consider that, but I don't have a good proposal for solving the issue.

---

I went ahead and reserved the `,` character to reduce the number of iterations, though I'm opened for suggestions.

srh (approved):

src/rdb_protocol/terms/string.cc
```diff
@@ -125,6 +127,71 @@ class match_term_t : public op_term_t {
...
+                    char next_char{templ[pos + param_end_pos]};
+
+                    rcheck(next_char != '{',
+                           base_exc_t::LOGIC,
+                           "Nested template parameters are not allowed");
+
+                    // Reserve "," for later use to implement formatting
+                    rcheck(next_char != ':' && next_char != '\\',
+                           base_exc_t::LOGIC,
+                           strprintf("Formatting separator \"%c\" is not allowed in parameter name", next_char));
```
srh, line 161: Actually, one more thing.  The other error messages all use backtick characters for quoting variable names as well as user-supplied strings so it should be consistent with that.  Also, backslash isn't a formatting separator, so, I changed its message.

I looked and some rcheck lines had a trailing period character `.` on the error messages, but others did not.  So I guess that can go either way.

```suggestion
                    // Reserve ":" for later use to implement formatting, "\\" for escaping.
                    rcheck(next_char != ':',
                           base_exc_t::LOGIC,
                           strprintf("Formatting separator `%c` is not allowed in parameter name", next_char));
                    rcheck(next_char != '\\',
                           base_exc_t::LOGIC,
                           strprintf("Reserved escape character `%c` is not allowed in parameter name", next_char));

```

gabor-boros:
FTR: Before we merge the Python driver must be adjusted too.

gabor-boros:
The PR is rebased; I'll finish the PR after the next 2.4.x release.

gabor-boros:
@srh If I would want to fix the Python tests to work with the new driver implementation and Python 3, I would at least 4x the PRs size with somewhat irrelevant changes. How about the following?

- I move the second commit (674682082efa9067d683241fa72620e9aad9c562) to a separate PR where I update the repo to use Python 3
- I create a new __default__ branch from the current v2.4.x and call it `main`
- Then I merge this PR into `main` which will not affect the new 2.4.x release.

We already discussed the second one, but wanted to bring into your attention again as I've seen your plans to release a new 2.4.x version.

srh:
@gabor-boros I am hoping to get FreeBSD-related compilation fixes merged before forking the branches.

gabor-boros:
@srh 👌I'll wait for that to not interfere

srh:
Just reminding myself: after we break off the Python/test code changes and merge these, we will want to

- [ ] make driver changes
- [ ] update the JS driver in the admin UI in `old_admin` in a later PR

gabor-boros:
@srh Finally all the dependencies of this PR are done:

- https://github.com/rethinkdb/rethinkdb/pull/7065
- https://github.com/rethinkdb/docs/pull/1332
- https://github.com/rethinkdb/rethinkdb-javascript/pull/14
- https://github.com/rethinkdb/rethinkdb-python/pull/277
- https://github.com/rethinkdb/rethinkdb-ruby/pull/28

Two notes:

- the Python driver got a huge refactor (if you want to take a look, feel free to do so, but it is huge).
- the Java driver is not updated, on demand I can do that if nobody wants to (cc: @NotJustAnna maybe?)

At this point, I think we can merge this and all related PRs and decide what else to go into `v2.5.0`. What do you think?

srh:
Sure!

PR #7074, opened 2022-06-02
Fix compilation on windows

unxcepted:

### Description
This PR revives the windows support, which was dropped after the 2.3.6 release. I'm keeping this on a separate branch because some time is still needed to audit quickjspp code changes.

Since this PR, windows build additionally requires:
* Visual Studio 2022 with MSVC tools v141
* Strawberry Perl for OpenSSL build
* Premake5 for quickjspp build

Visual Studio 2015 and CMake installed globally are no longer required. Instructions in WINDOWS.md have been updated.

### List of changes
* Solution retargeted against VS 2022, MSVC v141
* Dependencies' windows build scripts fixed
* QuickJS replaced with c-smile/quickjspp
* GTest updated to the latest version supporting both MSVC v141 and traditional make compilation on Linux - v1.8.1
* Minor code changes (mostly adjusted includes)

### Issues
* quickjspp doesn't monitor stack size on windows - potential stack overflow in r.js causes a worker crash. As a consequence, `JSProc.InfiniteRecursionFunction` test fails.
* Boost complains about *unknown compiler* during compilation - this could be fixed by updating boost to >=1.65.1. It doesn't yield any problems during runtime though.

<hr/>
Special thanks to @MrBoombastic for an idea and help with testing.


src/paths.cc
```diff
@@ -37,7 +37,7 @@ int remove_directory_helper(const char *path) {
     } else {
         res = DeleteFile(path);
     }
-    if (res != 0) {
+    if (res == 0) {
```
srh, line 40: I had to double-check, and yes, DeleteFile and RemoveDirectory *both* return non-zero upon success.

src/rdb_protocol/geo/s2/base/port.h
```diff
@@ -11,7 +11,12 @@
 #include <limits.h>         // So we can set the bounds of our types
 #include <string.h>         // for memcpy()
 #include <stdlib.h>         // for free()
-#include <sys/socket.h>     // for MSG_NOSIGNAL
+//#include <sys/socket.h>     // for MSG_NOSIGNAL
```
srh, line 14: Commented line should be removed.

srh (approved):
It looks great!  I've left notes for myself for some petty fix-ups after merging.

PR #7076, opened 2022-06-16
Change to fix #6969

### LINKED ISSUE #6969: SSL Configuration does not support certificate chains which are required for java driver with real certs
certificate chains are really necessary when using officially signed certificates by well known CAs that use intermediate certificates (and you don't want to have to have each client include the CAs intermediate certificates in the client configuration, which is the whole reason for using officially signed certificates).

This should be a quick fix - on line 
https://github.com/rethinkdb/rethinkdb/blob/ec80a184092f5484326b9a581c21b227274099d3/src/clustering/administration/main/command_line.cc#L745 - it currently uses `SSL_CTX_use_certificate_file` but can be switched to use `SSL_CTX_use_certificate_chain_file`

The documentation (Notes section) https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_use_certificate_chain_file.html describes why `SSL_CTX_use_certificate_chain_file()` should be preferred

MarianoRD:

### Description
This commit fixes issue #6969 and also improves the general support of RethinkDB for TLS.

On a personal note, using Let's Encrypt has been a little pain with permissions and tweaking the configuration so RethinkDB can use the certificates. With this change, everything worked first try on a clean environment.

### DIFF
--- a/src/clustering/administration/main/command_line.cc
+++ b/src/clustering/administration/main/command_line.cc
@@ -791,7 +791,7 @@ bool load_tls_key_and_cert(
         return false;
     }
 
-    if(SSL_CTX_use_certificate_file(tls_ctx, cert_file.c_str(), SSL_FILETYPE_PEM) <= 0) {
+    if(SSL_CTX_use_certificate_chain_file(tls_ctx, cert_file.c_str()) <= 0) {
         ERR_print_errors_fp(stderr);
         return false;
     }


srh:
Merged in v2.4.x as of a44f89917b81ad7b049bfb27ad2f62685a3fc7f9.  I reran unit tests locally and they passed.

PR #7084, opened 2022-08-11
Switch PHP driver in README back to danielmewes' php-rql

srh:
The tbolier driver is incomplete, according to its [roadmap](https://github.com/tbolier/php-rethink-ql/blob/master/docs/roadmap.md),
and a quick perusal of the codebase.

There is some discussion at https://github.com/rethinkdb/rethinkdb/commit/327e8aace86fd132a27c48e6dbca935a3a9679d1

@tbolier
@ForbiddenEra

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Besides our four official drivers, we also have many [third-party drivers](https
 * **Elixir:** [rethinkdb-elixir](https://github.com/rethinkdb/rethinkdb-elixir)
 * **Go:** [GoRethink](https://github.com/dancannon/gorethink)
 * **Haskell:** [haskell-rethinkdb](https://github.com/atnnn/haskell-rethinkdb)
-* **PHP:** [php-rethink-ql](https://github.com/tbolier/php-rethink-ql)
+* **PHP:** [php-rql](https://github.com/danielmewes/php-rql)
 * **Rust:** [reql](https://github.com/rust-rethinkdb/reql)
 * **Scala:** [rethink-scala](https://github.com/kclay/rethink-scala)
 


tbolier:
@srh Unfortunately I build the driver when the project was dead at that moment. After that we revamped RethinkDB with a group together, and it never really got of the ground, so the driver development also got to a halt.

My focus was always to continue with this driver development and make the API complete, I'll add my thoughts to the discussion topic.

Cheers

ForbiddenEra:
> @srh Unfortunately I build the driver when the project was dead at that moment. After that we revamped RethinkDB with a group together, and it never really got of the ground, so the driver development also got to a halt.
> 
> My focus was always to continue with this driver development and make the API complete, I'll add my thoughts to the discussion topic.
> 
> Cheers

Yeah I was definitely worried RethinkDB was going to die. Thank @srh and others for keeping it alive; IMHO it should be way more popular than it is, I've never really needed to use any other NoSQL-style DB since vowing to never use SQL again and I <3 RQL.

srh:
@ForbiddenEra It's *very* incomplete and it isn't good for users to be given it as an option.

tbolier:
> @ForbiddenEra The driver is _very_ incomplete. I think the list should be an exhaustive list of drivers we'd feel comfortable with users trying out and using. According to https://github.com/tbolier/php-rethink-ql/blob/master/docs/roadmap.md and a quick confirmation by searching the codebase, it doesn't have so many _basic_ commands, like `add` or `sub` or `branch`.

Thanks @srh and @ForbiddenEra I would like to keep supporting this driver, especially since I need to get my hands dirty on PHP again :-D mainly working in Java and Kotlin these days.

What are the basic commands we can agree on to publish it back into the readme? We can indeed keep as an alternative to daniel-mewes library. I have no interest in replacing it, the goal was to rebuild a more solid fluent library instead that is maintained by the community.

I would like to work on implementing the basic commands in the next iterations, and maybe spread out the word more, so we maybe have some contributors to the lib to maintain it.

What do you think?

srh:
@tbolier It would need to support pretty much all of ReQL's commands to be linked from the README.  Missing features would need to be the exception, rather than the rule.

srh:
Generally, the purpose of the list, its only customer, is the end-user, who is trying to use the product.  It has to benefit the end-user to have the driver listed.

srh:
Of relevance to this thread, I have brought https://github.com/srh/php-rql/tree/srh/dev up to 2.4 ReQL compatibility.

PR #7098, opened 2022-10-28
Add much include-what-you-use for futureproofing

srh:
In reaction to #7091, this change adds a bunch of includes of system headers or library headers.  This is NOT a blindly obedient change that has everything from include-what-you-use.  It's probably too eager about including C++ headers.


srh:
It is strange how check_style.sh has different behavior in the automated checks than when run locally.

Issue #7102, opened 2022-11-10
Codespell report for "RethinkDB" (on fossies.org)


jschleus:
The FOSS server [fossies.org](https://fossies.org) - supporting now also the [RethinkDB](https://fossies.org/linux/rethinkdb) project - offers among others a feature named "[Source code misspelling reports](https://fossies.org/features.html#codespell)". Such reports are normally only generated on request, but as Fossies administrator I have just created such a [codespell](https://github.com/codespell-project/codespell) based analysis for the "RethinkDB" GitHub v2.4.x development version:

&nbsp;https://fossies.org/linux/test/rethinkdb-v2.4.x-snap.tar.gz/codespell.html

That URL should redirect always to the report for the latest version identified by the short GitHub commit ID and a year-month-day string (YYMMDD) representing the according git pull date (mostly = commit date). The data are residing within a special restricted "test" folder that isn't really integrated into the standard Fossies services and should also not be accessible to search engines. The report should be available at least for some weeks and is continuously updated (currently with a half-hourly refresh rate).

By the way, the context type of the spelling errors and typos are marked by a single character (within the sortable column "T"): The less interesting, probably not user visible ones (in "comments") are greyed, while the probably more interesting ones are black-colored. Especially interesting may be unassigned errors marked by a "?" since they may be contained in the source code itself. Some more according information is offered by a tooltip if the mouse is over the string "error context type character" above the table, over the "T" in the table header or over such a context character itself within the table. If JavaScript is enabled all table columns are sortable.

Although after a first review some obviously wrong matches ("False Positives" = FPs) are already filtered out (ignored) please inform me if you find more of them so that I can exclude them if applicable.

Most misspellings are found only in source code comments but e.g. the typo "processs" in the file [test/common/driver.py](https://fossies.org/linux/test/rethinkdb-v2.4.x.snap.tar.gz/rethinkdb-v2.4.x/test/common/driver.py?M=892#l_891) (at line 892) may affect the code itself.

Just for information there are also three supplemental pages showing some used "codespell" [configuration details](https://fossies.org/linux/test/rethinkdb-v2.4.x-snap.tar.gz/codespell_conf.html), all obvious [false positives](https://fossies.org/linux/test/rethinkdb-v2.4.x-snap.tar.gz/codespell_fps.html) and a [misspelling history](https://fossies.org/linux/test/rethinkdb-v2.4.x-snap.tar.gz/codespell_hist.html) (log).

[...]

PR #7103, opened 2022-11-14
Fix typos

srh:
Addresses #7102.  Didn't fix any comments.

Some error message strings in the server were affected.


srh:
This actually shouldn't go into 2.4.3, because it changes error messages.

PR #7135, opened 2023-12-05
Regenerate the web_assets.cc for v2.4.4

srh:
This both (1) updates the version number to v2.4.4, and also (2) regenerates web_assets.cc with file paths in sorted order.  Future regenerations will have smaller diffs.


srh:
WTF
```
       "ire==\"function\"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({\""
-      "/home/srh/rdb_ui/admin/static/coffee/app.coffee\":[function(require,module,e"
-      "xports){\n"
+      "/home/srh/rdbadmin/admin/static/coffee/app.coffee\":[function(require,module"
+      ",exports){\n"
       "var Driver,driver,is_disconnected,r,system_db,bind=function(e,r){return func"
```
I don't think there should be full paths like this.  Maybe this is an internal cosmetic issue, but I'm going to have to look at how this happened and how to fix it.

PR #7193, opened 2026-03-25
fix: correct tautological invariant in raft_core.tcc

Qian-Cheng-nju:
Hi!

Tiny copy-paste fix in the debug invariant check.

At raft_core.tcc:852, the guarantee checks `readiness_for_change || !readiness_for_change` which is always true. The intended check is `readiness_for_change || !readiness_for_config_change` — i.e. "we shouldn't be accepting config changes but not regular changes". The second variable name just got copied from the first.


srh:
Thanks for the change @Qian-Cheng-nju .

Could you add a Developer's Certificate of Origin to your commit?

You can use `git rebase --signoff` and re-push to do that.

Qian-Cheng-nju:
Done — added signoff via rebase.

srh (approved):

srh:
Thank you again for the change.  I'll merge it when the automatic build finishes.

